Issue #323, opened 2016-09-17
Support for Python 3.2 broken


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/323
- Originally reported by: @Ch00k
- Originally created at: 2016-04-11T10:45:39.325

I just came across a problem with tox and Python 3.2 that is described here: https://github.com/travis-ci/travis-ci/issues/5485
I am running tox by the ShiningPanda Jenkins plugin which installs tox like this:

```
#!

[py32] $ /var/lib/jenkins/shiningpanda/jobs/f517a249/tools/bin/python -c "import pip; pip.main();" install --upgrade tox
Collecting tox
  Using cached tox-2.3.1-py2.py3-none-any.whl
Collecting virtualenv>=1.11.2 (from tox)
  Using cached virtualenv-15.0.1-py2.py3-none-any.whl
Collecting py>=1.4.17 (from tox)
  Using cached py-1.4.31-py2.py3-none-any.whl
Collecting pluggy<0.4.0,>=0.3.0 (from tox)
  Using cached pluggy-0.3.1-py2.py3-none-any.whl
```

This installs virtualenv version 15.0.1 which comes from the tox's [setup.py](https://bitbucket.org/hpk42/tox/src/4dbfb1f1eb86b35cbd2c83461e1e294651e89cf6/setup.py?fileviewer=file-view-default#setup.py-39)
Looks like changing that to `virtualenv<14.0.0` would fix the problem.

hpk42:
I agree with @The-Compiler that py32 support is not tox core concern.  You could do a "tox-withpy32" package on pypi which pulls in the right dependencies before installing tox.

jayvdb:
> I agree with @The-Compiler that py32 support is not tox core concern.
> You could do a "tox-withpy32" package on pypi which pulls in the right dependencies before installing tox.

Note that [`tox-travis`](https://pypi.python.org/pypi/tox-travis) has done this, and it isnt easy to do properly due to broken environment marker support in older versions of pip and setuptools (and as a result I appreciate why tox core doesnt want to provide the solution for this).

Anyone wanting to do this should steal the code from `tox-travis`' [`setup.py`](https://github.com/ryanhiebert/tox-travis/blob/5c51e4cd738bcfc9cde80940df932d5d8e88c62a/setup.py), which is ugly but necessary to be effective.

jaraco:
[...]

Issue #338, opened 2016-09-17
Generalize package builds and virtualenv creation


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/338
- Originally reported by: @obestwalter
- Originally created at: 2016-06-24T15:13:14.472

This can serve as a meta issue collecting some outstanding issues that would be fixed by this.

The approach is to write documentation with examples for this and having a discussion around this in this issue.

There also are  drafts with discussion notes which can be used to flesh out the documentation for the functionality in a [draft](https://github.com/tox-dev/tox/blob/master/doc/drafts/extend-envs-and-packagebuilds.md) and some [notes](https://github.com/tox-dev/tox/blob/master/doc/drafts/tox_conda_notes_niccodemus.md).

related issues:
- Handle mutliple backends for virtual environments (#313)
- Support conda envs when using Miniconda (#273)
- maybe Support py32 environments (#323)
- use venv or rwt: https://github.com/tox-dev/tox/issues/394#issuecomment-257295082

There is also a project that solved the problem for conda specifically: https://github.com/hayd/ctox

PR #387, opened 2016-10-25
Support creating custom environments

thomwiggers:
This is an attempt at trying to support custom environments to address issue #323. I am using the suggested implementation by @danilobellini to fix python 3.2 support in a more generic way

Lots of review is welcome and probably needed, I don't fully understood how everything works and have left some `FIXME` comments.

### DIFF
--- a/tox/venv.py
+++ b/tox/venv.py
@@ -5,7 +5,7 @@ import re
 import codecs
 import py
 import tox
-from .config import DepConfig, hookimpl
+from .config import DepConfig, hookimpl, EnvcreatecmdOption
 
 
 class CreationConfig:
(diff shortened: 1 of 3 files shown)


tox/venv.py
```diff
@@ -397,25 +402,51 @@ def tox_testenv_create(venv, action):
...
-    venv.session.make_emptydir(venv.path)
-    basepath = venv.path.dirpath()
-    basepath.ensure(dir=1)
-    args.append(venv.path.basename)
     venv._pcall(args, venv=False, action=action, cwd=basepath)
 
 
 @hookimpl
+def tox_testenv_prepare(venv, action):
+    # FIXME this is probably not the right way to do things
```
thomwiggers: Doing it this way also runs it without a shell, so things like `curl a | b` don't work.
Also, we're going through the whitelist, which makes it annoying to use.

hpk42:
Thanks for your effort! FYI there also is https://github.com/tox-dev/tox/issues/338 which deals with the issue of tox delegating env creation and introducing new config.  What i'd like to see is **documentation** of how things are going to work according to use-case/user views  And i suggest to drive flexibilizing tox's env creation by by conda support with  py32 support as an additional thought only.

PR #392, opened 2016-10-27
fix #348 - add py2 and py3 as default envs

### LINKED ISSUE #348: Consider supporting 'py2' and 'py3' default environments
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/348
- Originally reported by: [tobias.mcnulty](http://bitbucket.org/tobias.mcnulty)
- Originally created at: 2016-07-19T22:12:41.440

For packages where I want to test *some* version of python3 but I don't really care which one, support for `py2` and `py3` environments would be nice, e.g.:

	[testenv:py2]
	basepython = python2

	[testenv:py3]
	basepython = python3

I can create these easily enough myself, but thought I'd submit this issue in case anyone else would find it useful.

hpk42:


ewjoachim:
Just so you know, I'm grateful that this change happened, and I'll never thank you folks enough for tox, but this change broke my build :sob:

(It was an env that tested py3 compatibility but launched by py2, on a machine that didn't have py3 installed. It was named "lint-py3" so that it would not match existing default filter. And I know it's my fault that I didn't pin the tox version on my CI on the first place.)

But again, this was my fault and thanks for making it go forward !

hpk42:
thanks for letting us know.  hope it wasn't too hard to fix the breakage.

ewjoachim:
No, really, it was just a matter of renaming the env "lint-py-three" or specifying the interpreter explicitely `basepython = python2`.

Issue #353, opened 2016-09-17
More verbose when installing deps!


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/353
- Originally reported by: [danilobellini](http://bitbucket.org/danilobellini)
- Originally created at: 2016-08-03T05:22:07.989

I've got some projects whose requirements take a long time until pip installs all the deps, e.g., projects that require scipy. Running such projects in a CI results in a timeout error before the installation finishes, due to the lack of new output for too long.

In the past I tried to install these packages as part of the commands, but as I said, it takes too long, that's not something I wish to reinstall every time I'm calling the tox for testing my code. And even if I call `pip install -I` in my commands, the CI wouldn't be able to distinguish between an installation error (the ones that might happen when I call `tox --notest` during installation) and an actual project failure. I think there should be a verbosity level that shows the installer outputs while it's installing the deps.

hpk42:
@danilobelini maybe a new option like ``-no-output-redirect`` or so could work?  I am also open to changing the meaning of ``-vv``.  Happy about a PR that goes for either of that.

rogalski:
I guess it can be closed, since my PR is merged.

hpk42:
yes, thanks Lukasz!

PR #403, opened 2016-11-05
Verbosity now affects package installation

rogalski:
Solves  #353.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -29,6 +29,9 @@ unreleased
 - fix #305: ``downloadcache`` test env config is now ignored as pip-8
   does caching by default. Thanks holger krekel.
 
+- output from install command in verbose (-vv) mode is now printed to console instead of
+  being redirected to file, thanks Lukasz Rogalski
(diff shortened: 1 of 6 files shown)


hpk42:
Thanks for the PR.  See one inline comment.  I wonder if some people will be surprised that ``-vv`` turns of redirection.  If so we might have to rather go for a separate option.  Please amend the help string of ``-v`` to say that if it's supplied twice output from commands will be shown, not redirected to a file.  This way people can discover it via ``tox -h | grep output`` or so.

The-Compiler:
FWIW I'm surprised by the current behaviour - I'd absolutely expect `-vv` to show me pip install output :wink:

rogalski:
After code review is completed, I'll happily squash and rebase change, since right now history is quite messy.

hpk42:
Looks good, thanks!  Please rebase and then we can merge.

nicoddemus:
Went ahead with @hpk42's instructions. 👍

PR #415, opened 2016-12-02
Pass setenv variables to setup.py during usedevelop's reinstall check

eli-collins:
This fixes venv._needs_reinstall() so that setenv variables are passed along to the setup.py call.

### DIFF
--- a/tests/test_venv.py
+++ b/tests/test_venv.py
@@ -292,6 +292,7 @@ def test_env_variables_added_to_needs_reinstall(tmpdir, mocksession, newconfig, 
     assert 'TEMP_NOPASS_VAR' in env
     assert env["TEMP_NOPASS_VAR"] == "456"
 
+
 def test_test_hashseed_is_in_output(newmocksession):
     original_make_hashseed = tox.config.make_hashseed
     tox.config.make_hashseed = lambda: '123456789'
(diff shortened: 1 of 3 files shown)


tests/test_venv.py
```diff
@@ -265,6 +265,31 @@ def test_develop_extras(newmocksession, tmpdir):
...
+    venv._needs_reinstall(tmpdir, action)
+
+    l = mocksession._pcalls
+    assert len(l) == 1
+    env = l[0].env
+    assert 'CUSTOM_VAR' in env
+    assert env['CUSTOM_VAR'] == '789'
+    assert 'TEMP_PASS_VAR' in env
+    assert env['TEMP_PASS_VAR'] == "123"
+    assert 'TEMP_NOPASS_VAR' in env
```
nicoddemus: Shouldn't `TEMP_NOPASS_VAR` **not** be in `env`?

eli-collins: I'm not familiar with tox's overall structure, but that seemed more consistent with the old behavior (since it was previously being invoked under the outside environment).  And it's inline with the venv.run_install_command() -- which invokes venv._pcall() in such a way that it's *also* called with the full outside environment + setenv vars.   Per venv._getenv(), it looked to me like only the test commands themselves get the restricted environment.

hpk42: makes sense i think.  could you add a comment to the test and/or venv.py code ?

PR #432, opened 2017-01-05
Added py36 to tox.ini

andriisoldatenko:
>Python 3.6.0 is the newest major release of the Python language, and it contains many new features and optimizations.
[Python 3.6.0](https://www.python.org/downloads/release/python-360/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,15 @@ python:
 - '3.3'
 - '3.4'
 - '3.5'
-- 3.6-dev
+- '3.6'
+- 3.7-dev
 - nightly
 - pypy
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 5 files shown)


nicoddemus:
Thanks! 

Could you please also update `appveyor.yml` and `.travis.yml`? In `.travis.yml` I believe we can use the standard `'3.6'` Python instead of `- 3.6-dev` (untested).

andriisoldatenko:
Yes sure.

nicoddemus:
Update: it seems Travis still needs to use `3.6-dev` though.

andriisoldatenko:
@nicoddemus I've already done this change in couple projects.
I think now we have `3.7-dev`.

Issue #263, opened 2016-09-17
A real-world problem with Tox 2+ environment variable isolation


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/263
- Originally reported by: @saaj
- Originally created at: 2015-07-04T21:10:43.053

Another v2 problem that took me several hours of wondering _why_.

Briefly, it was a CI service _drone.io_ and a code coverage service _codecov.io_. Here's the relevant snippet from its client that can explain the problem without me talking much about it:

```
#!python

elif os.getenv('CI') == "true" and os.getenv('DRONE') == "true":
    # http://docs.drone.io/env.html
    defaults.update(dict(branch=os.getenv('DRONE_BRANCH'),
                         service='drone.io',
                         build=os.getenv('DRONE_BUILD_NUMBER'),
                         build_url=os.getenv('DRONE_BUILD_URL'),
                         commit=os.getenv('DRONE_COMMIT')))
```

And an QA env that may look something like this obviously no longer sends coverage data because it can't get information about the environment.

```
#!ini

[testenv:qa]
deps = coverage
commands = 
  coverage run --branch --source=lib setup.py test
  codecov
```

I bet most of much QA services and their clients work the same way to identify service, commit, branches, etc. And it's quite reasonable to do such things from within a Tox env or even have dedicated Tox envs. So there's a serious doubt that this idea of environment isolation (which already leaks because of LANG exception) turned on by default is compatible with real-world Tox usage. At very least there should be an option to restore outer environment at once.

Of course one can look at the env spec of a CI service, then examine QA service's client and put bunch of  `passenv` in `tox.ini`. But have doubts than many developer will be happy about it. 

As a side note, because it's already third v2 problem (others were missing LANG with broken py3, missing {envsitepackagesdir}) I tend to think current v2 series is not production ready and is beta at most. Reverted to v1 again.

obestwalter:
I'll close this discussion as basically settled now. AFAICT the default isolating behaviour is now accepted at this point and it can easily be changed by either passing `TOX_TESTENV_PASSENV=*` on invocation or using `passenv=*` inside the testenv section. Changing the behaviour again in a future major version would only introduce more inconsistencies and confusion I guess.

PR #433, opened 2017-01-05
Fix glob resolution from TOX_TESTENV_PASSENV env variable.

a-feld:
https://github.com/tox-dev/tox/issues/263#issuecomment-247790754
Glob resolution from the TOX_TESTENV_PASSENV env variable should be
supported.

This update allows for globs to be resolved from the TOX_TESTENV_PASSENV
variable.

Thanks for contributing a PR!

- [x] ~~if an enhacement PR please create docs and at best an example~~

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -38,3 +38,4 @@ Manuel Jacob
 Eli Collins
 Andrii Soldatenko
 Igor Duarte Cardoso
+Allan Feldman
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -935,6 +935,17 @@ class TestConfigTestEnv:
         assert "A1" in env.passenv
         assert "A2" in env.passenv
 
+    def test_passenv_glob_from_global_env(self, tmpdir, newconfig, monkeypatch):
+        monkeypatch.setenv("A1", "a1")
+        monkeypatch.setenv("A2", "a2")
+        monkeypatch.setenv("TOX_TESTENV_PASSENV", "A*")
+        config = newconfig("""
+            [testenv]
+        """)
+        env = config.envconfigs["python"]
+        assert "A1" in env.passenv
+        assert "A2" in env.passenv
+
     def test_changedir_override(self, tmpdir, newconfig):
         config = newconfig("""
             [testenv]
--- a/tox/config.py
+++ b/tox/config.py
@@ -456,7 +456,8 @@ def tox_addoption(parser):
         # read in global passenv settings
         p = os.environ.get("TOX_TESTENV_PASSENV", None)
         if p is not None:
-            passenv.update(x for x in p.split() if x)
+            env_values = [x for x in p.split() if x]
+            value.extend(env_values)
 
         # we ensure that tmp directory settings are passed on
         # we could also set it to the per-venv "envtmpdir"


obestwalter:
@a-feld I do not understand the purpose of this PR. The comment you reference states that the wildcard passing of env vars does work already, so there is nothing to be done. 

e.g. `passenv=A*` (er the env var equivalent) like in you testcase works already in current releases of tox. Or am I missing something?

a-feld:
Hi @obestwalter , the test is created to show that the test **does not** work in current releases of tox.  The fix for the currently failing test is included in this PR.

a-feld:
@obestwalter I have separated the test / fix commits to make the issue more easily discernible.

To reproduce the problem, run:

1. `git checkout aa6c42f`
2. `python setup.py test`

Verify that the test added fails.

Run
1. `git checkout 44b8d14`
2. `python setup.py test`

Verify that the test added now passes.

obestwalter:
Woah - now I see. I was assuming that the `passenv` entry in `tox.ini` does exactly the same as `TOX_TESTENV_PASSENV` but it doesn't so that's the problem. While you were updating the PR (thanks - very helpful) I did my own little experiment but only with passenv:

with tox.ini:

    [testenv]
    passenv = ENV_*
    commands = python envtester.py

envtester.py:

    import os

    assert os.getenv('ENV_VAR_TEST') == 'WOHOO!1!!'

This works:

    (tox) 00:52:14 oliver@ob1 [0] < ~/work/tox/tests >  1714 %
    ENV_VAR_TEST='WOHOO!1!!' tox                        
    GLOB sdist-make: /home/oliver/work/tox/tests/setup.py
    python inst-nodeps: /home/oliver/work/tox/tests/.tox/dist/dummy-setup-0.0.0.zip
    python installed: dummy-setup==0.0.0
    python runtests: PYTHONHASHSEED='1508607320'
    python runtests: commands[0] | python envtester.py
    ______________________________ summary ______________________________________________
      python: commands succeeded
      congratulations :)

But you are correct - this doesn't

with tox.ini:

    [testenv]
    commands = python envtester.py

    (tox) 00:55:11 oliver@ob1 [0] < ~/work/tox/tests >  1715 %
    TOX_TESTENV_PASSENV='ENV_*' ENV_VAR_TEST='WOHOO!1!' tox
    GLOB sdist-make: /home/oliver/work/tox/tests/setup.py
    python inst-nodeps: /home/oliver/work/tox/tests/.tox/dist/dummy-setup-0.0.0.zip
    python installed: dummy-setup==0.0.0
    python runtests: PYTHONHASHSEED='3223092831'
    python runtests: commands[0] | python envtester.py
      File "envtester.py", line 3, in <module>
        assert os.getenv('ENV_VAR_TEST') == 'WOHOO!1!!'
    AssertionError
    ERROR: InvocationError: '/home/oliver/work/tox/tests/.tox/python/bin/python envtester.py'
    ______________________________ summary ______________________________________________
    ERROR:   python: commands failed

obestwalter:
I want to have a closer look why this is tomorrow. I am not that familiar with the code base yet, but I was living under the impression that there is some magic going on that synchronizes ini settings with env vars and this is even shared between pytest and tox ... I guess I was wrong. Will check tomorow and then probably merge your fix. Thank you!

a-feld:
Here's a debug summary I've compiled that may be helpful as well

Debug Summary
--------------
* in `make_envconfig`, `res=[]` if the attribute being processed is a "line-list"
  attr type and the `tox.ini` file does not contain entries for that attribute
  ([link]( https://github.com/tox-dev/tox/blob/2.5.0/tox/config.py#L795))
* `make_envconfig` calls `passenv` through the `env_attr.postprocess` attribute
  ([link](https://github.com/tox-dev/tox/blob/2.5.0/tox/config.py#L800))
* passenv reads the `TOX_TESTENV_PASSENV` environment variable and immediately
  adds it to the set of environment variables to pass to the downstream tox
  environment
  ([link](https://github.com/tox-dev/tox/blob/2.5.0/tox/config.py#L457))
* The values extracted from the `tox.ini` (held in variable `value`) are
  iterated over to resolve globs (using `fnmatchcase`) and added to the set of
  "passed" environment variables (passenv)
  ([link](https://github.com/tox-dev/tox/blob/2.5.0/tox/config.py#L472-L475))

Note that with the logic above, the globbing is not resolved for the
`TOX_TESTENV_PASSENV` environment variable

Proposed Solution
-----------------
The "value" list should be updated with the values extracted from the
`TOX_TESTENV_PASSENV` environment variable.

This solution allows globbing maintains a common code path for both the
`tox.ini` passenv attribute and the `TOX_TESTENV_PASSENV` environment variable.

Additional Considerations
-----------------------
Precedence between `TOX_TESTENV_PASSENV` and the `tox.ini` attributes should be
considered.
With the current implementation, the passenv set is the union of
set(tox_ini_attrs) and set(TOX_TESTENV_PASSENV).
The proposed solution maintains this behavior.

Duplicate variables / duplicate matches should be considered.
i.e. if the `tox.ini: passenv = foo` and `TOX_TESTENV_PASSENV=foo`, passenv
should resolve to set('foo').
The proposed solution above allows for this case since `add` is idempotent on
set data containers (adding an item already in a set will not change the
contents of the container)

PR #438, opened 2017-01-17
Use venv module in python 3

rpkilby:
This is a proof of concept for issue #436. The fix is incoming after the failing test run is complete. I'm assuming this will probably fail on python 3.3, since the `--copies` option wasn't implemented yet.


rpkilby:
bleh - test isn't working as expected. It was failing on my system for a different reason.

rpkilby:
I've pushed the code to demonstrate that migrating to `venv` shouldn't be too difficult. Either way, closing, as the original issue is actually invalid.

PR #441, opened 2017-01-22
Correct fallback behavior on Python 3.2

jaraco:
Thanks to the review by @webknjaz, there's a bug in the patch from #428. This change corrects that bug.


hpk42:
thanks @jaraco!

PR #444, opened 2017-01-25
Minor doc fixes

merwok:
I noticed wrong rendering on a page, fixed it in the first commit.

Then I wondered if install_command was used to install deps, and added cross-references to make that more obvious.


obestwalter:
@merwok thank you. Do you want to add yourself to https://github.com/tox-dev/tox/blob/master/CONTRIBUTORS before I merge?

merwok:
I’ll do that when I contribute something of substance :)

PR #451, opened 2017-02-07
Run tests on macOS. Fixes #446.

### LINKED ISSUE #446: Run tests on OS X with Travis CI?
It's [possible](https://docs.travis-ci.com/user/osx-ci-environment/#Homebrew) and as we officially support it, it might be a good idea.

It would be great, if somebody who works mainly on OS X would step up to do this.

jaraco:

### DIFF
--- a/.travis-osx
+++ b/.travis-osx
@@ -1,0 +1,6 @@
+# Perform the manual steps on osx to install python3 and activate an environment
+brew update
+brew install python3
+rm /usr/local/bin/python
+ln -s python3 /usr/local/bin/python
+ln -s pip3 /usr/local/bin/pip
--- a/.travis-osx
+++ b/.travis-osx
@@ -3,4 +3,4 @@ brew update
 brew install python3
 rm /usr/local/bin/python
 ln -s python3 /usr/local/bin/python
-ln -s pip3 /usr/local/bin/pip
+ln -fs pip3 /usr/local/bin/pip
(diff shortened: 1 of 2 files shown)


jaraco:
It took a while to get right due to the 30-70 minute delay between push and test run, but it's now passing. I've noticed other projects have disabled macOS tests in Travis because they're so slow. If you decide not to accept this, I'm okay with it.

obestwalter:
@jaraco thanks a lot. Is this a temporary bottle neck or a permanent problem on Travis CI? I have no experience woth running OS X tests on Travis, so I couldn't say. I'll merge it and if the delay is proving problematic we see what we can do. Let's hope Travis CI buys a few more Macs soon :)

If this turns out to be not working at all, I might adapt @nicoddemus approach: https://github.com/nicoddemus/devpi-cloud-tester - this won't be integrated on Github then though, but at least on devpi.net: https://github.com/nicoddemus/devpi-cloud-tester

jaraco:
> Is this a temporary bottle neck or a permanent problem on Travis CI?

I'm not sure. I've regularly noticed the builds being delayed, but they always seem reliable, if slow. I've enabled them for a few high-profile projects and it's been worthwhile. In my mind, the slowness will only prove problemmatic if some other process is gated on the test acceptance (such as pull requests are). I believe there are some Travis options we could employ to make macOS tests optional.

obestwalter:
Well, [results are in](https://travis-ci.org/tox-dev/tox/jobs/200384153) and everything works fine and I don't really think the slowness will bother us too much then.

Thank you again :)

Issue #454, opened 2017-02-12
Regression since 2.5.0: empty commands section causes error


asottile:
We often use this pattern such that tox is creating our virtual environments for us:

## tox.ini

```ini
[tox]
skipsdist = true

[testenv:venv]
deps = pre-commit
commands =
```

## Under 2.4.1

I get (what I expect):
```
$ tox -e venv
venv recreate: /tmp/fo/.tox/venv
venv installdeps: pre-commit
venv installed: appdirs==1.4.0,aspy.yaml==0.2.2,cached-property==1.3.0,functools32==3.2.3.post2,jsonschema==2.6.0,nodeenv==1.1.2,packaging==16.8,pre-commit==0.12.2,pyparsing==2.1.10,PyYAML==3.12,six==1.10.0,virtualenv==15.1.0
venv runtests: PYTHONHASHSEED='2591024316'
___________________________________ summary ____________________________________
  venv: commands succeeded
  congratulations :)
```

## 2.5.0

This is an error:
```
$ tox -e venv
venv create: /tmp/fo/.tox/venv
venv installdeps: pre-commit
venv installed: appdirs==1.4.0,aspy.yaml==0.2.2,cached-property==1.3.0,functools32==3.2.3.post2,jsonschema==2.6.0,nodeenv==1.1.2,packaging==16.8,pre-commit==0.12.2,pyparsing==2.1.10,PyYAML==3.12,six==1.10.0,virtualenv==15.1.0
ERROR: Commands not specified. Please update relevant section of /tmp/fo/tox.ini
___________________________________ summary ____________________________________
ERROR:   venv: nothing to do
```

## workaround
A (kinda unfortunate) workaround is to do something like:

```ini
commands = python -c 'print("done!")'
```

obestwalter:
Hi @asottile, that's corect. already identified [here](https://github.com/tox-dev/tox/issues/316#issuecomment-270795986), but we should have a proper bug for this. Closed the other one in favour of this.

asottile:
Since this doesn't seem simple to fix, should we start by reverting #407?

The parser doesn't seem to differentiate between "missing" and "present but empty"

obestwalter:
I think that's a good idea, as it is better not to have a regression than a broken new feature.

asottile:
This was reverted in #456 -- perhaps the original issue should be reopened and this one closed?

obestwalter:
Done. Thanks @asottile!

Issue #316, opened 2016-09-17
Warn user if they forgot commands section


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/316
- Originally reported by: @The-Compiler
- Originally created at: 2016-02-25T11:55:21.179

When one forgets to add `commands = ...` to `tox.ini`, this is most likely a mistake - however, tox happily runs nothing and tells us the commands succeeded.

I think instead it should output a warning, similar to the yellow "no tests ran" which pytest does.

obestwalter:
@The-Compiler I think this is a good idea to add that. I changed the name of the issue to reflect what should happen rather then what not happens.

The-Compiler:
Cool - thanks for all the issue triaging you're doing lately, it's much appreciated! :smile:

vlcinsky:
The ticket talks about WARNing, but the modification is reporting an ERROR and exits with status code 1.

I am often using tox.ini as builder for virtual envs and in those cases I often do not expect any command to be run.

The ERROR message is thus confusing. My first impression was, that the build of virtualenv failed what was not true.

Consider changing this warning only and keep the empty command section as acceptable scenario not reported as a failure.

obestwalter:
> I am often using tox.ini as builder for virtual envs and in those cases I often do not expect any command to be run.

@vlcinsky I agree the commands section must not be mandatory. The issue explicitly states also that we want to warn the user and not throw an error. #407 needs a bit of work as @rogalski already said - I would suggest to print an informative warning about the missing commands section but not throw an error.

vlcinsky:
@obestwalter so we agree.

Using tox v 2.5.0 I get the ERROR reported if command is empty. I wonder, if this issue shall be reopen or I shall file a new one.

obestwalter:
Let's reopen this. As this made it into a release already I am not sure how to proceed. Let's hear some more opinions about this ...

obestwalter:
I just had a look in the [docs](http://tox.readthedocs.io/en/latest/config.html?highlight=commands#virtualenv-test-environment-settings) and there are no mandatory settings in the `testenv` section. It says explicitly: "Complete list of settings that you **can** put into testenv* sections:" (emphasis by me). So we should consider this a bug that we now get an error if no commands section is present. 

I also think that it is a valid use case to use tox to only create a virtualenv without executing any commands - I even seem to remember that this is mentioned in the docs somewhere.

vlcinsky:
[...]

PR #456, opened 2017-02-14
Revert "Warn about commands key missing in tox.ini"

asottile:
Reverts tox-dev/tox#407

See #454 and #316

I'd rather fix this to have the following behaviour but at a glance it doesn't seem easy / possible with the current `argvlist` parser (which doesn't differentiate between missing and present-but-empy):
- If there are no `commands`: warning (or error)
- If `commands` is present but empty: successfully do nothing (no warning).

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 2.x
 -----
 
+- #454: Revert #407, empty commands is not treated as an error.
+  Thanks Anthony Sottile (@asottile).
+
 - #446: (infrastructure) Travis CI tests for tox now also run on OS X now.
   Thanks Jason R. Coombs (@jaraco).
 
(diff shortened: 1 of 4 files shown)


CHANGELOG
```diff
@@ -59,9 +59,6 @@
 - fix #399.  Make sure {envtmpdir} is created if it doesn't exist at the
   start of a testenvironment run. Thanks Manuel Jacob.
 
-- fix #316: Lack of commands key in ini file is now treated as an error.
-  Reported virtualenv status is 'nothing to do' instead of 'commands
-  succeeded', with relevant error message displayed. Thanks Lukasz Rogalski.
```
obestwalter: The changelog must not be changed retroactively, so this should not be removed, but instead en entry needs to be added, that this fix did not what was intended to do (throw a warning instead of causing an error) in the changelog for the current version.

Issue #458, opened 2017-02-17
Warn user that PYTHONPATH is discarded if it is set.


obestwalter:
Since 2.4.0 we discard PYTHONPATH silently. The user should be warned if they have set it in there environment though as this can be a source of confusion otherwise.

This is likely very simple to implement and would be a good issue for new contributors.

jarshwah:
Fixed by https://github.com/tox-dev/tox/pull/459

obestwalter:
Thanks @jarshwah.

PR #459, opened 2017-02-18
passenv respects PYTHONPATH. Fixes #457 and #458

### LINKED ISSUE #457: Change preservation behaviour for PYTHONPATH (backwards compatible)
I help to maintain https://github.com/django/django-box which is a vagrant machine setup with all of the dependencies needed to test Django. Django provides a tox file with explicit configuration to support the django-box vagrant machine. See: https://github.com/django/django/blob/3d14cbc86781ea1051af7f0c421bee3ecf2f9842/tox.ini 

```
# relevant section
[testenv]
usedevelop = true
passenv = DJANGO_SETTINGS_MODULE PYTHONPATH HOME DISPLAY
setenv =
    PYTHONDONTWRITEBYTECODE=1
deps =
    py{3,34,35,36}: -rtests/requirements/py3.txt
    postgres: -rtests/requirements/postgres.txt
    mysql: -rtests/requirements/mysql.txt
    oracle: -rtests/requirements/oracle.txt
changedir = tests
commands =
    {envpython} runtests.py {posargs}
```

Various aliases are configured to run the test suite such as:

```
alias runtests36-postgres='PYTHONPATH=/home/vagrant/djangodata/ tox -c /django/tox.ini -e py36-postgres -- --settings=test_postgres'
```

The `/home/vagrant/djangodata/` contains a bunch of python configuration files that define database connections. This directory is added to `PYTHONPATH` to make the configuration files available to the `--settings=` argument. Note the `passenv` in the tox configuration.

This worked perfectly fine in tox 2.3.0. Unfortunately, (after much debugging), I discovered that 2.4.0 removed support for PYTHONPATH entirely, even when explicitly defined in `passenv`.

> remove PYTHONPATH from environment during the install phase because a tox-run should not have hidden dependencies and the test commands will also not see a PYTHONPATH. If this causes unforeseen problems it may be reverted in a bugfix release.

Our goals:

1) Avoid poluting django test directory with unrelated settings files. The settings files are related to the system doing the testing, and are passed to django.

2) Avoid hardcoding paths in the tox file, which isn't designed to only be used with this virtual machine.

If there's a way forward to make the settings directory available to the environment without PYTHONPATH, I'm all ears.

For me though, if I've explicitly configured `PYTHONPATH` in the `passev` configuration, then please, let me pass it through. Tox shouldn't be blessing or blocking environment variables with no work arounds.

I'd like to see this change reverted, but once again, I'm open to discussions on how best to fix this problem.

jarshwah:
Thanks for contributing a PR!

- [ ] if an enhacement PR please create docs and at best an example

### DIFF
--- a/tox/venv.py
+++ b/tox/venv.py
@@ -275,8 +275,17 @@ class VirtualEnv(object):
             argv[i:i + 1] = list(options)
 
         for x in ('PIP_RESPECT_VIRTUALENV', 'PIP_REQUIRE_VIRTUALENV',
-                  '__PYVENV_LAUNCHER__', 'PYTHONPATH'):
+                  '__PYVENV_LAUNCHER__'):
             os.environ.pop(x, None)
+
+        if 'PYTHONPATH' not in self.envconfig.passenv:
+            # If PYTHONPATH not explicitly asked for, remove it.
+            if 'PYTHONPATH' in os.environ:
+                self.session.report.warning(
+                    "Discarding $PYTHONPATH from environment, to override "
+                    "specify PYTHONPATH in 'passenv' in your environment."
+                )
+                os.environ.pop('PYTHONPATH')
 
         old_stdout = sys.stdout
         sys.stdout = codecs.getwriter('utf8')(sys.stdout)
--- a/tox/venv.py
+++ b/tox/venv.py
@@ -283,7 +283,7 @@ class VirtualEnv(object):
             if 'PYTHONPATH' in os.environ:
                 self.session.report.warning(
                     "Discarding $PYTHONPATH from environment, to override "
-                    "specify PYTHONPATH in 'passenv' in your environment."
+                    "specify PYTHONPATH in 'passenv' in your configuration."
                 )
                 os.environ.pop('PYTHONPATH')
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -9,7 +9,7 @@ Barry Warsaw
 Chris Rose
 Jannis Leidel
 Ronny Pfannschmidt
-Lukasz Balcerzak 
+Lukasz Balcerzak
 Philip Thiem
 Monty Taylor
 Bruno Oliveira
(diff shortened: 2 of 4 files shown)


jarshwah:
From https://github.com/tox-dev/tox/issues/457

A warning is now generated if `PYTHONPATH` is in the environment and is not listed in the `passenv` directive. That should avoid future users being confused when it doesn't automatically flow through.

If a user has explicitly set `passenv = PYTHONPATH`, it'll now remain in the environment.

jarshwah:
I'm missing docs. Is there a particular section in the documentation where this should go?

obestwalter:
@jarshwah The main explanation should go into http://tox.readthedocs.io/en/latest/config.html?highlight=passenv#confval-passenv=SPACE-SEPARATED-GLOBNAMES 

Also PYTHONPATH is explixitly used [here](http://tox.readthedocs.io/en/latest/example/basic.html#setting-environment-variables) as an example for setting env vars: so there could be a link to the main explanation there to raise awareness, but as we have the warning now this might not even be necessary.

jarshwah:
I've added docs to the `passenv` setting page, but not to the `setenv` potion. As you say, the warning should be enough. Also, this change does not affect `setenv` in any way (see amended test), so it would be weird to complicate usage in `setenv` examples.

I think I'm done on this patch, but please let me know if there's anything further I can or need to do.

Thanks for the pointers!

tox/venv.py
```diff
@@ -275,9 +275,18 @@ def run_install_command(self, packages, action, options=()):
...
-                  '__PYVENV_LAUNCHER__', 'PYTHONPATH'):
+                  '__PYVENV_LAUNCHER__'):
             os.environ.pop(x, None)
 
+        if 'PYTHONPATH' not in self.envconfig.passenv:
+            # If PYTHONPATH not explicitly asked for, remove it.
+            if 'PYTHONPATH' in os.environ:
+                self.session.report.warning(
+                    "Discarding $PYTHONPATH from environment, to override "
+                    "specify PYTHONPATH in 'passenv' in your environment."
```
jarshwah: Perhaps "'passenv' in your configuration" reads better here?

nicoddemus: Sounds good to me.

PR #450, opened 2017-02-04
Stop after the first installdeps hook succeeds

asottile:
By changing this hook to `firstresult=True`, a plugin may override this step and avoid running the base step

I'd like to write a plugin which overwrites the install_deps command, I've got it mostly working except when I run it it triggers dependency install twice (the second time is a noop, but a slow noop).  

## tox2.ini
```
[tox]
envlist = py27     
tox_pip_extensions_ext_venv_update = true

[testenv]
deps =
    pre-commit
    setuptools
commands =
    pip freeze --all
```

## Current behaviour

```
$ tox -v -r -e py27 -c tox2.ini
using tox.ini: /home/asottile/workspace/tox-pip-extensions/tox2.ini
using tox-2.6.0 from /home/asottile/workspace/tox-pip-extensions/venv/local/lib/python2.7/site-packages/tox/__init__.pyc
GLOB sdist-make: /home/asottile/workspace/tox-pip-extensions/setup.py
  /home/asottile/workspace/tox-pip-extensions$ /home/asottile/workspace/tox-pip-extensions/venv/bin/python /home/asottile/workspace/tox-pip-extensions/setup.py sdist --formats=zip --dist-dir /home/asottile/workspace/tox-pip-extensions/.tox/dist >/home/asottile/workspace/tox-pip-extensions/.tox/log/tox-0.log
py27 recreate: /home/asottile/workspace/tox-pip-extensions/.tox/py27
  /home/asottile/workspace/tox-pip-extensions/.tox$ /home/asottile/workspace/tox-pip-extensions/venv/bin/python -m virtualenv --python /home/asottile/workspace/tox-pip-extensions/venv/bin/python2.7 py27 >/home/asottile/workspace/tox-pip-extensions/.tox/py27/log/py27-0.log
py27 bootstrap: venv-update>=2
  /home/asottile/workspace/tox-pip-extensions$ /home/asottile/workspace/tox-pip-extensions/.tox/py27/bin/pip install venv-update>=2 >/home/asottile/workspace/tox-pip-extensions/.tox/py27/log/py27-1.log
py27 installdeps: pre-commit, setuptools
  /home/asottile/workspace/tox-pip-extensions$ /home/asottile/workspace/tox-pip-extensions/.tox/py27/bin/pip-faster install pre-commit setuptools --prune >/home/asottile/workspace/tox-pip-extensions/.tox/py27/log/py27-2.log
py27 installdeps: pre-commit, setuptools
[...]

PR #469, opened 2017-03-06
Stop after the first testenv create hook succeeds

asottile:
See #450 for discussion

Mostly opening this so both can be considered at the same time and/or separately from #450.


asottile:
Going to merge this into #450

Issue #460, opened 2017-02-20
CLI option to change directory


philtay:
It would be useful to have a `-C` option (like `make`) to change to a specified directory before reading the `tox.ini` file.

obestwalter:
@philtay tox has `-c` which IIUC is basically doing that aleady only that you have to give the path to the config file instead of the directory where the config is located. Would that already solve your problem?

There is also a [changedir option](http://tox.readthedocs.io/en/latest/config.html#confval-changedir=path) that could be used inside the `tox.ini` if that helps.

Otherwise it would be helpful if you explain your use case in a bit more detail, so that we can see if it really would make sense to have `-C` additionally to the existing options.

philtay:
I know about `-c`, but I don't want to specify a full config file path, only the "target" directory. With `-C` (uppercase) `tox` would be looking for a standard `tox.ini` inside the specified directory. The use-case is running `tox` from a `Makefile`: currently the only way is `cd /mydir && tox` (or having to specify a full path when the filename `tox.ini` is essentially a given).

obestwalter:
> I know about -c, but I don't want to specify a full config file path, only the "target" directory. 

So basically `-c` almost does what you want, but you don't want to have to explicitly pass the name of `tox.ini` - which is reasonable as this is in 99% of the cases the name of the tox config.

How about changing the behaviour of the `-c` option that if you point to a directory `tox.ini` is assumed as the name of the config?

That would be your `-C` option without having to add another option. This is a reasonably simple change without breaking backwards compatibility unless there are workflows who rely on getting an error if a path to a directory instead of a filepath is passed, but for me that would fall into the area of [relevant XKCD](https://xkcd.com/1172/).

I guess, a PR with that change would be accepted.

nicoddemus:
Hah nice reference!

nicoddemus:
(currently at home with a stomach bug, so this kind of reference is a plus 😁 )

philtay:
@obestwalter Yes, your idea would solve the problem. Nice XKCD!

pawelad:
Is this feature worked on? I would be happy to do a PR if not.

And just to reiterate- `tox -c relative/path` should have the same behaviour as `tox -c relative/path/tox.ini`, right?

philtay:
@pawelad I'm not working on it, please go ahead. Thanks.
[...]

PR #472, opened 2017-03-08
Made custom config file path parameter more flexible (#460)

pawelad:
Made custom config file path parameter (`-c`) more flexible. It now allows both a path to the `tox.ini` file and a path to the directory where `tox.ini` is located.
Tests added.
I also tried to add some info to docs but couldn't really find a good place for it - any suggestions welcome

Resolves issue #460

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -205,6 +205,21 @@ class TestParseconfig:
             old.chdir()
         assert config.toxinipath == toxinipath
 
+    def test_explicit_config_path(self, tmpdir):
+        """
+        Test explicitly setting config path, both with and without the filename
+        """
+        path = tmpdir.mkdir('tox_tmp_directory')
+        config_file_path = path.ensure('tox.ini')
+
+        config = parseconfig(['-c', str(config_file_path)])
+        assert config.toxinipath == config_file_path
+
+        # Passing directory of the config file should also be possible
+        # ('tox.ini' filename is assumed)
+        config = parseconfig(['-c', str(path)])
+        assert config.toxinipath == config_file_path
+
 
 def test_get_homedir(monkeypatch):
     monkeypatch.setattr(py.path.local, "_gethomedir",
--- a/tox/config.py
+++ b/tox/config.py
@@ -324,7 +324,7 @@ def tox_addoption(parser):
                         dest="listenvs", help="show list of test environments")
     parser.add_argument("-c", action="store", default="tox.ini",
                         dest="configfile",
-                        help="use the specified config file name.")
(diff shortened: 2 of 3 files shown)


pawelad:
I also tried to add some info to docs but couldn't really find a good place for it - any suggestions welcome.

tests/test_config.py
```diff
@@ -205,6 +205,21 @@ def test_search_parents(self, tmpdir):
             old.chdir()
         assert config.toxinipath == toxinipath
 
+    def test_explicit_config_path(self, tmpdir):
```
The-Compiler: This could be a parametrized test FWIW

tox/config.py
```diff
@@ -219,8 +219,11 @@ def parseconfig(args=None, plugins=()):
 
     # parse ini file
     basename = config.option.configfile
-    if os.path.isabs(basename):
+    if os.path.isfile(basename):
```
The-Compiler: Doesn't this change the behavior if someone has a symlink as the config file?

obestwalter: No problem here - isfile also tru if it is a symlink: https://docs.python.org/2/library/os.path.html#os.path.isfile

tests/test_config.py
pawelad: I actually tried that at some point locally but didn't find a nice enough way, as we need line 212 and 213 in both runs anyway. Do you have any suggestions?

That said - I'm happy to try that again this evening.

The-Compiler: Ah, right, I missed that.

nicoddemus:
LGTM as well, the only thing I suggest is to add something to [the help](https://github.com/tox-dev/tox/blob/master/tox/config.py#L322) of the `-c` flag, perhaps:

```python
help="config file name or directory with 'tox.ini' file."
```

pawelad:
@nicoddemus yup, good point - will do that this evening. Any suggestions on adding something to docs as well?

nicoddemus:
I scanned the docs and couldn't find any reference to the `-c` flag, so I think that's good enough. 👍

pawelad:
Should we think about adding it somewhere? : -)
Maybe along with other available parameters (or just paste `tox --help` somewhere)?

Either way - that's probably out of scope of this PR.

nicoddemus:
IMHO `-c` is enough. 😁

pawelad:
Sure, no worries.

pawelad:
Don't know how it works here, but it seems ready to merge.

nicoddemus:
I'm going ahead and merging this then. Thanks a lot @pawelad for the contribution!

Issue #271, opened 2016-09-17
Let `tox --listenvs` list all defined environments, not only from `envlist` in `[tox]` section


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/271
- Originally reported by: [vlcinsky](http://bitbucket.org/vlcinsky)
- Originally created at: 2015-08-28T00:01:52.447

Issue #52 asks for having a command to list all configured environments.

However, I found, that it lists only those environments, which are in `envlist`.

If I do not want to run some of the environments when `tox` is invoked without any argument, I keep the `envlist` in `[tox]` shorter. However, this does not print it out when::

```
$ tox -l
```

It would be nice to list all existing environments.

tox version: 2.1.1

OS: Debian Jessie

Python 2.7.9

Sample `tox.ini`:

```
#!ini

# Tox (http://tox.testrun.org/) is a tool for running tests
# in multiple virtualenvs. This configuration file will run the
# test suite on all supported python versions. To use it, "pip install tox"
# and then run "tox" from this directory.

[tox]
envlist = help
skipsdist = true

[testenv:help]
basepython = python2.7
commands = python -c "print(open('tox.help').read());" 
deps =

[testenv:docs]
basepython = python2.7
whitelist_externals =
    make
changedir = docs
commands = make html
deps = 
    -rrequirements-docs.txt

[testenv:testsamples]
basepython = python2.7
envdir = {toxworkdir}/test
commands = py.test -sv tests/test_samples.py
deps =
    -rrequirements-json.txt

[testenv:testschema]
basepython = python2.7
envdir = {toxworkdir}/test
commands = py.test -sv tests/test_schema.py
deps =
    -rrequirements-json.txt

[testenv:testyaml]
basepython = python2.7
envdir = {toxworkdir}/test
commands = py.test -sv tests/test_sample.py --yamlsample []
deps =
    -rrequirements-json.txt

[testenv:testjson]
basepython = python2.7
envdir = {toxworkdir}/test
commands = py.test -sv tests/test_sample.py --jsonsample []
deps =
    -rrequirements-json.txt

```

For given `tox.ini` it lists only `help`:

```
#!bash

$ tox --listenvs
help
```

and I would like to see there all:

```
#!bash

$ tox --listenvs
help
docs
testsamples
testschema
testyaml
testjson
```

volans-:
+1, how a user could _discover_ an environment not run by default?

obestwalter:
[...]

Issue #462, opened 2017-02-24
option to display a short documentation with envlist


gaborbernat:
Currently:

```bash
tox -l
py27
pylint
flake8
doc
```
Would it be possible to also display a short more verbose description for envs? E.g.:
```bash
tox -l
py27    run tests against Python 2.7 via pytest with coverage report
pylint   check code style with pylint
flake8  check code style with flake8
doc      generate documentation to build/doc with Sphinx
```

Thanks!

obestwalter:
@gaborbernat interesting suggestion.

Just some remarks: it would have to be implemented in a way that it is optional and it would also have to work with [generated envs](http://tox.readthedocs.io/en/latest/config-v2.html#example-generating-and-selecting-variants).

The-Compiler:
Something like this maybe?

```ini
[descriptions]
py36 = foo bar
pylint = description here
```

gaborbernat:
@obestwalter totally agree. Naturally it can fallback to an empty description (same as we have now). And for the generated envs we probably just need to make possible env variables are usable in description too, not?

obestwalter:
> we probably just need to make possible env variables are usable in description too, not?

Just to make sure, we don't misunderstand: I was not talking ebout environment variables, but about generated environments like described in the link. The names get generated by combining multiple factors together and I don't know how and if to integrate that into what you want to do, so I just mentioned it that it needs to be taken into account.

e.g. if you have:

    envlist={py27,py35}-{pylint,tests}

you will have four environments named `py27-pylint, py35-pylint, py27-tests, py35-tests` - how would you deal with descriptions for things like that?

The-Compiler:
@obestwalter You could do so with my proposal above. The syntax could even allow for e.g. `*-pylint` as key.

hpk42:
@The-Compiler 's example would describe factors and for displaying the description, we'd concatenate all descriptions in the same order as the env name consists of factors.

asottile:
@The-Compiler I'd suggest `[tox-descriptions]` to keep it reasonably namespaced (other tools read from `tox.ini` as well).  Definitely like the idea though ++

gaborbernat:
I will probably try to do this in the next weeks, feel free to assign it to me.

obestwalter:
Thanks @gaborbernat - this will be in 2.7.0 coming out next friday if nothing bad happens :)

PR #489, opened 2017-03-21
display a short documentation with the -l/ show envlist (a concatenation of the environments description) 

gaborbernat:
This is the implementation for https://github.com/tox-dev/tox/issues/462. 

For example:
```ini
[tox]
envlist={py27,py36}-{windows,linux}
[testenv]
description= py27: run py.test on Python 2.7
             py36: run py.test on Python 3.6
             windows: on Windows platform
             linux: on Linux platform
             docs: generate documentation
commands=py.test {posargs}

[testenv:docs]
changedir = docs
```

```bash
tox -l 
py27-windows run py.test on Python 2.7 on Windows platform
py27-linux   run py.test on Python 2.7 on Linux platform
py36-windows run py.test on Python 3.6 on Windows platform
py36-linux   run py.test on Python 3.6 on Linux platform
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.vagrant
 *.pyc
 *.pyo
 *.swp
--- a/tox/session.py
+++ b/tox/session.py
@@ -627,8 +627,10 @@ class Session:
                                  % (attr.name, getattr(envconfig, attr.name)))
 
     def showenvs(self):
+        max_length = max(len(env) for env in self.config.envlist)
         for env in self.config.envlist:
-            self.report.line("%s" % env)
+            self.report.line("{0} {1}".format(env.ljust(max_length),
+                                              self.config.envconfigs[env].description).strip())
 
     def info_versions(self):
         versions = ['tox-%s' % tox.__version__]
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,6 +1,7 @@
 
 contributions:
 
+Bernat Gabor
 Krisztian Fekete
 Marc Abramowitz
 Aleaxner Schepanovski
--- a/doc/config.txt
+++ b/doc/config.txt
@@ -337,6 +337,11 @@ Complete list of settings that you can put into ``testenv*`` sections:
     For example, ``extras = testing`` is equivalent to ``[testing]`` in a
     ``pip install`` command.
 
+.. confval:: description=SINGLE-LINE-TEXT
+
+    a short description of the environment, this will be used to explain
+    the environment to the user upon listing environments for the command
+    line. **default**: empty string
 
 Substitutions
 -------------
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1995,6 +1995,31 @@ class TestCmdInvocation:
             *docs*
         """)
 
+    def test_listenvs_description(self, cmd, initproj):
+        initproj('listenvs', filedefs={
+            'tox.ini': '''
+            [tox]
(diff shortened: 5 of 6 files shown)


tox/session.py
```diff
@@ -627,8 +627,9 @@ def showconfig(self):
                                  % (attr.name, getattr(envconfig, attr.name)))
 
     def showenvs(self):
+        max_length = max(len(env) for env in self.config.envlist)
         for env in self.config.envlist:
-            self.report.line("%s" % env)
+            self.report.line("{} {}".format(env.ljust(max_length), self.config.envconfigs[env].description))
```
The-Compiler: This should be `{0} {1}` to work on Python 2.6 - it seems like it doesn't get caught in the tests though, and they report "can not run tests involving calling tox on python2.6. (and python2.6 is about to be deprecated anyway)"...

gaborbernat: @The-Compiler indeed, fixed that, thanks!

gaborbernat:
@The-Compiler , @obestwalter now that the CI checks passed, can you guys take a look? thanks!

The-Compiler:
LGTM (though I don't know tox' code very well), but I'll let @obestwalter press the button :wink:

nicoddemus:
LGTM as well, but I feel we should add docs about this new feature first.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.vagrant
```
obestwalter: ha - nice to see it's being used already - does not really belong to this PR, but won't hurt anyone to just add it now, I guess :)

obestwalter:
@The-Compiler what an honour :D - I agree with  @nicoddemus that this deserves some documentation (somewhere [here-ish](http://tox.readthedocs.io/en/latest/config.html)) before we call this done.

gaborbernat:
@obestwalter hmm, I thought that's automatically taken during documentation generation from here:

```
 parser.add_testenv_attribute(
        name="description", type="string", default='', postprocess=merge_description,
        help="short description of this environment")
```

Where is that shown otherwise?

The-Compiler:
When doing `tox --help`

gaborbernat:
@obestwalter , @The-Compiler done :+1:

obestwalter:
Played around with it locally a bit and it looks good. Looking forward to the next step that fixes #271.

The-Compiler:
I can't get this to work with the 2.7.0 preview with [this tox.ini](https://github.com/qutebrowser/qutebrowser/blob/master/tox.ini). Both `description = Hello World` in `[testenv:py36-cov]` (which is what I'd expect to work by reading the docs) and `description = py36-cov: Hello World` in `[testenv]` (which seems odd, but what the tests do) still gives me no description in the env list.

obestwalter:
o.k. just tried the unconditional approach with 2.7.0 from devpi with this ini:

    [tox]
    envlist={py27,py36}-{windows,linux}
    [testenv]
    description= hello world

    [testenv:docs]
    description = blablabla


     $ tox -l

    py27-windows hello world
    py27-linux   hello world
    py36-windows hello world
    py36-linux   hello world

    $ tox -l -e docs
    docs blablabla

So from how I understand it, it works like ti should. Am I missing something?

The-Compiler:
Oh, damn, my bad! I was running my systemwide tox instead of the one from the virtualenv... nevermind! 😊

gaborbernat:
@obestwalter cool, submitted the PR for #271 in #495

Issue #276, opened 2016-09-17
remove easy_install


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/276
- Originally reported by: [thomas-guettler](http://bitbucket.org/thomas-guettler)
- Originally created at: 2015-10-09T20:37:23.467

AFAIK easy_install is deprecated. It confuses new users.

Please remove it from this page:

http://tox.readthedocs.org/en/latest/install.html

obestwalter:
If there are no objections we can actually through it out slowly I guess ...

sifuraz:
I'll take this.

sifuraz:
@obestwalter You might want to close this 🙂

PR #510, opened 2017-04-21
remove easy_install from documentation (issue #276)

sifuraz:
Original issue text:

AFAIK easy_install is deprecated. It confuses new users.
Please remove it from this page:
http://tox.readthedocs.org/en/latest/install.html

### DIFF
--- a/doc/example/basic.txt
+++ b/doc/example/basic.txt
@@ -151,12 +151,7 @@ By default tox uses `pip`_ to install packages, both the
 package-under-test and any dependencies you specify in ``tox.ini``.
 You can fully customize tox's install-command through the
 testenv-specific :confval:`install_command=ARGV` setting.
-For instance, to use ``easy_install`` instead of `pip`_::
-
-    [testenv]
-    install_command = easy_install {opts} {packages}
-
-Or to use pip's ``--find-links`` and ``--no-index`` options to specify
+For instance, to use pip's ``--find-links`` and ``--no-index`` options to specify
 an alternative source for your dependencies::
 
     [testenv]
--- a/doc/config.txt
+++ b/doc/config.txt
@@ -99,7 +99,7 @@ Complete list of settings that you can put into ``testenv*`` sections:
     Must contain the substitution key
     ``{packages}`` which will be replaced by the packages to
     install.  You should also accept ``{opts}`` if you are using
-    pip or easy_install -- it will contain index server options
+    pip -- it will contain index server options
     such as ``--pre`` (configured as ``pip_pre``) and potentially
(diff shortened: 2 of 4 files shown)


obestwalter:
Hey @sifuraz, thanks. I just did a quick serach though and there are quite a lot of hits for easy_install. Did you check them all, if they apply or need to be kept (for histrical or whatever other reasons)?

https://github.com/tox-dev/tox/search?utf8=%E2%9C%93&q=easy_install&type=

doc/example/basic.txt
```diff
@@ -151,12 +151,7 @@ By default tox uses `pip`_ to install packages, both the
 package-under-test and any dependencies you specify in ``tox.ini``.
 You can fully customize tox's install-command through the
 testenv-specific :confval:`install_command=ARGV` setting.
-For instance, to use ``easy_install`` instead of `pip`_::
-
-    [testenv]
-    install_command = easy_install {opts} {packages}
-
```
obestwalter: We should not simply delete that but use a different example. I wish we could use conda already, so I am a bit stumped for a good example. pip with different parameters would be something ... or something far out like:

    install_command = sudo apt install {opts} {packages}

which is obviously crazy but shows that you can stick everything in there, whether it makes sense or not.

sifuraz: @obestwalter Should I go with your example for now then?

obestwalter: hmm ... no better not - I am good at having really bad ideas. Let's let this sit a bit and see if something else comes up ...

sifuraz: As for pip with different paramters, there's example just below the removed part.
``` 
    [testenv]
    install_command = pip install --pre --find-links http://packages.example.com --no-index {opts} {packages}```

obestwalter: ah, o.k. - maybe that is actually enough then ...

sifuraz:
@obestwalter Hello, I see this is still in "changes requested" state, is there something else you would like me to change?

obestwalter:
@sifuraz this is out of scope of this PR, but we really have no proper docs for the installer comand in the actual config docs. So deleting that big chunk gets rid of the only piece of docs for this. I'll have a closer look and merge your PR then, so there is nothing else needed for this particular PR. Thanks!

obestwalter:
Also: the config-v2.txt should not be changed anyway as it's a "historic" document :) - but don't worry about it :)

obestwalter:
Unless you want to fix up your PR yourself, please revert config-v2.txt - cheers.

obestwalter:
flake8 is broken on appveor - has nothing to do with PR.

PR #512, opened 2017-04-27
Also catch ENOTDIR when initializing logdir

LordGaav:
There is an implementation difference between CPython and Jython. On
CPython, only ENOENT is raised when the directory does not exist
(see posixmodule.c for instance). On Jython, ENOTDIR is raised when
the directory does not exist, and EACCESS is raised when there is
no read access.

This patch also catches ENOTDIR, but not EACCESS. The code flow
ensures that the logdir is created when it does not exist.

Ref #326 .

Thanks for contributing a PR!

- [X] Add yourself to `CONTRIBUTORS`;
- [X] make a descriptive Pull Request text (it will be used for changelog)

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -44,3 +44,4 @@ Josh Smeaton
 Paweł Adamczak
 Oliver Bestwalter
 Selim Belhaouane
+Nick Douma
--- a/tox/session.py
+++ b/tox/session.py
@@ -109,7 +109,7 @@ class Action(object):
             logdir = self.session.config.logdir
         try:
             l = logdir.listdir("%s-*" % actionid)
-        except py.error.ENOENT:
+        except (py.error.ENOENT, py.error.ENOTDIR):
             logdir.ensure(dir=1)
             l = []
         num = len(l)


LordGaav:
I have looked for tests, but I didn't find an existing test to extend with the extra catch. Please let me know if I need to add a (new/separate) test for this, or what existing one to extend.

peterjc:
Could this be merged please? I would love to be able to use Jython under tox for continuous integration testing.

obestwalter:
@peterjc  thanks for the patch. I guess we can merge this and make a small release soonish.

peterjc:
Thank you. Note credit belongs to @LordGaav for the patch - thank you too 👍

obestwalter:
ups thanks @LordGaav :D

peterjc:
Is it worth adding this to the ``CHANGELOG`` ready for the next release?

nicoddemus:
> Is it worth adding this to the CHANGELOG ready for the next release?

👍

obestwalter:
@nicoddemus I start thinking that we could also start using towncrier. I was out of action for a while an now the changelog merge conflicts start to pile up :)

nicoddemus:
@obestwalter highly recommended! 😁

PR #520, opened 2017-05-18
Forward NUMBER_OF_PROCESSORS by default on Windows

### LINKED ISSUE #517: Foward NUMBER_OF_PROCESSORS environment variable by default
Hi there!

In one project, I have a Python 2.7-only dependency that uses [`multiprocessing.cpu_count()`](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.cpu_count) on Windows.  This has a hidden dependency on the `NUMBER_OF_PROCESSORS` environment variable.  This is quite confusing when porting this type of project into Tox.

**Note**: in Python 3.4+, this is API is superseeded by [`os.cpu_count()`](https://docs.python.org/3/library/os.html#os.cpu_count) which does not depend on this environment variable.

The case is a bit weak (Python up to 3.3 only, plus multiprocessing is not popular on Windows), but I feel like the standard library should work by default inside a Tox test environment.

I'm willing to work on a patch if you would accept such a contribution!

AndreLouisCaron:
This fixes `multiprocessing.cpu_count()` on Windows for Python versions prior
to 3.4.

Fixes #517.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,8 @@ Not released yet
 
 - #474: Start using setuptools_scm for tag based versioning.
 - #506: With `-a`: do not show additional environments header if there are none
+- #517: Forward NUMBER_OF_PROCESSORS by default on Windows to fix
+        `multiprocessor.cpu_count()`.
 
 2.7.0
 -----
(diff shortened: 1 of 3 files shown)


obestwalter:
Don't know why the PR checks failed, but master is green again with the merge, so should be o.k.

PR #521, opened 2017-05-18
Isolate environment variable substitutions

### LINKED ISSUE #515: Environment variables required across unrelated environments
Hi there,

I have an issue where if one environment uses an `{env:X}` subsitution (without default), then that environment variable is required in all environments regardless of whether they are necessary or not.

Given this example `tox.ini` file:

```ini
[tox]
skipsdist = true

[testenv:standard-greeting]
commands =
  python -c 'print("Hello, world!")'

[testenv:custom-greeting]
passenv =
  NAME
commands =
  python -c 'print("Hello, {env:NAME}!"'
```

Then `custom-greeting` environment cannot be used unless `NAME` is present in the calling environment, even if that environment doesn't use the `{env:NAME}`.

```shell
$ tox -e standard-greeting
tox.ConfigError: ConfigError: substitution env:'NAME': unknown environment variable 'NAME'  or recursive definition.
```

I got this with tox 2.7.0 with both Python 3.5 and 2.7 on Windows.

AndreLouisCaron:
Perform `{env:X}` replacement only in environments where `X` is used.  This
enables different environments to use different subsets of the environment
(e.g. don't require union of all passenv sections in all environments).

Fixes #515.

**Note to reviewers**: this is a tentative fix.  I'm definitely not sure how to best proceed with this fix.  Feel free to recommend an alternate approach!

### DIFF
--- a/tox/config.py
+++ b/tox/config.py
@@ -787,7 +787,8 @@ class parseini:
             factors = set(name.split('-'))
             if section in self._cfg or factors <= known_factors:
                 config.envconfigs[name] = \
-                    self.make_envconfig(name, section, reader._subs, config)
+                    self.make_envconfig(name, section, reader._subs, config,
+                                        replace=name in config.envlist)
 
         all_develop = all(name in config.envconfigs
                           and config.envconfigs[name].usedevelop
@@ -803,7 +804,7 @@ class parseini:
                 factors.update(*mapcat(_split_factor_expr, exprs))
         return factors
 
-    def make_envconfig(self, name, section, subs, config):
+    def make_envconfig(self, name, section, subs, config, replace=True):
         factors = set(name.split('-'))
         reader = SectionReader(section, self._cfg, fallbacksections=["testenv"],
                                factors=factors)
@@ -818,7 +819,7 @@ class parseini:
             atype = env_attr.type
             if atype in ("bool", "path", "string", "dict", "dict_setenv", "argv", "argvlist"):
                 meth = getattr(reader, "get" + atype)
-                res = meth(env_attr.name, env_attr.default)
+                res = meth(env_attr.name, env_attr.default, replace=replace)
             elif atype == "space-separated-list":
                 res = reader.getlist(env_attr.name, sep=" ")
             elif atype == "line-list":
@@ -937,9 +938,9 @@ class SectionReader:
         if _posargs:
             self.posargs = _posargs
 
-    def getpath(self, name, defaultpath):
+    def getpath(self, name, defaultpath, replace=True):
         toxinidir = self._subs['toxinidir']
-        path = self.getstring(name, defaultpath)
+        path = self.getstring(name, defaultpath, replace=replace)
         if path is not None:
             return toxinidir.join(path, abs=True)
 
@@ -949,12 +950,12 @@ class SectionReader:
             return []
         return [x.strip() for x in s.split(sep) if x.strip()]
 
-    def getdict(self, name, default=None, sep="\n"):
+    def getdict(self, name, default=None, sep="\n", replace=True):
         value = self.getstring(name, None)
         return self._getdict(value, default=default, sep=sep)
 
-    def getdict_setenv(self, name, default=None, sep="\n"):
-        value = self.getstring(name, None, replace=True, crossonly=True)
+    def getdict_setenv(self, name, default=None, sep="\n", replace=True):
+        value = self.getstring(name, None, replace=replace, crossonly=True)
         definitions = self._getdict(value, default=default, sep=sep)
         self._setenv = SetenvDict(definitions, reader=self)
         return self._setenv
@@ -971,8 +972,8 @@ class SectionReader:
 
         return d
 
-    def getbool(self, name, default=None):
-        s = self.getstring(name, default)
+    def getbool(self, name, default=None, replace=True):
+        s = self.getstring(name, default, replace=replace)
         if not s:
             s = default
         if s is None:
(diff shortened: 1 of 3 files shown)


The-Compiler:
Can't speak for the code, but I've had this problem before too.

AndreLouisCaron:
Ping?

obestwalter:
Hi @AndreLouisCaron thanks for the fix. Can you quickly tell me why you are not confident that the fix is the best way to go forward? I did not check the code yet, just wanted to get more insight, before I review it.

AndreLouisCaron:
@obestwalter Some methods already had a `replace` argument, but not all of them.  Also, it seemed (to me, without context) to be inconsistently applied.  I couldn't figure out why it was applied in some places and not others, but maybe it was written like that for a good reason that is unknown to me :-)

obestwalter:
o.k. thanks - I will take that into account and dig a bit through history to check whether there is some background to it.

tox/config.py
```diff
@@ -818,7 +819,7 @@ def make_envconfig(self, name, section, subs, config):
             atype = env_attr.type
             if atype in ("bool", "path", "string", "dict", "dict_setenv", "argv", "argvlist"):
                 meth = getattr(reader, "get" + atype)
-                res = meth(env_attr.name, env_attr.default)
+                res = meth(env_attr.name, env_attr.default, replace=replace)
```
AndreLouisCaron: **Note to reviewers**: adding this here is why I needed to add a `replace` argument all over the place.

tox/config.py
```diff
@@ -787,7 +787,8 @@ def __init__(self, config, inipath):
             factors = set(name.split('-'))
             if section in self._cfg or factors <= known_factors:
                 config.envconfigs[name] = \
-                    self.make_envconfig(name, section, reader._subs, config)
+                    self.make_envconfig(name, section, reader._subs, config,
+                                        replace=name in config.envlist)
```
AndreLouisCaron: **Note to reviewers**: this is the core of the change I wanted to make.  Don't replace environment substitutions unless the environment is in `envlist`.

tox/config.py
```diff
@@ -1171,31 +1172,34 @@ def getargvlist(cls, reader, value):
...
+                    else:
+                        word = word[9:-1]
+                new_arg = ""
+                new_word = reader._replace(word)
+                new_word = reader._replace(new_word)
+                new_word = new_word.replace('\\{', '{').replace('\\}', '}')
+                new_arg += new_word
+                newcommand += new_arg
+        else:
+            newcommand = command
```
AndreLouisCaron: **Note to reviewers**: the new `else` clause here is essential to the change.  When `replace=False`, we should leave the command uninterpreted (we can't evaluate it since we're potentially missing some of the variables).

AndreLouisCaron:
@obestwalter I added a few notes in the review to cue in on my reasoning.  I hope it helps without skewing your independent review :-)

obestwalter:
Thanks, very appreciated :)

tox/config.py
AndreLouisCaron: Actually, my intent is to use `replace=False` when the environment is not scheduled in the current run.  I'm not 100% sure `config.envlist` is what I want here.

tox/config.py
```diff
@@ -953,12 +954,12 @@ def getlist(self, name, sep="\n"):
             return []
         return [x.strip() for x in s.split(sep) if x.strip()]
 
-    def getdict(self, name, default=None, sep="\n"):
+    def getdict(self, name, default=None, sep="\n", replace=True):
         value = self.getstring(name, None)
```
AndreLouisCaron: **Note to self**: may have forgotten the `replace=replace` here.

obestwalter:
I'll review and test this at the EuroPython tox sprint weekend and then it will likely be in 2.8.

AndreLouisCaron:
Thanks a bunch!

tox/config.py
AndreLouisCaron: Yes, that's what I meant.

tox/config.py
AndreLouisCaron: I'm not sure I understand your question.  IMO, if you request to run an environment and we can't perform substitutions because something is missing, I would keep the existing behaviour (e.g. faily loudly) for the moment.

I agree that as a user, I would expect something else (e.g. run other envs, skip/fail the one(s) with the missing vars), but I would do that in a separate PR.

tox/config.py
```diff
@@ -1188,31 +1189,34 @@ def getargvlist(cls, reader, value):
...
         # shlex.
-        newcommand = ""
-        for word in CommandParser(command).words():
-            if word == "{posargs}" or word == "[]":
-                newcommand += posargs_string
-                continue
-            elif word.startswith("{posargs:") and word.endswith("}"):
-                if posargs:
+        if replace:
+            newcommand = ""
```
gaborbernat: this should be ``new_command`` to be consistent with other variable names used around it

AndreLouisCaron:
@obestwalter Can't reply directly to your comment.

>> Actually, my intent is to use replace=False when the environment is not scheduled in the current run.

> Do you mean: all the runs scheduled in this invocation? e.g. tox -e py27,py34 would then only do substitutions for these two environments when parsing the ini?

Yes.

AndreLouisCaron:
> In combination with the current way of determining if replacements should happen that would mean though that it is possible that an environment that should run does not get any replacements if one of the variables is not set instead of rightfully crashing because there is a variable missing. Correct? I think I have create some example tox.ini and play this through to get my head around it though ...

Like I said, I'd like to answer, but I'm not sure I understand what your question is.

tox/config.py
obestwalter: oh .. I was not notified that you answered here :)

Actually I also prefer failing, but I am not sure if this is happening. I have to look closer at the test and do some exploring myself. Thanks.

obestwalter:
o.k. - I'll merge this now and fix up anything that is not quite there yet (e.g. one missing passing of replace). Will test this today and cut  a release candidate from it at some point at the weekend. @AndreLouisCaron it would be great if you test the rc then yourself a bit and check if everything is as it should be.

AndreLouisCaron:
Sorry, was on vacation the past 2 weeks and did not follow up.  Thanks for proceeding with this!  Let me know when the RC is up, I will gladly try it out :-)

obestwalter:
@AndreLouisCaron - hope you had a nice holiday :) rc1 with your changes is out already.

Issue #499, opened 2017-04-02
Use descriptions on tox.ini


obestwalter:
From 2.7.0 upwards you can set descriptions for the tox environments to be shown with `-lv` or `-av`.

The tox project itself should make use of the descriptions in its own `tox.ini`.

gaborbernat:
I'll take this.

PR #526, opened 2017-05-20
document environments for projects tox.ini

gaborbernat:
- also now all config values can be substituted (so e.g. basepython can be used inside the description)
- ignore IDEA project files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.orig
 *.rej
 *~
+__pycache__
 
 build
 dist
@@ -14,3 +15,7 @@ doc/_build/
 tox.egg-info
 .tox
 .cache
+
+.idea
+.eggs/
+py27/
--- a/tox.ini
+++ b/tox.ini
@@ -1,49 +1,50 @@
 [tox]
-envlist=py27,py26,py34,py33,py35,py36,pypy,flakes,py26-bare
+envlist = py27,py26,py34,py33,py35,py36,pypy,flakes,py26-bare
+minversion = 2.7.0
 
 [testenv:X]
-commands=echo {posargs}
+commands = echo {posargs}
+description = print the positional arguments passed in with echo
 
 [testenv]
-commands= pytest --timeout=180 {posargs:tests}
-
-deps=pytest>=3.0.0
-    pytest-timeout
+commands = pytest --timeout=180 {posargs:tests}
+description = run the unit tests with PyTest under the current Python env
+deps = pytest >= 3.0.0
+       pytest-timeout
 
 [testenv:py26-bare]
 deps =
 commands = tox -h
 install_command = pip install {opts} {packages}
 list_dependencies_command = pip freeze
+description = invoke the tox help message under Python 2.6
 
 [testenv:py26]
 install_command = pip install {opts} {packages}
 list_dependencies_command = pip freeze
 
-
 [testenv:docs]
-basepython=python
-changedir=doc
-deps=sphinx
-     {[testenv]deps}
-commands=
-    pytest -v check_sphinx.py {posargs}
+basepython = python
+changedir = doc
+deps = sphinx
+       {[testenv]deps}
+commands = pytest -v check_sphinx.py {posargs}
+description = invoke sphinx-build and try to build the HTML page (also check link validity)
 
 [testenv:flakes]
-deps = pytest-flakes>=0.2
+deps = pytest-flakes >= 0.2
        pytest-pep8
-
-commands =
-    pytest --flakes -m flakes tox tests
-    pytest --pep8 -m pep8 tox tests
+description = run static analysis and style check using flakes and pep-8
+commands = pytest --flakes -m flakes tox tests
+           pytest --pep8 -m pep8 tox tests
 
 [testenv:dev]
 # required to make looponfail reload on every source code change
 usedevelop = True
 
-deps =
-      pytest-xdist>=1.11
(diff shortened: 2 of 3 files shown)


gaborbernat:
@obestwalter @The-Compiler any reasons why we don't run the unit tests with the project itself? (and instead install a dependency that runs it, basically guarantying that we run the tests with an older version of tox?)

obestwalter:
@gaborbernat no, no idea, but just a hunch: make sure that tox is always tested with a stable version of tox.

gaborbernat:
so how should we handle this then? 

A. bump the tests to use the latest
B. keep testing with an older version:  remove substitution of basepython for now, and derail it until we release a new tox version

nicoddemus:
IMHO it makes sense to have at least one job in Travis and AppVeyor where the tests are run "in-place" instead of using `tox` itself.

obestwalter:
Any idea, why all tests fail with

    ERROR: tox version is 2.7, required is at least 2.7.1.dev13

?

The-Compiler:
Because the PR added `minversion=2.7.1.dev13` - see the comments above yours.

obestwalter:
It shouldn't then.

gaborbernat:
@obestwalter but it needs that to pass because the tox.ini uses a feature added in this version/PR. hence my earlier question here:

[@obestwalter](https://github.com/obestwalter) [@The-Compiler](https://github.com/the-compiler) any reasons why we don't run the unit tests with the project itself? (and instead install a dependency that runs it, basically guarantying that we run the tests with an older version of tox?)

obestwalter:
Hi @gaborbernat - sorry, I was a bit too dense yesterday. What I meant, was that we just can't do it this way at the moment. We have to add the change that make substitutions possible now and release that and then add the docs. So for now the descriptions in the tox.ini can't contain substitutions on from 2.8+ they can. That would be my preferred path to sneak that in.

So if you update the PR that way (please have a look at my other comments) with 2.8+ the use of substitutions will be possible.

We can definitely discuss the way how we test tox and how we can improve it, but I think this should happen in a separate issue or on the tox-dev mailing list (or IRL next week at EuroPython :)). There usually are downsides to any such change and they need to be considered first and also communicated to the other contributors. So this would be something that we can tackle after 2.8 is out of the way.

obestwalter:
Adding missing issue ref: https://github.com/tox-dev/tox/issues/499

obestwalter:
Sorry: s/then add the docs/add the substitutions in tox.ini/

gaborbernat:
Hello, I'll amend it tomorrow 👍 sadly my company is not supporting me yet to attend EuroPython. But yes we can discuss how this is handled. 

In a more general I would like to tackle the project review/ownership how it works. This PR has stayed in limbo for a month until you got some time on your hand, and I do understand it's summer and people have lot on their hand. As I stated earlier I would like to contribute but with this rate of response  even simple changes such as this takes a significant amount of time that discourages me to do so.

obestwalter:
Hi @gaborbernat, I understand that it is sometimes frustrating that things take longer in OSS projects but that's just how it is: there is always significantly more work than people who are willing and able to do it. And the work involves so much more than writing the code. Just for comparison: my own PR about starting to use setuptools_scm was open for more than a month and it is still not released. I don't consider that particularly slow, it's just how it is. If you look at other projects there are tons of unanswered issues and PRs that are open for more than a year, so I really think we're doing alright here.

It is also a question of priorities. One "problem" that tox has, is that a lot of projects rely on it in their build chain, so my philisophy since I somehow took over release management was to provide stable releases that don't introduce new bugs, rather than quick releases that might break thousands of CI builds. 

As I still have no intimate knowledge of the code base yet, I might be rather on the conservative side as it is hard for me to evaluate the impact of certain changes. So, I for one am also a bit frustrated at the moment, because I almost entirely served as a project manager/release manager/pr reviewer/issue triager since I joined the project and if this is the only thing you do (as a developer), this gets old quite quickly. Therefore I would also really like to tackle some issues in the code myself and do less of the other stuff, but I want to put that on a better footing by improving the way we organize the project first. I started writing things down this weekend and will open the discussion about that soon. I also opened issues about code coverage and release candidates to help speed up releases.

If you want to join the project and also help with issue traiging, reviews and testing, let us know. I would be thrilled to have more active people in the project, as the "old" maintainers are very busy with their other projects, so tox was a bit on the back burner for quite a while, I guess. One caveat though: you would still have to open PRs for non trivial changes and you still might have to wait just like now. 

Talking prioritization: in the present phase of the project it would be important to fix known bugs (see https://github.com/tox-dev/tox/projects/1), improve existing behaviour (e.g. https://github.com/tox-dev/tox/projects/3) and have a good look at the internal structure and do some refactoring to keep the code base maintainable. This is what I want to concentrate on in the next months, so I might even get more conservative when it's about introducing new features, unless many people ask for them. On the other hand: I really like the introduction of descriptions that you initiated and I use them in my own projects, so thanks a lot for that :)

gaborbernat:
@obestwalter thanks for the detailed outline, sounds like a good plan!

PS. As for the change sorry for being late, but my laptop died on me, and it will be a few more days until I get a replacement.

obestwalter:
Wow - thats bad. Hope nothing important got lost. Take your time - patience goes both ways :)

gaborbernat:
@obestwalter done now :+1:

obestwalter:
Thanks @gaborbernat  :)

PR #535, opened 2017-06-21
Use $XDG_CACHE_HOME by default for distshare

### LINKED ISSUE #346: Tox should adhere to the XDG BaseDir standard on Linux and possibly OS X
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/346
- Originally reported by: [SShrike](http://bitbucket.org/SShrike)
- Originally created at: 2016-07-14T23:40:36.512

You can find the specification [here](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

This [blog post](https://ploum.net/207-modify-your-application-to-use-xdg-folders/) lists the advantages of using the XDG Base Directory Specification as follows:
- `$HOME` is a lot less cluttered.
- Backups are a lot more safer and easier (you know that creating a backup of your `$XDG_DATA_HOME` along with your files is enough).
- A lot easier to reset a default configuration if you want/need it (and without any risk of losing information).
- Avoid some strange bugs that happen because you had a old version of some configuration file.
- It is a lot more flexible and portable because no paths are hard-coded. You can use the XDG library that does the job for you _or_, if you don’t want the dependency, implementing the XDG specification is only a few lines of code.

For example, what is currently `~/.tox/distshare` would be `~/.cache/tox/distshare` or something similar.

jleclanche:
Closes #346

https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

This defaults distshare to $XDG_CACHE_HOME. Per the spec, if that variable is not defined, $HOME/cache is used (os.path.expanduser("~") is windows-compatible as well). All this removes extraneous clutter in $HOME.

----

### DIFF
--- a/doc/config.txt
+++ b/doc/config.txt
@@ -19,7 +19,7 @@ List of optional global options::
     toxworkdir=path   # tox working directory, defaults to {toxinidir}/.tox
     setupdir=path     # defaults to {toxinidir}
     distdir=path      # defaults to {toxworkdir}/dist
-    distshare=path    # (DEPRECATED) defaults to {homedir}/.tox/distshare
+    distshare=path    # (DEPRECATED) defaults to $XDG_CACHE_HOME/tox/distshare
     envlist=ENVLIST   # defaults to the list of all environments
     skipsdist=BOOL    # defaults to false
 
--- a/doc/example/general.txt
+++ b/doc/example/general.txt
@@ -102,8 +102,10 @@ that another tox run can find the "latest" dependency.  This feature
 allows to test a package against an unreleased development version
 or even an uncommitted version on your own machine.
 
-By default, ``{homedir}/.tox/distshare`` will be used for
-copying in and copying out artifacts (i.e. Python packages).
+By default, the distshare directory is located in
+``$XDG_CACHE_HOME/tox/distshare``, where ``$XDG_CACHE_HOME`` defaults to
+``{homedir}`` if not set. This directory will be used for copying in and
+copying out artifacts (i.e. Python packages).
 
 For project ``two`` to depend on the ``one`` package you use
 the following entry::
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 Not released yet
 ----
 
+- #346: Use $XDG_CACHE_HOME by default for distshare
 - #474: Start using setuptools_scm for tag based versioning.
 - #506: With `-a`: do not show additional environments header if there are none
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -46,3 +46,4 @@ Paweł Adamczak
 Oliver Bestwalter
 Selim Belhaouane
 Nick Douma
+Jerome Leclanche
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -178,7 +178,13 @@ class TestConfigPackage:
 
     def test_defaults_distshare(self, tmpdir, newconfig):
         config = newconfig([], "")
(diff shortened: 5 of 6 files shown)


doc/config.txt
```diff
@@ -19,7 +19,7 @@ List of optional global options::
     toxworkdir=path   # tox working directory, defaults to {toxinidir}/.tox
     setupdir=path     # defaults to {toxinidir}
     distdir=path      # defaults to {toxworkdir}/dist
-    distshare=path    # (DEPRECATED) defaults to {homedir}/.tox/distshare
+    distshare=path    # (DEPRECATED) defaults to $XDG_CACHE_HOME/tox/distshare
```
obestwalter: As the docs say: this feature is deprecated, so why should we put extra work into this - or is the deprecation deprecated?

jleclanche: I have no idea about the feature, but its default has been updated due to how the PR works.

jleclanche:
My understanding is that it's the distshare *config option* that is deprecated, not the concept of a distshare.

obestwalter:
That might well be - I simply don't know, as I don't use that feature, I'll dig around a bit.

obestwalter:
In the [CHANGELOG](http://tox.readthedocs.io/en/latest/changelog.html?highlight=distshare) it also looks like distshare as such is deprecated:

> DEPRECATE distshare in documentation

There is [not much about it](http://tox.readthedocs.io/en/latest/search.html?q=distshare&check_keywords=yes&area=default) - looks like the original use case was for Jenkins to share builds between different runs, and the only other mention is in tips and tricks, so I guess this is really deprecated, maybe someone else can shed some light on this ...

jleclanche:
I find that as far as personal usage goes, tox does a bit of a poor job in caching wheels and updating venvs when the requirements update (eg. on travis, caching the .tox directory is a huge speedup, but then causes build failures when the requirements are updated).

As far as distshare goes, I think it should at the very least be disabled by default if tox no longer cares about it and it's truly deprecated.

obestwalter:
> As far as distshare goes, I think it should at the very least be disabled by default if tox no longer cares about it and it's truly deprecated.

That's what I am trying to find out now. I personally don't care about it (at least not until now), but I don't know why it's deprecated either.

> I find that as far as personal usage goes, tox does a bit of a poor job in caching wheels and updating venvs when the requirements update

If you look around in the issues a bit, you'll find out that you are not the only one who noticed that and that there are several efforts to improve that.

doc/example/general.txt
```diff
@@ -102,8 +102,10 @@ that another tox run can find the "latest" dependency.  This feature
 allows to test a package against an unreleased development version
 or even an uncommitted version on your own machine.
 
-By default, ``{homedir}/.tox/distshare`` will be used for
-copying in and copying out artifacts (i.e. Python packages).
+By default, the distshare directory is located in
+``$XDG_CACHE_HOME/tox/distshare``, where ``$XDG_CACHE_HOME`` defaults to
+``{homedir}`` if not set. This directory will be used for copying in and
```
merwok: This default is not consistent with the XDG spec: “If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.”

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

obestwalter:
@jleclanche I Won't merge this for this release as #346 is actually not clarified yet, if understand it correctly (windows behaviour). Also @merwok suggested a fix that has not been integerated yet. For 2.8 there is more than enough already, so this will hopefully go into 2.9 once it's clarified.

Issue #137, opened 2016-09-17
Unexpected setenv behavior


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/137
- Originally reported by: @cjerdonek
- Originally created at: 2013-11-14T11:20:48.926

The following behavior was unexpected to me.  With an ini section like the following:

```
[testenv]
setenv = FOO = bar
commands = echo {env:FOO}
```

The environment variable `FOO` set via `setenv` is not reflected in the `{env:FOO}` substitution.  I think the documentation should clarify this because it may be surprising to some.  Also, there is a typo [here](https://bitbucket.org/hpk42/tox/src/72ba41dbebee260b19f8ef1a337c83d45ab8fcf3/tox/_config.py?at=default#cl-550) (unkown -> unknown).

obestwalter:
As the merged test from @sifuraz shows, this bug has been fixed already.

PR #547, opened 2017-07-11
Issue #137 test

sifuraz:
Hello @obestwalter, I was checking issue #137 and I've tried to replicate it, but it is passing.
Could you check out if I'm missing something? Thanks.


obestwalter:
Hi @sifuraz it looks like this has been fixed and that bug is already invalid for a long time. Your test is good though to ensure that behavior. Thank you :)

PR #548, opened 2017-07-13
Allow tests to be run through a user-defined hook.

Steap:
Currently, users can only extend/overwrite the pre/post phases. They should be able to change the way the tests are run. This would be useful to implement a GNU Guix backend, for instance.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -46,3 +46,4 @@ Paweł Adamczak
 Oliver Bestwalter
 Selim Belhaouane
 Nick Douma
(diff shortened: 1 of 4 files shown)


obestwalter:
Hi @Steap, thanks for the patch. I guess this is a sensible place for another hook. Patch looks good. That pep8 error can be fixed after merging, if you don't get round to it. If no one objects, we can merge this at the weekend and make it part of the upcoming 2.8 release.

obestwalter:
Thanks @Steap :)

Issue #203, opened 2016-09-17
tox fails when ":" is in the current directory path


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/203
- Originally reported by: [toabctl](http://bitbucket.org/toabctl)
- Originally created at: 2014-12-18T18:19:58.669

This is on linux. Trying to execute tox when the path contains a ":" doesn't work:

```
$ pwd
/home/tom/devel/bla:blub/tox
tom@basilikum:~/devel/bla:blub/tox$ tox
GLOB sdist-make: /home/tom/devel/bla:blub/tox/setup.py
Traceback (most recent call last):
  File "/usr/bin/tox", line 9, in <module>
    load_entry_point('tox==1.7.2', 'console_scripts', 'tox-2.7')()
  File "/usr/lib/python2.7/site-packages/tox/_cmdline.py", line 26, in main
    retcode = Session(config).runcommand()
  File "/usr/lib/python2.7/site-packages/tox/_cmdline.py", line 310, in runcommand
    return self.subcommand_test()
  File "/usr/lib/python2.7/site-packages/tox/_cmdline.py", line 448, in subcommand_test
    if self.setupenv(venv):
  File "/usr/lib/python2.7/site-packages/tox/_cmdline.py", line 377, in setupenv
    envlog.set_python_info(commandpath)
  File "/usr/lib/python2.7/site-packages/tox/result.py", line 47, in set_python_info
    "import sys; "
  File "/usr/lib/python2.7/site-packages/py/_path/local.py", line 695, in sysexec
    stdout, stderr,)
py.process.cmdexec.Error: ExecutionFailed: 1  /home/tom/devel/bla:blub/tox/.tox/py27/bin/python
ImportError: No module named site
```

Moving the same code to another directory and executing tox works fine.

obestwalter:
There is still a problem, but it's different now:

setup.py:

    from setuptools import setup
    setup(name='dummy-setup')

tox.ini:

    [tox]
    envlist = py27
    
    [testenv]
    commands = pip --version
    
results in:
    [....]
    ImportError: No module named site
    New python executable in /home/oliver/work/tox/tests/bla:blub/.tox/py27/bin/python2.7
    Also creating executable in /home/oliver/work/tox/tests/bla:blub/.tox/py27/bin/python
    ERROR: The executable /home/oliver/work/tox/tests/bla:blub/.tox/py27/bin/python2.7 is not functioning
[...]

PR #553, opened 2017-07-15
Add regression test for upstream virtualenv bug

bartsanchez:
https://github.com/tox-dev/tox/issues/203


obestwalter:
@bartsanchez thanks a lot. I will mark the corresponding issue as upstream bug and close it, when we have the new labels.

obestwalter:
Thanks @bartsanchez :)

Issue #121, opened 2016-09-17
White space in path causes tox to fail with "No such file or directory"


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/121
- Originally reported by: [fizista](http://bitbucket.org/fizista)
- Originally created at: 2013-08-30T02:15:31.500

As shown below, if the path to the source program are white space (in this case the spaces), the "tox" does not work and throws this error.

When I moved the source to the directory "/home/wojciech/django-money" then the "tox" works.

```
#!python

django-money$ tox
GLOB sdist-make: /home/wojciech/Aptana Studio 3 Workspace/django-money/setup.py
py27 create: /home/wojciech/Aptana Studio 3 Workspace/django-money/.tox/py27
py27 installdeps: -r"/home/wojciech/Aptana Studio 3 Workspace/django-money/requirements.txt"
Traceback (most recent call last):
  File "/home/wojciech/.pyenv/versions/2.7.5/bin/tox", line 9, in <module>
    load_entry_point('tox==1.6.0', 'console_scripts', 'tox')()
  File "/home/wojciech/.pyenv/versions/2.7.5/lib/python2.7/site-packages/tox/_cmdline.py", line 26, in main
    retcode = Session(config).runcommand()
  File "/home/wojciech/.pyenv/versions/2.7.5/lib/python2.7/site-packages/tox/_cmdline.py", line 301, in runcommand
    return self.subcommand_test()
  File "/home/wojciech/.pyenv/versions/2.7.5/lib/python2.7/site-packages/tox/_cmdline.py", line 433, in subcommand_test
    if self.setupenv(venv):
  File "/home/wojciech/.pyenv/versions/2.7.5/lib/python2.7/site-packages/tox/_cmdline.py", line 358, in setupenv
    status = venv.update(action=action)
  File "/home/wojciech/.pyenv/versions/2.7.5/lib/python2.7/site-packages/tox/_venv.py", line 142, in update
    self.install_deps(action)
  File "/home/wojciech/.pyenv/versions/2.7.5/lib/python2.7/site-packages/tox/_venv.py", line 272, in install_deps
    self._install(deps, action=action)
  File "/home/wojciech/.pyenv/versions/2.7.5/lib/python2.7/site-packages/tox/_venv.py", line 330, in _install
    extraenv=extraenv)
[...]

PR #556, opened 2017-07-15
Report more meaningful errors on why virtualenv creation failed

vlaci:
`virtualenv` fails if the destination directory contains unicode characters or spaces. A more descriptive error is printed if it's likely that that was the case.
As long as the upstream problem exists #121 cannot be fixed.

Open questions:

- [x] Should hard-fail even before executing `virtualenv` if the destination directory contains unsupported characters? -- I choosed not to in order to not block support of such paths if upstream is fixed
- [x] Should check for unsupported characters and only print error message about them if they are found? -- I decided not to, because testing this part of the code would be very hard and I don't like the idea to introduce logic without tests. I could abstract away the error reporting of venv.update() calls and test it in isolation if needed.

---

Thanks for contributing a PR!

- [ ] if an enhancement PR please create docs and at best an example
- [ ] Add yourself to `CONTRIBUTORS`;
- [ ] make a descriptive Pull Request text (it will be used for changelog)

### DIFF
--- a/tox/session.py
+++ b/tox/session.py
@@ -449,8 +449,19 @@ class Session:
             envlog = self.resultlog.get_envlog(venv.name)
             try:
                 status = venv.update(action=action)
-            except tox.exception.InvocationError:
-                status = sys.exc_info()[1]
+            except FileNotFoundError as e:
+                status = (
+                    "Error creating virtualenv. "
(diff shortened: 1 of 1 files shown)


vlaci:
#203  is also related

obestwalter:
Thanks @vlaci :)

PR #557, opened 2017-07-16
config: expand nested substitutions (fixes #301)

### LINKED ISSUE #301: {env:} macro is not expanded in setenv if the default value contains {envdir}

vlaci:
In case of a nested substitution values are expanded starting from the
innermost expression and continued until no successful substitutions
could be found.

Thanks for contributing a PR!

### DIFF
--- a/tox/config.py
+++ b/tox/config.py
@@ -1059,8 +1059,20 @@ class Replacer:
         self.reader = reader
         self.crossonly = crossonly
 
-    def do_replace(self, x):
-        return self.RE_ITEM_REF.sub(self._replace_match, x)
+    def do_replace(self, value):
+        '''
+        Recursively expand substitutions starting from the innermost expression
+        '''
+        def substitute_once(x):
+            return self.RE_ITEM_REF.sub(self._replace_match, x)
+
+        expanded = substitute_once(value)
+
+        while expanded != value:  # substitution found
+            value = expanded
+            expanded = substitute_once(value)
+
+        return expanded
 
     def _replace_match(self, match):
         g = match.groupdict()
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -48,3 +48,4 @@ Selim Belhaouane
 Nick Douma
 Cyril Roelandt
 Bartolome Sanchez Salado
+Laszlo Vasko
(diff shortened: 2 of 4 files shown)


obestwalter:
Thanks @vlaci :)

tox/config.py
```diff
@@ -1059,8 +1059,20 @@ def __init__(self, reader, crossonly=False):
         self.reader = reader
         self.crossonly = crossonly
 
-    def do_replace(self, x):
-        return self.RE_ITEM_REF.sub(self._replace_match, x)
+    def do_replace(self, value):
+        '''
```
gaborbernat: I think that the PEP style guide requires docstrings to be using double quotes - this here is just a multi-line string at the start of the function

obestwalter: The PEP8 is not very strong with this project (yet) - so I am really not bothered with things like that, when reviewing PRs. We run pep8 and flakes as part of the tests and if they are happy I am happy. 

Although PEP8 suggests double quoted strings, This still creates a valid doc string:

    >>> print(Replacer.do_replace.__doc__)
    
    Recursively expand substitutions starting from the innermost expression
 
I agree though that it would be nice, if the code base would be made more PEP8 conform.

Issue #270, opened 2016-09-17
skipsdist and usedevelop don't play well together


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/270
- Originally reported by: @aconrad
- Originally created at: 2015-08-24T23:55:17.766

I'm trying to do `usedevelop` and `skipsdist` at the same time but it feels that they are mutually exclusive with tox.

I have 2 tasks, `py34` and `pep8`. When I run py34 I want to use `usedevelop=true` but when I run `pep8` I don't want my project to be installed at all and use `skipsdist=true`.

Please correct me if I'm wrong: skipsdist is only a global setting recognized under the section `[tox]` which would then skip the sdist step for all `[testenv]` sections which in turns would ignore any `usedevelop=true` flag since it's tox will not even try to install the testenv in the first place.

Can we have `skipsdist` be recognized under a `[testenv]`?

```
#!ini

[tox]
envlist = py34, pep8

[testenv]
usedevelop = true
deps =
    -rtest-requirements.txt
commands =
    py.test {posargs:tests/}

[testenv:pep8]
skipsdist = true
basepython = python3.4
deps = flake8
commands =
    flake8 pricingsvc/ --max-line-length=100

[pytest]
addopts =
    --junitxml=junit.xml
    --cov=pricingsvc --cov-report=xml --cov-report=term-missing
    --verbose
```

obestwalter:
@aconrad can we close this as resolved then?

ferdonline:
I confirm that `usedevelop` breaks `skip_install = true`

```
(sparkenv) [leite@centos-linux-7 pyspark]$ tox -e flake8
flake8 create: /media/psf/Home/dev/Functionalizer/pyspark/.tox/flake8
flake8 installdeps: flake8
flake8 develop-inst: /media/psf/Home/dev/Functionalizer/pyspark
ERROR: invocation failed (exit code 1), logfile: 

________________________________________________________________________________________________ summary _________________________________________________________________________________________________
ERROR:   flake8: InvocationError: /media/psf/Home/dev/Functionalizer/pyspark/.tox/flake8/bin/pip install -e /media/psf/Home/dev/Functionalizer/pyspark (see /media/psf/Home/dev/Functionalizer/pyspark/.tox/flake8/log/flake8-2.log)
```

By commenting out `usedevelop = True` everything works as expected.

Can someone change the title to mention "use_develop" as well?

obestwalter:
[...]

PR #572, opened 2017-08-09
Fixing #571. skip_install should override usedevelop

### LINKED ISSUE #571: Let skip_install override usedevelop instead of the other way around
`skip_install` and `usedevelop` can be used together in the same environment but the behaviour is not documented. If both flags are `True` the package will be installed although `skip_install` is requested (see also discussion [here](https://github.com/tox-dev/tox/issues/270#issuecomment-320941011)).

The behaviour should be changed that `skip_install` overrides `usedevelop` as `usedevelop` is an option to choose the installation type if the package is installed and `skip_install` determines if it should be installed at all.

This is a somewhat breaking change of undocumented behaviour but I think this should be done to make the behaviour intuitively correct (but it should also be documented).

It should be also stated prominently in the changelog that this changed and how it can be fixed though. The documentation should be updated mention the behaviour when these two flags are used together.

ferdonline:
From the discussion in #270 one realises the need of supporting both skip_install and usedevelop together, with skip_install having priority (usedevelop being considered as an install option).
This PR implements such change.

For the tests to run also in py27 it was necessary to fix using FileNotFoundError
Also, in centos7 virtualenv alwayscopy is broken, so for the moment it detects the platform and skips the test. 
Fixes #571

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,6 +7,7 @@ Not released yet
 - #517: Forward NUMBER_OF_PROCESSORS by default on Windows to fix
         `multiprocessor.cpu_count()`.
 - #518: Forward `USERPROFILE` by default on Windows.
+- #571: skip_install should override usedevelop
 
 2.7.0
 -----
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -49,3 +49,4 @@ Nick Douma
 Cyril Roelandt
 Bartolome Sanchez Salado
 Laszlo Vasko
+Fernando L. Pereira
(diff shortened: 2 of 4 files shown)


obestwalter:
Thanks. Do we have a bug fo that CentOs thing already?

obestwalter:
ahh ... upstream bug - ok.

obestwalter:
@ferdonline I want to make a release from the existing PRs now. I won't merge this yet, as I am hoping for a bit more feedback about this and we have quite a lot of changes merged already for the next release. But I really don't want too much time pass between releases anymore, so this will be out soonish if there no objections to this change pop up.

ferdonline:
Alright! Thanks for keeping me up to date! 👍

Issue #507, opened 2017-04-20
Use the virtual environment to generate the sdist


dstufft:
Currently it appears that tox uses the system Python when creating the sdist (specifically, ``/usr/bin/python`` if that's where tox is installed). This means that if you're trying to use a setuptools feature that is newer than what your system provides in your ``setup.py``, then you're bound by what your current environment provides (which is often times fairly old). I think it would be a great idea if tox re-used one of the virtual environments it's creating to build the sdist (honestly it could be any of them) which would give more control over what version of setuptools is being used.

Additionally, I *think* this would also affect someone trying to use tox with a package whose ``setup.py`` only runs on Python 3.

For reference, here is the output I'm seeing that suggests this is what is happening:

```
+ tox -r -e docs -- --color=yes
GLOB sdist-make: /var/lib/jenkins/jobs/cryptography-pr-docs/workspace/cryptography/setup.py
ERROR: invocation failed (exit code 1), logfile: /var/lib/jenkins/jobs/cryptography-pr-docs/workspace/cryptography/.tox/log/tox-0.log
ERROR: actionid: tox
msg: packaging
cmdargs: ['/usr/bin/python', local('/var/lib/jenkins/jobs/cryptography-pr-docs/workspace/cryptography/setup.py'), 'sdist', '--formats=zip', '--dist-dir', local('/var/lib/jenkins/jobs/cryptography-pr-docs/workspace/cryptography/.tox/dist')]
env: None

Traceback (most recent call last):
  File "setup.py", line 332, in <module>
    **keywords_with_side_effects(sys.argv)
  File "/usr/lib/python2.7/distutils/core.py", line 111, in setup
    _setup_distribution = dist = klass(attrs)
  File "/usr/lib/python2.7/dist-packages/setuptools/dist.py", line 239, in __init__
    self.fetch_build_eggs(attrs.pop('setup_requires'))
  File "/usr/lib/python2.7/dist-packages/setuptools/dist.py", line 264, in fetch_build_eggs
    replace_conflicting=True
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 595, in resolve
    requirements = list(requirements)[::-1]  # set up the stack
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2605, in parse_requirements
    line, p, specs = scan_list(VERSION,LINE_END,line,p,(1,2),"version spec")
[...]

Issue #350, opened 2016-09-17
Support installing deps before setup.py sdist


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/350
- Originally reported by: [aragilar](http://bitbucket.org/aragilar)
- Originally created at: 2016-07-23T16:29:05.760

Some projects require dependencies to be installed before calling setup.py, it'd be nice to be able to do this (rather than trying to work around this by not building the sdist).

obestwalter:
Closing as this would be solved with #272.

Issue #232, opened 2016-09-17
dist creation commands


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/232
- Originally reported by: @RonnyPfannschmidt
- Originally created at: 2015-03-25T10:11:06.002

in order to support creation/installation of wheels as well as custom distribton generators it seems sensible to add a global option for the sdist creation command

it should default to `python setup.py sdist` but allow people to override
with something like `python setup.py sdist bdist_wheel`

in addition a global option is required to check if all requirements for the sdist creation are availiable

i propose to add the keys

`dist_command` for the command and `dist_requires` for the requirements needed to create the distribution (this would also allow to add requirements for recent setuptoolsand setup time dependencies)

whether the implementation creates a virtualenv to install those or tells the user to do so is up for implementation

in case of virtualenv creation we can also support specifying exact/desired tox versions there

gaborbernat:
This is now solved by having PEP-517/518 support. We offer sdist for now. Wheel support is encouraged to be added as a tox package plugin.

ionelmc:
@gaborbernat did anyone work on this wheel plugin? how come the wheel support was dropped (i saw some prs with it but the discussion is confusing)?

gaborbernat:
@ionelmc not as far as I'm aware of it. It wasn't dropped I just ended up conceding the benefit for projects I work on would be small. 😄 so it's just a matter of prioritization.

ionelmc:
@gaborbernat Ok so I've looked at the internals, looks like tox_package is the hook but it's very coarse - I'd only need to override the build_package function. I mean it's fine for now, i'll just copy-paste tox_package+get_package+acquire_package for now but maybe we want some more granular hooks at some point.

gaborbernat:
@ionelmc true that 👍 as a more proof safe implementation I would suggest you just monkeypatch the build_package? Feel free to add me as a reviewer once you have it done 👍

ionelmc:
@gaborbernat So I've looked a bit more now, so it looks it's really hard to also support wheels for pep517 projects (I'd need to copy-paste perform_isolated_build basically).

Also, isolated_build=true doesn't work without pyproject.toml - is that intentional?

gaborbernat:
> Also, isolated_build=true doesn't work without pyproject.toml - is that intentional?

PEP-517 requires PEP-518 which mandates ``pyproject.toml`` so yeah that's by PEP design.

[...]

PR #577, opened 2017-08-14
Groundwork for supporting PEP-518/`pyproject.toml`

alexander255:
This PR currently adds two new options:

  * `install_setuptools` (defaulting to `True`) to the per-testenv settings
  * `venvsdist` (defaulting to `False`) to the global `tox` settings

When `venvsdist` is `True` and a sdist is to be created then tox will provision a new virtual environment and run `./setup.py sdist` in there. All settings for that environment (except `changedir`, `commands` & `list_dependencies_command`) can be tweaked using the new `toxenv:sdist` configuration section.

(All new introduced names are ofc subject to change at your leisure.)

.

Here's a quick checklist of things that will be done once the current set of changes has been ACK'ed:

- [ ] if an enhancement PR please create docs and at best an example
- [ ] Add yourself to `CONTRIBUTORS`;
- [ ] make a descriptive Pull Request text (it will be used for changelog)

### DIFF
--- a/tox/config.py
+++ b/tox/config.py
@@ -595,9 +595,25 @@ def tox_addoption(parser):
     parser.add_testenv_attribute_obj(InstallcmdOption())
 
     parser.add_testenv_attribute(
+        name="install_setuptools", type="bool", default=True,
+        help="If true the ``setuptools`` and ``wheel`` packages will "
+             "automatically be installed into the virtual environment.")
+
+    def list_dependencies_command(testenv_config, value):
+        using_setuptools = testenv_config.install_setuptools
+
+        # Display setuptools and wheel when listing dependencies of an
+        # environment where they are not present by default
+        if value == ["pip", "freeze"] and not using_setuptools:
+            value.append("--all")
+
+        return value
+
+    parser.add_testenv_attribute(
         name="list_dependencies_command",
         type="argv",
         default="pip freeze",
+        postprocess=list_dependencies_command,
         help="list dependencies for a virtual environment")
 
     parser.add_testenv_attribute_obj(DepOption())
--- a/tox/config.py
+++ b/tox/config.py
@@ -673,6 +673,14 @@ class TestenvConfig:
         check later when the testenv is actually run and crash only then.
         """
 
+    def __str__(self):
+        return "<{0}: config={2!r}, envname={3!r}, factors={4!r} at 0x{1:x}>".format(
+            self.__class__.__name__, id(self),
+            self.config,
+            self.envname,
+            self.factors
+        )
+
     def get_envbindir(self):
         """ path to directory where scripts/binaries reside. """
         if sys.platform == "win32" and "jython" not in self.basepython and \
--- a/tox/config.py
+++ b/tox/config.py
@@ -856,6 +856,12 @@ class parseini:
             for env in _split_env(stated_envlist):
                 known_factors.update(env.split('-'))
 
+        # sdist creation environment
+        config.sdist_envconfig = self.make_envconfig(
+            "sdist", "toxenv:sdist", reader._subs, config, replace=True,
+            fallbacksections=[]
+        )
+
         # configure testenvs
         for name in all_envs:
             section = testenvprefix + name
@@ -867,6 +873,7 @@ class parseini:
                           config.envconfigs[name].usedevelop for name in config.envlist)
 
         config.skipsdist = reader.getbool("skipsdist", all_develop)
+        config.venvsdist = reader.getbool("venvsdist", False)
 
     def _list_section_factors(self, section):
         factors = set()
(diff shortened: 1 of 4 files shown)


gaborbernat:
@alexander255 it's hard to accept this without any unit tests to demonstrate the new features

alexander255:
I didn't ask you to, but I first want a confirmation that these changes will be accepted *in principle* (or comments on how to improve the current design), before I go through the trouble of writing tests & documentation.

tox/config.py
```diff
@@ -809,10 +840,11 @@ def _list_section_factors(self, section):
                 factors.update(*mapcat(_split_factor_expr, exprs))
         return factors
 
-    def make_envconfig(self, name, section, subs, config, replace=True):
+    def make_envconfig(self, name, section, subs, config, replace=True,
```
gaborbernat: note that the list here is globally change-able; you probably want to default to None, and if it's none then locally swap to "testenv"

tox/venv.py
```diff
@@ -178,7 +186,7 @@ def update(self, action):
                 self.envconfig.deps, v)
 
     def _getliveconfig(self):
-        python = self.envconfig.python_info.executable
+        python = str(self.envconfig.python_info.executable)
```
gaborbernat: care to detail why you need to wrap?

alexander255: Not sure why this was necessary, but I cannot reproduce the test failure that had been fixed during development of this comment. Thanks for pointing this out!

tox/config.py
alexander255: Fixed

obestwalter:
I have been thinkiing a lot recently about how we can make tox live long and prosper and IMHO extending the core to provide more hooks to make extensions like this possible is the way to go. That way the core won't get bloated and there is lots of roome to experiment and extend without having to wait for PRs merged. I will concentrate on making that possible and so I would not merge this PR but rather work towards that functionality like this can be implemented fully as a plugin.

gaborbernat:
@alexander255 I've recently got into the details of PEP-518 and PEP-517, so I'll partner on with this. Can you fixup the merge conflicts? And let's see how we can do this.

gaborbernat:
I've been pondering over this the last few weeks, and the more I think about this I think conceptually neither PEP-517 and PEP-518 are tox responsibility. 

They fall into the ``pip`` and ``setuptools``. ``tox`` is about creating the virtual environments and running the packages. How the package is built is handled by ``setuptools``, how the packages are installed is handled by ``pip``. So instead adding it to our project I think we should focus on adding these to those projects.

@obestwalter what do you think?

obestwalter:
Hi @gaborbernat, might be the case, but I would have to look at what this PR is actually trying to accomplish in detail. I would not merge this into core though even when it makes sense. I would rather work towards making it possible to have this as a plugin. core is already messy enough.

> How the package is built is handled by setuptools, how the packages are installed is handled by pip. So instead adding it to our project I think we should focus on adding these to those projects.

To that I agree - or to say it more generalized: How the package is built is handled by the build tool (e.g. setuptools) how the packages are installed is handled by the package manager (e.g. pip or conda). So instead adding it to our project I think we should focus on adding these to those projects.

aragilar:
The main problem I run into when setting up tox is dealing with non-setuptools build systems (primarily numpy.distutils, but widespread use of flit is also blocked by this). Pip (or conda) doesn't solve this either, as it's always the sdist component that is problematic (as sdist->wheel/install can be worked around by adding build deps to envs), not the install part. There's a bunch of related issues, which are solved by supporting PEP-518 (e.g. https://github.com/tox-dev/tox/issues/573, https://github.com/tox-dev/tox/issues/507, https://github.com/tox-dev/tox/issues/350, https://github.com/tox-dev/tox/issues/232). As long as there's no hooks or options to control the sdist process, tox blocks progress in making python packaging better. I'm happy to help with adding hooks and other improvements, but as an outsider I can't see any kind of plan to fix this, and the suggestion to implement this as a plugin seems to gloss over how much this is going to affect (e.g. what happens when use_develop specified)?

gaborbernat:
@aragilar you are correct; and thinking this through here's my current plan:
- PEP-518 is for build dependencies
- for build system choice the solution is PEP-517 (flit, numpy.distutils) etc. 

From tox side now we use ``python setup.py sdist``` to build and then install that with pip into the envs. Instead what we would need to do, first create a venv for building, do PEP-518 file support, install build dependencies; then use the API defined in PEP-517 to create the package, and then feed that to pip, as we do beforehand. 

A small side note, build requires is the build tool itself (e.g. flint, setuptools). The build systems may then install further dependencies to build, these are the packages of setup requires (e.g. setuptools_scm). I'm going to create a POC for these the next week inside tox.

gaborbernat:
For anyone wondering on the status of this, as a POC (either a playground to test out the rough edges of what this would mean for tox, or as a separate project) I've started doing tox3. It contains a lot of divergences from this (e.g. venv support - fallback to virtualenv, pyproject.toml configuration instead of tox.ini, PEP 517/518 support, asyncio driver) but will come back once I have something usable. WIP - https://github.com/gaborbernat/tox3.

gaborbernat:
Drawing from the experience implementing this in toxn (and with pip 10 coming out with PEP-518 support too), I'll add support for this in the next month via another PR. Given the merge conflicts I think it's easier to start clean. For people interested can follow via https://github.com/tox-dev/tox/issues/573.

PR #580, opened 2017-08-18
fix invalid pytest reference, and do not ignore Sphinx warnings anymore

gaborbernat:


doc/example/pytest.txt
```diff
@@ -2,10 +2,12 @@
 pytest and tox
 =================================
 
-It is easy to integrate `pytest`_ runs with tox.  If you encounter
+It is easy to integrate `PyTest`_ runs with tox.  If you encounter
```
obestwalter: pytest is really spelled "pytest" - all lower case - always.

doc/example/pytest.txt
```diff
@@ -2,10 +2,12 @@
...
 issues, please check if they are `listed as a known issue`_ and/or use
 the :doc:`support channels <../support>`.
 
+.. _`PyTest`: https://www.pytest.org
```
obestwalter: see above.

gaborbernat:
@obestwalter done.

obestwalter:
thanks :)

PR #582, opened 2017-08-18
use IOError and check error number (py2/py3 compat)

obestwalter:
FileNotFoundError was introduced in Python3, so this is not compatible with python2.

### DIFF
--- a/tox/session.py
+++ b/tox/session.py
@@ -450,7 +450,7 @@ class Session:
             try:
                 status = venv.update(action=action)
             except IOError as e:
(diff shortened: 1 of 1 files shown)


obestwalter:
I'd like to merge this and make it part of an rc2 soon. Anyone opposed?

nicoddemus:
Sounds good to me! 👍

gaborbernat:
agreed 👍

obestwalter:
thanks :)

PR #585, opened 2017-08-23
Document constraints.txt usage.

loechel:
I have documented the corresponding constrains.txt usage with tox, as it work similar to requirements.txt but was not documented.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@ Not released yet
 - #518: Forward `USERPROFILE` by default on Windows.
 - #515: Don't require environment variables in test environments
 	where they are not used.
+- : Add documentation for constrains.txt
 
 2.7.0
 -----
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,7 +9,7 @@ Not released yet
 - #518: Forward `USERPROFILE` by default on Windows.
 - #515: Don't require environment variables in test environments
 	where they are not used.
-- : Add documentation for constrains.txt
+- : Add documentation for constraints.txt
 
 2.7.0
 -----
--- a/doc/example/basic.txt
+++ b/doc/example/basic.txt
@@ -82,32 +82,38 @@ configuration::
 
 .. _multiindex:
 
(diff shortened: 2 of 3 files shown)


doc/example/basic.txt
```diff
@@ -97,6 +97,18 @@ All installation commands are executed using ``{toxinidir}``
 working directory.  Therefore, the underlying ``pip`` installation
 will assume ``requirements.txt`` to exist at ``{toxinidir}/requirements.txt``.
 
+defining constrains
+-----------------------------------------------
+
+.. versionadded:: 1.6.1
+
+(experimental) If you have a ``constrains.txt`` file you can add it to your ``deps``varaiable like this::
```
obestwalter: typo: varaiable

doc/example/basic.txt
```diff
@@ -97,6 +97,18 @@ All installation commands are executed using ``{toxinidir}``
...
+
+    deps = constrains.txt
```
obestwalter: This would have to be `deps = -cconstrains.txt`

obestwalter:
Hi @loechel thanks for the PR. I guess you mean this: https://pip.pypa.io/en/stable/user_guide/#constraints-files?

If yes: this is nothing that is implemented as such - it is basically a side effect of the way how deps is implemented. deps passes the `-<someflag>whatever.txt` through to pip, so you theoretically use any flag that would make sense to pip there.

Maybe it would be better explain that in above section documenting `-r` that mentions that it is also possible to pass e.g. the `-c` flag with a constraints file?

CHANGELOG
```diff
@@ -9,6 +9,7 @@ Not released yet
 - #518: Forward `USERPROFILE` by default on Windows.
 - #515: Don't require environment variables in test environments
 	where they are not used.
+- : Add documentation for constrains.txt
```
obestwalter: s/constrains/constraints/g

loechel:
@obestwalter thanks for the feedback, I have adjusted the documentation. I hope it is better now and ready to merge.

PR #588, opened 2017-08-26
use xfail_strict and adapt affected tests

### LINKED ISSUE #587: Execute tests with xfail_strict
We're using tests marked as xfail to reproduce known issues (e.g. upstream bugs), but at the moment we do not fail a test, when it passed unexpectedly. Adding `xfail_strict = True` to the pytest configuration makes sure we notice if a bug was "accidentally" fixed or fixed in upstream.

obestwalter:
fixes #587

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -2187,8 +2187,8 @@ class TestCmdInvocation:
         ])
 
     @pytest.mark.xfail(
-        "pypy not in sys.prefix",
-         reason='Upstream bug. See #203')
+        "'pypy' not in sys.prefix",
+        reason='Upstream bug. See #203')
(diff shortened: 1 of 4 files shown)


tests/test_config.py
```diff
@@ -1684,7 +1683,7 @@ def test_passing_empty_string(self, tmpdir, newconfig):
         args = ['--hashseed', '']
         self._check_testenv(newconfig, '', args=args)
 
-    @pytest.mark.xfail(sys.version_info >= (3, 2),
+    @pytest.mark.xfail("sys.version_info >= (3, 2) and sys.version < (3, 4)",
```
The-Compiler: `sys.version_info` - `sys.version` is a string

obestwalter: ups - thanks.

Issue #595, opened 2017-09-01
2.8.0: {[testenv]setenv} substitution broken 


yogi-sagar:
While installing tempest as part of devstack install, below commnd blows up with an error - 
`tox -r --notest -efull`

Error -
```
2017-09-01 18:22:20.661 | ++ lib/tempest:install_tempest:612          :   pushd /opt/stack/new/tempest
2017-09-01 18:22:20.661 | ~/tempest ~/devstack
2017-09-01 18:22:20.662 | ++ lib/tempest:install_tempest:613          :   tox -r --notest -efull
2017-09-01 18:22:21.720 | Traceback (most recent call last):
2017-09-01 18:22:21.720 |   File "/usr/local/bin/tox", line 11, in <module>
2017-09-01 18:22:21.720 |     sys.exit(cmdline())
2017-09-01 18:22:21.720 |   File "/usr/local/lib/python2.7/dist-packages/tox/session.py", line 38, in main
2017-09-01 18:22:21.720 |     config = prepare(args)
2017-09-01 18:22:21.720 |   File "/usr/local/lib/python2.7/dist-packages/tox/session.py", line 26, in prepare
2017-09-01 18:22:21.720 |     config = parseconfig(args)
2017-09-01 18:22:21.720 |   File "/usr/local/lib/python2.7/dist-packages/tox/config.py", line 242, in parseconfig
2017-09-01 18:22:21.720 |     parseini(config, inipath)
2017-09-01 18:22:21.720 |   File "/usr/local/lib/python2.7/dist-packages/tox/config.py", line 796, in __init__
2017-09-01 18:22:21.721 |     replace=name in config.envlist)
2017-09-01 18:22:21.721 |   File "/usr/local/lib/python2.7/dist-packages/tox/config.py", line 827, in make_envconfig
2017-09-01 18:22:21.721 |     res = meth(env_attr.name, env_attr.default, replace=replace)
2017-09-01 18:22:21.721 |   File "/usr/local/lib/python2.7/dist-packages/tox/config.py", line 964, in getdict_setenv
2017-09-01 18:22:21.721 |     definitions = self._getdict(value, default=default, sep=sep)
2017-09-01 18:22:21.721 |   File "/usr/local/lib/python2.7/dist-packages/tox/config.py", line 975, in _getdict
2017-09-01 18:22:21.721 |     name, rest = line.split('=', 1)
2017-09-01 18:22:21.721 | ValueError: need more than 1 value to unpack
[...]

PR #597, opened 2017-09-01
Don't attempt to parse dicts if not replacing

cboylan:
When reading in the config and processing dict content we were splitting
on = assuming we would always have key = value pairs. However in the
case where we don't want to replace because the testenv will not be used
this can result in ValueErrors as substitution variables are not string
key = value pairs.

Avoid all this by only attempting to process dict key = values if we are
replacing in the first place. Otherwise the content isn't needed and we
just return the default value or empty dict.

This should fix https://github.com/tox-dev/tox/issues/595

Thanks for contributing a PR!

- [X] Make sure to include one or more tests for your change;
- [NA] if an enhancement PR please create docs and at best an example
- [X] Add yourself to `CONTRIBUTORS`;
- [X] make a descriptive Pull Request text (it will be used for changelog)

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -345,15 +345,15 @@ class TestIniParserAgainstCommandsKey:
         assert envconfig.commands == [["echo", "bar"]]
 
     def test_reproduce_issue595(self, newconfig):
-        with pytest.raises(ValueError):
-            newconfig("""
-                [tox]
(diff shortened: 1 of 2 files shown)


obestwalter:
Great, thank you! I will fix up whatever is missing and push a bug fix release

PR #608, opened 2017-09-06
fix --showconfig setenv (show actual data)

obestwalter:
Tiny fix to actually see the setenv data when calling --showconfig


obestwalter:
I did not want to merge that myself actually, git made me do it :(

PR #611, opened 2017-09-07
revert changelog format

obestwalter:
I think we should not change the name of CHANGELOG and also not the format for now. The source is much easier on the eye and it gets automatically transformed to rst on release.

I suggest we discuss how to deal with that in general in the future by e.g. introducing towncrier as the manual updating of the log is quite tedious even for a moderately active project like tox.

I also think that the name shold remain CHANGELOG the way it is as long as we have no strong reason to change it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,70 +7,68 @@ Not released yet
 2.8.1 (2017-09-04)
 ------------------
 
-- :pull:`599`: fix problems with implementation of :issue:`515`.
-  Substitutions from other sections were not made anymore if they were not in `envlist`.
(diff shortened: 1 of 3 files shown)


nicoddemus:
FWIW I'm very happy with our changing over to using `towncrier`, the amount of annoying conflicts have dropped significantly (almost none).

obestwalter:
Sorry, PR from wrong state, need to learn to juggle multiple PRs better ... will open another one.

obestwalter:
there: https://github.com/tox-dev/tox/pull/612

PR #612, opened 2017-09-07
Keep existing changelog format

obestwalter:
I think we should keep the old CHANGELOG format for now. The source is much easier on the eye and it gets automatically transformed to rst on release.

I suggest we discuss how to deal with that in general in the future by e.g. introducing towncrier as the manual updating of the log is quite tedious even for a moderately active project like tox.

I also think that the name shold remain CHANGELOG the way it is as long as we have no strong reason to change it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,70 +7,68 @@ Not released yet
 2.8.1 (2017-09-04)
 ------------------
 
-- :pull:`599`: fix problems with implementation of :issue:`515`.
-  Substitutions from other sections were not made anymore if they were not in `envlist`.
-  Thanks to Clark Boylan (:user:`cboylan`) for helping to get this fixed (:pull:`597`).
+- #p599: fix problems with implementation of #515.
+         Substitutions from other sections were not made anymore if they were not in `envlist`.
+         Thanks to Clark Boylan (@cboylan) for helping to get this fixed (#p597).
 
 2.8.0 (2017-09-01)
 -------------------
 
-- :issue:`276`: Remove easy_install from docs (TL;DR: use pip). Thanks Martin Andrysík (:user:`sifuraz`).
-
-- :issue:`301`: Expand nested substitutions in ``tox.ini``. Thanks :user:`vlaci`. Thanks to Eli Collins
-  (:user:`eli-collins`) for creating a reproducer.
-
-- :issue:`315`: add ``--help`` and ``--version`` to helptox-quickstart. Thanks :user:`vlaci`.
-
-- :issue:`326`: Fix ``OSError`` 'Not a directory' when creating env on Jython 2.7.0. Thanks Nick Douma (:user:`LordGaav`).
-
-- :issue:`429`: Forward ``MSYSTEM`` by default on Windows. Thanks Marius Gedminas (:user:`mgedmin`) for reporting this.
-
-- :issue:`449`: add multi platform example to the docs. Thanks Aleks Bunin (:user:`sashkab`) and :user:`rndr`.
-
-- :issue:`474`: Start using setuptools_scm for tag based versioning.
-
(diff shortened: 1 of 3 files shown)


gaborbernat:
I don't agree; I find the changelog easier to edit as an rst file; I would prefer to rename it during packaging to just CHANGELOG. I also don't see how it's easier on the eye as it is.

obestwalter:
I think we should not waste too much energy on this, when we all agree anyway that we want to automate that. I am definetely not keen on wasting more time maintaining the CHANGELOG or even discussing about what format it has. I opened that PR with the thought in mind that I am pretty much the only one maintaining that file at the moment and I wanted to keep it easy for myself.

I also think that the CHANGELOG in the root should be called just that there is no need to add an exetension and change the format. It also should be as markdown light as possible, which it was before and now it is cluttered with rst links, that are generated for the end result. 

I am all for moving the project forward, but we don't have to change everything around if it does not make things simpler. Renaming the file and turning it into rst (which it simply wasn't before) is not making things simpler in my eyes.

But maybe it is more constructive if instead we move to towncrier right away and learn how to use it properly? Would you like to look into that and open a PR to that effect instead?

gaborbernat:
I will.

gaborbernat:
I investigated but found ``towncrier`` more a nuisance that helping. How it works that you give up on writing changelog, instead you write news fragments in specific files (newsfragment/571.feature, 500.doc, 600.bugfix).  And then run a command to merge the content of all those. The only thing we achieve is merge conflicts. Users still need to manually write changelog manually and because the file extensions aren't ``rst`` no syntax highlight will be provided. 

Also I prefer to have links in our official documentation (to users, pull requests, issues). If these obstruct for the Pypi let's generate the Pypi changelog on build time and strip out these links.

obestwalter:
@nicoddemus seems quite happy with towncrier and actually solves the merge conflicts, because as I understand it the changelog is generated from those newsfragments (correct me if I am wrong).

gaborbernat:
yes; it's generated at the end of day (I suppose make sense to do so before every release); and the only benefit is merge conflicts, so true for that works perfectly

nicoddemus:
> yes; it's generated at the end of day (I suppose make sense to do so before every release); and the only benefit is merge conflicts, so true for that works perfectly

Indeed, the absolute main benefit is avoiding merge conflicts, which in pytest it was a pain. Another reason is that you can get away with using link to the issues, towncrier can manage that for you (feel free to copy pytest's template).

> I also think that the CHANGELOG in the root should be called just that. There is no need to add an extension and change the format. It also should be as markdown light as possible, which it was before and now it is cluttered with rst links, that are normally just generated for the end result (on pypi.org).

IMHO I prefer the changelog with the `.rst` extension; GitHub can then render it correctly. I don't see any good reason to name it `CHANGELOG` without extension, but that's my preference of course.

gaborbernat:
@nicoddemus yes, but works only for issues, not for users or PRs (although issues do forward to PR links, but do so permanently)

nicoddemus:
> yes, but works only for issues, not for users or PRs (although issues do forward to PR links, but do so permanently)

Indeed. Btw, we gave up on thanking people on the CHANGELOG, mostly to make it easier to maintain it. We already mention people that helped the release on the release notes, we have thought this to be enough credit vs. maintaining the changelog.

obestwalter:
And they can be attached to the PR either by the contributor or the maintainer which leads to merging a PR automatically adding the necessary news item. This would be a great relief for me. I spend a lot of time getting the CHANGELOG right before a release. I can very well life without that and spent that time hacking.

@nicoddemus and @gaborbernat ... oh well, then let's leave the CHANGELOG as rst ☮️

gaborbernat:
@obestwalter I'll add the PR for towncrie tomorrow 👍

PR #622, opened 2017-09-10
Remove usage of py.process

asottile:
Refs #610

### DIFF
--- a/tox/session.py
+++ b/tox/session.py
@@ -12,7 +12,6 @@ import shutil
 import subprocess
 import sys
 import time
-from subprocess import STDOUT
 
 import py
 
@@ -138,7 +137,7 @@ class Action(object):
             cwd = py.path.local()
         try:
             popen = self._popen(args, cwd, env=env,
-                                stdout=stdout, stderr=STDOUT)
+                                stdout=stdout, stderr=subprocess.STDOUT)
         except OSError as e:
             self.report.error("invocation failed (errno %d), args: %s, cwd: %s" %
                               (e.errno, args, cwd))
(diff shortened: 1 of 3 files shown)


asottile:
ah, need to update the mock too!

asottile:
This missing coverage in the diff here seems to be a case I don't think is possible any more -- or at least I don't see any evidence of virtualenv being vendored into tox -- should I just remove that branch?

obestwalter:
Good catch. That changeset is from 2013 and had to do with py2.5 support, so this can definitely go. This is also only used in --showconfig, and where is that vendored code supposed to be? It's not in the project, so this must have gone long ago and that check was just overlooked.

tox/session.py
```diff
@@ -673,12 +672,12 @@ def report_env(e):
...
-        except py.process.cmdexec.Error:
-            versions.append("virtualenv-1.9.1 (vendored)")
-        else:
-            versions.append("virtualenv-%s" % version.strip())
+        proc = subprocess.Popen(
+            (sys.executable, '-m', 'virtualenv', '--version'),
+            stdout=subprocess.PIPE,
+        )
+        out, _ = proc.communicate()
+        versions.append('virtualenv-{0}'.format(out.decode('UTF-8').strip()))
```
obestwalter: I personally don't like the .format way of string formatting, but as long as no one is forcing me to use it myself I can live with it :)

asottile: interesting! I prefer it over `%` formatting 😆

The-Compiler: @obestwalter Maybe https://pyformat.info/ will convince you otherwise? :wink:

obestwalter: @The-Compiler I am afraid not. I don't see any value in using .format for **simple** string formatting. So for now I will still choose

    'some text %s bla bla' % somevar

over

    'some text {0} bla bla'.format(somevar)

...  until I know better.

I also wish that Guido had engaged his time machine and introduced f-strings right away rather than in 3.6 ...

asottile: with ~clever~ evil hacks, I could probably get this to work in python2.7

```
# -*- encoding: future_fstrings -*-
thing = 'world'
print(f'hello {thing}')
```

obestwalter: Can't wait to drop 2.6 then :D

asottile: It is done! https://github.com/asottile/future-fstrings

nicoddemus: Hah! Oh dear! 😆

obestwalter: O.M.G.

gaborbernat: @asottile any performance hit compared to format?

asottile: They'll be slower that native f string support but this hack will have identical performance to .format (it translates the f string into format calls in fact). Native f strings are only faster because they are implemented directly in byte code (which a backport can't quite do)

PR #620, opened 2017-09-10
Remove py25-specific code

asottile:

PR #623, opened 2017-09-10
Add a pre-commit configuration

asottile:
This adds a [pre-commit](http://pre-commit.com) configuration with a minimal set of hooks.  You can optionally install it while developing and it will enforce its checks on commit (so you don't have to wait for a round trip of CI).

For now I've only added a small set of hooks (I've split them into separate commits so you can compare the changes that each hook performs).

Perhaps the most _controversial_ of these hooks is an import sorter, but this will help reduce merge conflicts and enforce a uniform style.  There's some prior discussion of this in particular here: #620.  The particular style that the tool chooses is outlined [here](https://github.com/asottile/reorder_python_imports#what-does-it-do).  The specific style aims to avoid merge conflicts as much as possible.


obestwalter:
I think it is a good time to get the whole hook system up to speed and provide everything to make tox more plugin driven. I haven't had time to look at the details yet, but the direction is definetly the right one. Maybe we should think about a tox 3.0 and what should be part of it. Promoting hooks away from their experimental status and provide proper help and documentation for plugin writers would be my favorite thing to do. Then we can also finally add conda support.

asottile:
Ah this doesn't actually involve plugins at all -- this just adds a devtool that makes _developing tox_ easier.  Think of it like a package manager for linters / fixers.  It just so happens to be convenient to run the tool as part of a `tox` environment.

obestwalter:
ah, sorry, was only skimming to look at it later, thought that you added some hooks :)

asottile:
heh, _pre-commit_ hooks, not _tox_ hooks -- I can see where the confusion would be! 😆

obestwalter:
Just learned that what I did here was SR;MP

tl-dr-and-friends

gaborbernat:
@asottile I'm not sure why make this part of a pre-hook and not a tox env? I don't consider style to be more or less important than tests running.

asottile:
It's actually intended for *both*.  Running just in CI is quite annoying as you have to wait for the long turnaround period.  It runs as part of CI because using / installing git hooks is optional and can be bypasssed.  Running as part of commit is purely convenenient -- I no longer need to wait for the (slow) `style` environment to be built and run and I don't have to think about remembering to do it each time -- I also no longer need to think about the reformats it does!

gaborbernat:
@asottile but you sort of assume that the libraries doing those checks are installed with the correct version already exists; which is kinda dangerous, as they may be either out of date or ahead. Also you don't need to remember, you should be running all environments before you push anything for a PR.

asottile:
pre-commit makes no assumptions about previously installed things and manages all tools in isolated virtualenvs.  The configuration file aims to produce repeatable builds by locking the dependent tools to a specific version.  Unlike competing tools, it is designed to use **no** system tools (though you can shimmy them in if you must).

Running all environments is tedious and slow, I'd rather run the specific tests I'm targetting and then throw it against CI -- especially on a slow testsuite like tox.

pre-commit also targets *just the files you're staging* and *just the contents you're staging* so it's often much speedier than relinting the entire codebase each time.

gaborbernat:
fair enough, but I think the actual solution would be to speed up the test suite; I think the current speed is unacceptable; see https://github.com/tox-dev/tox/issues/613, although that is much harder to achieve that commit hooks 👍

Issue #466, opened 2017-03-01
Whitelist environment variables to be shown on error


The-Compiler:
When installing dependencies failed, tox shows the whole environment:

```
ERROR: invocation failed (exit code 1), logfile: /home/florian/proj/qutebrowser/git/.tox/fail/log/fail-1.log
ERROR: actionid: fail
msg: getenv
cmdargs: ['/home/florian/proj/qutebrowser/git/.tox/fail/bin/pip', 'install', 'afiojafsioajfaisjf']
[...]

PR #625, opened 2017-09-10
#466 stop env var leakage if popen failed with resultjson or redirect

obestwalter:
hotfix vor 2.8.x will integrate that into master as well. Master had too many changes to just put them out quickly, but I think this should be fixed now.


obestwalter:
Note: this is intentionally merged into a branch. The changes can be cherry picked into current master then later on, rather than merging.

PR #627, opened 2017-09-10
Use string types check

obestwalter:
as suggested here: https://github.com/tox-dev/tox/pull/617#pullrequestreview-61678497


obestwalter:
I have to up my git foo seriously, when juggling different changeset ... sigh

obestwalter:
that's better :)

asottile:
This has different semantics.  In python2 `string_types` is `(unicode, bytes`) in python3 it is just `(str,)` -- not sure if that matters in this case but it is the reason that I left it as two separate conditionals.

obestwalter:
The check only looks if it is some kind of string or a an iterable containing name and version, so that simple test is sufficient.

PR #633, opened 2017-09-14
Add a --no-skip-missing-interpreters option

### LINKED ISSUE #631: Need option for not skipping missing interpreters
I really like to have `skip_missing_interpreters = True` in my `tox.ini`, because I'm often working on machines that don't have every Python interpreter installed.

But it's a really bad idea in a CI scenario, because environments just pass without doing anything.

I'd love to see an option to set that value back to `False`, but there's only a CLI option for setting it to `True`… every flag really should have an opposite.

asottile:
Resolves #631

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -45,6 +45,12 @@ and will first lookup global tox settings in this section::
     always overrides this setting if passed on the invokation.
     **Default:** ``False``
 
+    .. versionadded:: (unreleased)
+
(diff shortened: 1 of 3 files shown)


asottile:
I wasn't sure whether to use `versionadded` or `versionchanged`, shrugs!

obestwalter:
I dunno if this is really necessary.

obestwalter:
see #634 for dicussion.

asottile:
Thought you closed this, I'll do it!

PR #638, opened 2017-09-15
Invalid escape sequence fixes

scop:


The-Compiler:
I'm wondering why we didn't notice earlier? Maybe it'd be a good idea to turn on `-Werror` for the testsuite?

obestwalter:
woah - if I run pytest -Werror basically everything fails for using old pytest constructs ...

PR #636, opened 2017-09-15
Add/update release and contributor howtos and adapt to how pytest does stuff

obestwalter:
closes #614

These are my first efforts of getting to an easy release process that still does all the fancy stuff needed.

I extened towncrier to also link to prs and added a contrib folder with little helpers to automate releases. 

towncrier pr is here: https://github.com/hawkowl/towncrier/pull/92

PR #640, opened 2017-09-20
[WIP] docs cleanup

obestwalter:
This one saddles on top of #636 and aims to make the docs more digestible using the standard readthedocs scheme and bringing everything up to speed slowly.

A current build of the docs is here: https://devpi.net/obestwalter/dev/tox/2.9.0rc2.dev64+g0fcd5ce/+d/index.html

Note to self: squash commits


obestwalter:
The advantages I see in this particular theme:

* It is the standard theme, so motst likely to be supported well and light on bugs
* It has far superior navigation than the old theme
* It looks familiar if you are living in the Python ecosystem, as many packages use the default theme.
* I did not have to search around long for another theme :)

The width is nothing that bothers me too much, the above points are far more important for me. The main work is going on under the hood regarding the content though, anyway. If we don't go for the default theme, I would try the theme pytest is using, to be visually similar to our "buddy project" - but default is good enough for me atm. The old tox theme was the default back then I think, so all I did, was staying with the defaultm but the 2017 default instead of the 2012 one.

obestwalter:
oh and another point I forgot - not atschnical one, but near to my heart none the less: if we use the default theme, we support readthedocs, by displaying their ads, which I think is good, because they have big trouble in getting sustainable funding, so this is a little thing we could do to help: http://blog.readthedocs.com/ads-on-read-the-docs/

gaborbernat:
"many" projects do use a different style, see requests or pytest; but I understand your point :+1: I just don't like how it feels on a reasonably sized monitor; but this is probably a "bug"  for readthedocs site than our project

obestwalter:
"many" is obviousy a very fuzzy metric :)

obestwalter:
Closing this - I realized that the scope of this PR is far to wide.

gaborbernat:
@obestwalter so are you breaking up this in multiple sub pull requests?

obestwalter:
Yes, but this will have to wait a bit. I am in London next week, giving a pytest/tox training :) - so maybe we can meet and discuss a few things IRL?

gaborbernat:
@obestwalter yes, let's do that.

PR #643, opened 2017-09-29
update changelog for 2.9 release

obestwalter:
Add changes since rc and explain that towncrier generated text is not sacrosanct. The human that does the release still can adjust it.


obestwalter:
I will merge this and make the release today after that I would also like to merge the towncrier fixes and finalize the updated release process.

Issue #331, opened 2016-09-17
Output with missing setup.py could be nicer


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/331
- Originally reported by: @The-Compiler
- Originally created at: 2016-05-13T11:16:35.722

When there's no `setup.py`, we get:

```
$ tox 
Traceback (most recent call last):
  File "./.venv/bin/tox", line 11, in <module>
    sys.exit(cmdline())
  File "/home/florian/tmp/.venv/lib/python3.4/site-packages/tox/session.py", line 39, in main
    retcode = Session(config).runcommand()
  File "/home/florian/tmp/.venv/lib/python3.4/site-packages/tox/session.py", line 375, in runcommand
    return self.subcommand_test()
  File "/home/florian/tmp/.venv/lib/python3.4/site-packages/tox/session.py", line 520, in subcommand_test
    path = self.get_installpkg_path()
  File "/home/florian/tmp/.venv/lib/python3.4/site-packages/tox/session.py", line 496, in get_installpkg_path
    path = self._makesdist()
  File "/home/florian/tmp/.venv/lib/python3.4/site-packages/tox/session.py", line 390, in _makesdist
    raise tox.exception.MissingFile(setup)
tox.MissingFile: MissingFile: /home/florian/tmp/setup.py
```

instead the output could be nicer, like if the `setup.py` is broken:

```
$ tox                               
GLOB sdist-make: /home/florian/tmp/setup.py
ERROR: No dist directory found. Please check setup.py, e.g with:
     python setup.py sdist
```

PR #650, opened 2017-10-15
Provide friendlier error for missing setup.py.

Volcyy:
As suggested in #331, this PR makes `tox` output a friendlier exception when `setup.py` could not be found. When this is the case, the following is outputted instead of throwing a `MissingFile` exception:

tox output missing setup.py

**Checklist**:
- [X] added/updated test(s)
Should this be documented and if yes, where?
- [X] added relevant [issue keyword](https://help.github.com/articles/closing-issues-using-keywords/)
      in message body
- [X] added news fragment in [changelog folder](https://github.com/tox-dev/tox/tree/master/changelog), see [this](https://github.com/Volcyy/tox/blob/nicer-missing-setup-output/changelog/331.misc.rst)
- [X] added yourself to `CONTRIBUTORS` (preserving alphabetical order)

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -23,6 +23,7 @@ Ionel Maries Cristian
 Ionel Maries Cristian
 Itxaka Serrano
 Jannis Leidel
+Johannes Christ
 Josh Smeaton
 Julian Krause
 Krisztian Fekete
--- a/changelog/331.misc.rst
+++ b/changelog/331.misc.rst
@@ -1,0 +1,1 @@
+Running ``tox`` without a ``setup.py`` now has a more friendly error message and gives troubleshooting suggestions - by @Volcyy.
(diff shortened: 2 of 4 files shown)


asottile:
Here's the answer to "why does it exit 0": https://github.com/Volcyy/tox/blob/d5a68ad9b6313d9505f9ecfdcd32556e040d4e20/tox/_pytestplugin.py#L309

`initproj` sets up a `setup.py` by default, you'll probably need to explicitly delete it in your test to demonstrate the behaviour

Volcyy:
Thanks for telling me that, I added the test case, it removes the file now. There seems to be some issue with [one of the Travis builds](https://travis-ci.org/tox-dev/tox/jobs/299622108) though, and I'm not sure how to fix that.

asottile:
I nudged travis-ci for ya :)

This looks good to me, curious to hear one of the more seasoned maintainers chime in on whether we're ok with switching the exception here.

PR #657, opened 2017-10-26
Exclude setuptools 34 & 35 from requirements

### LINKED ISSUE #656: Installation failures with setuptools 34 & 35
This is a continuation of tox-dev/tox-travis#76. 

Travis's python 3.4 environment ships a broken version of setuptools (more details in the tox-travis issue). When upgrading dependencies that setuptools itself depends on (such as six in this case) a race condition may occur where the package temporarily does not exist, causing the installation to fail.

rpkilby:
Travis's python 3.4 environment ships a broken version of setuptools (more details in the tox-travis issue). When upgrading dependencies that setuptools itself depends on (such as six in this case) a race condition may occur where the package temporarily does not exist, causing the installation to fail. 

By installing setuptools first, which re-vendors its dependencies, we can prevent the race condition. 

Fixes #656

## Contribution checklist:

- [x] ~~added/updated test(s)~~
- [x] ~~updated/extended the documentation~~
- [x] ~~added relevant [issue keyword](https://help.github.com/articles/closing-issues-using-keywords/) in message body~~
- [x] added news fragment in [changelog folder](https://github.com/tox-dev/tox/tree/master/changelog)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ def main():
     version = sys.version_info[:2]
     virtualenv_open = ['virtualenv>=1.11.2']
     virtualenv_capped = ['virtualenv>=1.11.2,<14']
-    install_requires = ['py>=1.4.17', 'pluggy>=0.3.0,<1.0', 'six']
+    install_requires = ['setuptools!=34,!=35', 'py>=1.4.17', 'pluggy>=0.3.0,<1.0', 'six']
     extras_require = {}
     if has_environment_marker_support():
         extras_require[':python_version=="2.6"'] = ['argparse']
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -45,6 +45,7 @@ Paweł Adamczak
 Philip Thiem
 Ronald Evers
 Ronny Pfannschmidt
+Ryan P Kilby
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -50,7 +50,7 @@ def main():
     version = sys.version_info[:2]
     virtualenv_open = ['virtualenv>=1.11.2']
     virtualenv_capped = ['virtualenv>=1.11.2,<14']
-    install_requires = ['py>=1.4.17', 'pluggy>=0.3.0,<1.0', 'six']
+    install_requires = ['setuptools!=34,!=35', 'py>=1.4.17', 'pluggy>=0.3.0,<1.0', 'six']
```
asottile: This will add a dependency `tox->setuptools`, there isn't really a good way to indicate a "conflicts" in setup.py

tox doesn't really have a runtime need for setuptools so I don't think this is appropriate for `setup_requires`

rpkilby: I agree, this was simply one option presented. 

Another option would be to have a note in the docs, and to fix the issue in tox's own CI. Instead of `pip install -U --pre tox`, do `pip install -U --pre setuptools tox`.

PR #660, opened 2017-10-26
cmd simplify, use pytest stderr/stdout management, and use regex instead of fnmatch

gaborbernat:
this is part of trying to speed up the unit tests with three changes that make the test less tox specific (use what the builtin stdlib/pytest offers/uses) and remove some indirection:

- the Cmd class is needlessly a class - it has just an init and one call - make it instead just a simple fixture (which returns a callable that can  be invoked with the command)
- remove calling subprocess  from Cmd, call instead directly to remove a sub-process indirection 
-  instead of hacking the stdin/stdout away during hacking, use the pytest provided capfd
- the fnmatch is kinda hard to read, and allows non precise test outcome specification - migrate to regex instead

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
     language: generic
 
 before_install: if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then source .travis-osx; fi
-install: pip install six && pip install --pre -U tox
+install: pip install --pre -U tox --ignore-installed six
 script: tox -e py
 after_success:
   - tox -e coverage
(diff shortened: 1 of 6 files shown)


gaborbernat:
this is part of https://github.com/tox-dev/tox/issues/613, initially I wanted to have a single PR, but now I've decided to put it through though chunks for easier reviewing purpose

obestwalter:
Hi @gaborbernat - looks great and even if it is doesn't get much faster, it's a lot tidier :)

gaborbernat:
@obestwalter will have a look to improve the coverage 👍 and once that done, we can merge it in

Issue #651, opened 2017-10-20
tox test is failed due to CHANGELOG file


stanislavlevin:
`TOXENV=py27 /usr/bin/python setup.py test`  is failed.

```
running test
running egg_info
writing requirements to tox.egg-info/requires.txt
writing tox.egg-info/PKG-INFO
writing top-level names to tox.egg-info/top_level.txt
writing dependency_links to tox.egg-info/dependency_links.txt
writing entry points to tox.egg-info/entry_points.txt
reading manifest file 'tox.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'CHANGELOG'
writing manifest file 'tox.egg-info/SOURCES.txt'
running build_ext
using tox.ini: /usr/src/RPM/BUILD/python-module-tox-2.9.1/tox.ini
using tox-2.9.1 from /usr/src/RPM/BUILD/python-module-tox-2.9.1/tox/__init__.pyc
GLOB sdist-make: /usr/src/RPM/BUILD/python-module-tox-2.9.1/setup.py
  /usr/src/RPM/BUILD/python-module-tox-2.9.1$ /usr/bin/python /usr/src/RPM/BUILD/python-module-tox-2.9.1/setup.py sdist --formats=zip --dist-dir /usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox/dist >/usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox/log/tox-0.log
py create: /usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox/py
  /usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox$ /usr/bin/python -m virtualenv --python /usr/bin/python py >/usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox/py/log/py-0.log
py installdeps: pytest >= 3.0.0, pytest-cov, pytest-timeout, pytest-xdist
  /usr/src/RPM/BUILD/python-module-tox-2.9.1$ /usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox/py/bin/pip install pytest >= 3.0.0 pytest-cov pytest-timeout pytest-xdist >/usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox/py/log/py-1.log
py inst: /usr/src/RPM/BUILD/python-module-tox-2.9.1/.tox/dist/tox-2.9.1.zip
[...]

PR #662, opened 2017-10-26
Upgrade pre-commit hooks

asottile:
This should remove the symlink inside pre-commit-hooks.

piggybacking on https://github.com/pre-commit/pre-commit-hooks/commit/c5b7c35d81174fa8e69c50f6561fa9537f1334e5

Resolves #661

PR #664, opened 2017-10-26
tox no longer uses `python setup.py test`

asottile:
Refs #651


asottile:
tests will fail until https://github.com/tox-dev/tox/pull/662 is merged unfortunately :S

gaborbernat:
@asottile I suppose if you rebase this now on top of master the CI should pass

PR #665, opened 2017-10-26
Remove unneeded reading from absolute paths

asottile:


gaborbernat:
@asottile I suppose if you rebase this now on top of master the CI should pass

Issue #149, opened 2016-09-17
virtualenv is not recreated when deps change if '-r' is used


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/149
- Originally reported by: @gavrie
- Originally created at: 2014-01-28T15:36:07.365

In my tox.ini, I have the following deps:

```
deps = -rpackaging/requirements.txt
```

Normally, when deps change, the virtualenv is recreated by tox.
However, when the contents of requirements.txt change, tox does not notice this and the environment is reused.

A workaround would be to list the deps directly in tox.ini, but that would require duplicating the contents of the existing requirements.txt that is used by other tools.

dimaqq:
Any chance of porting `tox-battery` here?

RonnyPfannschmidt:
@dimaqq it already is on github - @signalpillar is the maintainer/owner

obestwalter:
Are there opinions about how we should deal with this?

I would tend to close this and recommend using [tox-battery](https://github.com/signalpillar/tox-battery) for now.

The-Compiler:
I'll have to disagree - this is an issue I hear from various people, and it's quite unexpected. I view it as "tox does caching without invalidating the cache correctly" essentially. The core does caching, so I think it really should do this correctly.

obestwalter:
o.k. - I stand corrected then :)

hpk42:
we once had better support for requirements parsing but we had to revert it ... i admit i never extensively used -r requyirements.txt myself and the docs still mark that feature as experimental http://tox.readthedocs.io/en/latest/example/basic.html?highlight=requirements#depending-on-requirements-txt -- someone needs to step up to go for it more properly ... meanwhile recommending to manually use "tox -r" for recreating an env is a kludge to get around tox's ignorance.

merwok:
FWIW, in my latest project I avoid the issue by using pip-tools.  I never recreate virtualenvs, but dependencies (in one or more requirements files, e.g. run.txt and test.txt) are always synchronized, so I get fast failures before commit and CI if I forget to update requirements.

obestwalter:
@merwok that sounds interesting, but I have no idea what you are talking about - could you link to some documentation/blog/howto/whatever with more details about this approach? Thanks!

asottile:
[...]

PR #668, opened 2017-11-02
recreate venv when -rrequire.txt changes

kapilt:
This seems to be a long standing fundamental issue with tox, thats been long sidelined by calling -r in deps spec experimental, but given its common usage,  one worth correcting given its a common pitfall, and reported numerous times per discussion and refs in #149. i took a look at the suggestions in there of various other tools, but they all had various pitfalls, and also took a look at what some other real world projects are doing with tox (like botocore which does a separate installer script in commands to avoid this pitfall) before deciding the simplest thing was just fixing tox, via checksum on requirements files.

### DIFF
--- a/tox/config.py
+++ b/tox/config.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import argparse
 import itertools
+import hashlib
 import os
 import random
 import re
@@ -909,6 +910,18 @@ class DepConfig:
     def __init__(self, name, indexserver=None):
         self.name = name
         self.indexserver = indexserver
+
+    @property
+    def digest(self):
+        fname = None
+        if self.name.startswith('-r'):
+            fname = self.name[2:]
+        else:
+            fname = self.name
+        path = py.path.local(fname)
+        if not path.check(file=1):
+            return hashlib.md5(path).hexdigest()
+        return path.computehash()
 
     def __str__(self):
         if self.indexserver:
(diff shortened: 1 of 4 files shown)


asottile:
You may be interested in [tox-pip-extensions](https://github.com/tox-dev/tox-pip-extensions) which attempts to solve this problem more generically (the venv-update extension diffs requirements and gets to a final state by installing/uninstalling)

obestwalter:
Hi @kapilt and @asottile if you both agree that this is a problem worth solving (I definitely also agree :)) and you both already have approaches it would be great if you could put your heads together and tackle that together. This really is something that comes up an awful lot and confuses users who assume that this works properly. I collected all tickets that have to do with that general problem here: https://github.com/tox-dev/tox/projects/9

tox/config.py
```diff
@@ -910,6 +910,16 @@ def __init__(self, name, indexserver=None):
         self.name = name
         self.indexserver = indexserver
 
+    @property
+    def digest(self):
+        fname = str(self.name)
+        if fname.startswith('-r'):
+            fname = fname[2:]
```
merwok: There may be optional whitespace: `-r requirements.txt`

cryvate: Actually, there can be at the moment, but it won't work as expected: it would search for requirements.txt (note space). See PR #669.

You should check for --requirement as well, in which case you will have to strip an =.

A requirements file itself can depend on a requirements file (e.g. dev requirements on install requirements). I guess that is not being checked here?

gaborbernat:
In lack of activity I'll close this for now, we can reopen it when we come up wit ha common solution.

PR #670, opened 2017-11-08
Skip whitespace in envstr.

andreif:
This change would allow the following version matrix:

```ini
[tox]
envlist = py24-django{ 12, 13                                       },
          py25-django{ 12, 13, 14                                   },
          py26-django{     13, 14, 15, 16                           },
          py27-django{         14, 15, 16, 17, 18, 19, 110, 111     },
          py32-django{             15, 16, 17, 18                   },
          py33-django{             15, 16, 17, 18                   },
          py34-django{                     17, 18, 19, 110, 111, 20 },
          py35-django{                         18, 19, 110, 111, 20 },
          py36-django{                                      111, 20 }
```

In the meanwhile, one can install `tox` enabling such config by
```
pip install git+https://github.com/andreif/tox@feature/skip-whitespace-in-envstr
```

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -567,6 +567,11 @@ You can still list environments explicitly along with generated ones::
 
     envlist = {py26,py27}-django{15,16}, docs, flake
 
+Keep in mind that whitespace characters (except newline) within ``{}``
+are stripped, so the following line defines the same environment names::
+
+    envlist = {py26, py27}-django{ 15, 16 }, docs, flake
+
 .. note::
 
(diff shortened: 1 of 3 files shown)


asottile:
I think this is a good idea -- do we need any documentation updates associated with this?

andreif:
@asottile Updated docs.

obestwalter:
I have to admit, I don't quite why this sometimes happens. I guess it has to do with timing of collecting the results from different testruns.

PR #671, opened 2017-11-08
Fix pycodestyle E741 errors (#663)

### LINKED ISSUE #663: followup to #662 - Fix E741 (flake8)
I've temporarily turned this off in #662 such that we can get our build back to green.

We should go through and fix E741 (ambiguous variable name) -- the variable name in question appears to be `l` (and I agree, I have no idea what it means!)

cryvate:
Fix pycodestyle E741 errors and re-introduce testing for this in tox. In all these cases, it was ``l`` and it represented a ``list``. This is not a good name, so I tried to get something more descriptive. In one case, it was passed to a constructor which had ``list`` as an argument, so I decided to stick with that.

This fixes #663.

### DIFF
--- a/tox/result.py
+++ b/tox/result.py
@@ -63,8 +63,8 @@ class EnvLog:
             version=version)
 
     def get_commandlog(self, name):
-        l = self.dict.setdefault(name, [])
-        return CommandLog(self, l)
+        list = self.dict.setdefault(name, [])
+        return CommandLog(self, list)
 
     def set_installed(self, packages):
         self.dict["installed_packages"] = packages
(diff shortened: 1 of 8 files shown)


tox/result.py
```diff
@@ -63,8 +63,8 @@ def set_python_info(self, pythonexecutable):
             version=version)
 
     def get_commandlog(self, name):
-        l = self.dict.setdefault(name, [])
-        return CommandLog(self, l)
+        list = self.dict.setdefault(name, [])
```
asottile: `list` shadows the builtin `list` function, I'd pick a different name here

Issue #601, opened 2017-09-04
Adding -q to install_command does not make it avoid listing output of pip freeze


ssbarnea:
Adding `-q` to install_command does not make it avoid listing output of pip freeze, making the console output of tox very verbose, especially for people using lots of target environments.

Tried both approaches but without any luck.
```
install_command = pip install -q {opts} {packages}
list_dependencies_command = pip -q freeze > pip-freeze.txt
```

cryvate:
In the end, ``subprocess.Popen`` gets called with ``shell=False``, so no piping is possible. ``pip freeze`` ignores all arguments/options so it simply runs ``pip freeze`` to stdout.

This has the same root cause as issue #73 and so I think should be closed. Maybe there should be an explanation of this somewhere in the docs, if there isn't already? This is also the cause of e.g. issue #658. In terms of a work around, can I suggest the horrible:

    list_dependencies_command=bash -c \'pip -q freeze > pip-freeze.txt\'

obestwalter:
Hi @ssbarnea, as @cryvate already said: shell like behaviour like piping output is not supported and will very likely never be.

`list_dependencies_command` is only used for report purposes, so setting it to any command that yields no output is effectively shutting it up. Not pretty but does the job - e.g.:

    list_dependencies_command=echo

yields:

```
tox -r
python recreate: /home/oliver/work/tox/lab/601/.tox/python
python installdeps: flake8, pytest
python installed: 
python runtests: PYTHONHASHSEED='1918743262'
```

It would be simple to also make the <envname> installed part, of one would be inclined to open a PR for that.

cryvate:
@obestwalter: was that last sentence meant to read: "It would be simple to also make the installed part *disappear*, if one would be inclined to open a PR for that."? If so, I can have a go at that. Similarly for installdeps I guess (or even completely general, I will have a look).

I think though @ssbarnea point was 2 speared: not getting output on the command line (achieved) but secondly, he wanted to write the output of ``pip freeze`` to a file, which because ``commands`` has the same 'issue' cannot be fixed there either. So I think the issues at hand are:

- how does one get the output of pip freeze from a tox virtual environment?
- is this something one should be doing?

obestwalter:
> @obestwalter: was that last sentence meant to read: "It would be simple to also make the installed part disappear

Exactly - fixed :)

> how does one get the output of pip freeze from a tox virtual environment?

As part of the tox run only as part of the normal report.

> is this something one should be doing?

In my view not.

PR #673, opened 2017-11-18
Do not print empty headings

cryvate:
Headings like installed: <packages> will not be printed if there is no output to display after the :, unless verbosity is set.

Fixes issue #601.


cryvate:
Travis failed due to #656. All environments bar 3.4 passed.

gaborbernat:
rebase on master and it should be good now 👍

PR #674, opened 2017-11-18
always update setuptools before travis run

### LINKED ISSUE #656: Installation failures with setuptools 34 & 35

gaborbernat:
resolves #656

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
     language: generic
 
 before_install: if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then source .travis-osx; fi
-install: pip install --pre -U tox
+install: pip install six && pip install --pre -U tox
 script: tox -e py
 after_success:
   - tox -e coverage


gaborbernat:
@cryvate technically this is a pip issue, :-) and the underlying issue is already with six (which is a dependency of setuptools), so here I've treated just the bad part :-)

cryvate:
@gaborbernat thanks, I didn't realise that was happening, for others interested, see https://github.com/pypa/setuptools/issues/964

rpkilby:
Hi @gaborbernat - I think you need to add the `-U` flag to the six install. If you look at the build output, it shows that the requirement is already satisfied ([output](https://travis-ci.org/tox-dev/tox/jobs/304005231#L452-L453)). It doesn't upgrade until the upgrade of tox.

rpkilby:
As evidence of this, see this [build](https://travis-ci.org/tox-dev/tox/builds/304474808).

PR #684, opened 2017-11-28
Fixed running the internal tox test suite from a path containing spaces

jurko-gospodnetic:
Quoted `config.ini` command paths potentially containing spaces.

This avoids code like the code-coverage pytest plugin failing due to tests being run when the tox project's development folder path contains spaces.

Seen fail on Windows, so not 100% positive the change is ok for non-Windows platforms.


gaborbernat:
@jurko-gospodnetic can you redo this to fix the merge conflict

jurko-gospodnetic:
@gaborbernat - fix branch rebased on top of the current `master`

gaborbernat:
@jurko-gospodnetic yeah, mostly not works that works, need to investigate what's the issue with it.

PR #685, opened 2017-11-28
do not show failed py launcher call stderr output

jurko-gospodnetic:
This avoid stderr output like `Requested Python version (X.Y) not installed` when running tox (at least on Windows) and there is a test environment defined in the tox configuration file for which a suitable Python interpreter version is not installed on the system.

The error described above was occurring even when just running tox with the `--help` command-line option, and even if explicitly requesting a different test environment using the `-e` command-line option.


jurko-gospodnetic:
redoing the pull request due to a branch name typo

PR #690, opened 2017-11-29
Session: use realpath when ensuring that logdir exists

blueyed:
This allows for having a dangling symlink `.tox` (e.g. to somewhere into
`/tmp`), and the directory will be created instead of causing an error
(because it is not a dir, but exists already):

      File "/usr/lib/python3.6/site-packages/py/_error.py", line 66, in checked_call
        return func(*args, **kwargs)
    FileExistsError: [Errno 17] File exists: '…/src/covimerage/.tox'

    During handling of the above exception, another exception occurred:

      File "/usr/bin/tox", line 11, in <module>
        load_entry_point('tox==2.9.1', 'console_scripts', 'tox')()
      File "/usr/lib/python3.6/site-packages/tox/session.py", line 40, in main
      File "/usr/lib/python3.6/site-packages/tox/session.py", line 337, in __init__
        config.logdir.ensure(dir=1)
      File "/usr/lib/python3.6/site-packages/py/_path/local.py", line 530, in ensure
        return p._ensuredirs()
      File "/usr/lib/python3.6/site-packages/py/_path/local.py", line 512, in _ensuredirs
        parent._ensuredirs()
      File "/usr/lib/python3.6/site-packages/py/_path/local.py", line 515, in _ensuredirs
        self.mkdir()
      File "/usr/lib/python3.6/site-packages/py/_path/local.py", line 465, in mkdir
        py.error.checked_call(os.mkdir, fspath(p))
      File "/usr/lib/python3.6/site-packages/py/_error.py", line 86, in checked_call
        raise cls("%s%r" % (func.__name__, args))
    py.error.EEXIST: [File exists]: mkdir('…/src/covimerage/.tox',)

## Contribution checklist:

- [ ] added/updated test(s)
      in message body
- [ ] added news fragment in [changelog folder](https://github.com/tox-dev/tox/tree/master/changelog)

### DIFF
--- a/tox/session.py
+++ b/tox/session.py
@@ -336,7 +336,7 @@ class Session:
         self.resultlog = ResultLog()
         self.report = Report(self)
         self.make_emptydir(config.logdir)
-        config.logdir.ensure(dir=1)
+        config.logdir.realpath().ensure(dir=1)
         # self.report.using("logdir %s" %(self.config.logdir,))
         self.report.using("tox.ini: %s" % (self.config.toxinipath,))
         self._spec2pkg = {}


blueyed:
Came across this with a `tox2tmp` script I am experimenting with.

This creates (or moves an existing) `.tox` in `/tmp`.

```sh
#!/bin/sh

# Move/create .tox symlink below /tmp/tox/….

set -x

test -L .tox && exit

tmppath=/tmp/tox$PWD

if ! [ -d "$tmppath" ]; then
  mkdir -p "$tmppath"
fi

if [ -d .tox ]; then
  mv .tox "$tmppath"
fi

ln -sfn "$tmppath" .tox
```

asottile:
Note that your script will have other problems, notably that _moving_ `.tox` (or any parent directory) will break most executables in the virtualenvs contained (as the shebangs contain a full path to the interpreter used to install them).

As for the PR: you'll probably have additional problems with `tox -r` as well not playing well with your symlink.  I'm also not entirely comfortable with tox writing outside of the source directory.

Also curious about your motivation of moving the tox working directory -- is it related to shebang length limits (would be my motivation!)

And actually, I think your entire problem is solved by using `tox --workdir=/tmp/...`: see [this](http://tox.readthedocs.io/en/latest/example/jenkins.html?highlight=workdir#avoiding-the-path-too-long-error-with-long-shebang-lines)

blueyed:
@asottile 
Thank you!

My motivation is to use a tmpfs mainly (no cruft `.tox` dirs everywhere, and for performance).

I see that moving the `.tox` dir is problematic, but would then still like to have (different/deterministic) (dangling) symlinks - and therefore this PR still makes sense I guess?

An alternative might be to use a wrapper function/script around `tox` that uses the symlink's target as `--workdir` directly.. will consider this.

asottile:
might be a good candidate for a tox plugin?

blueyed:
@gaborbernat 
What work is needed here?  A test likely?!
Is the change acceptable in your opinion?

gaborbernat:
for the PR itself a test, a news fragment :+1: 

you still not elaborate if the ``--workdir`` is better solution. works for you or not?

blueyed:
Yes, `--workdir` is nice.
It can be used in a shell alias or function easily

```sh
tox() { command tox --workdir="/tmp/tox$PWD" "$@" }
```
or
```sh
alias tox='tox --workdir="/tmp/tox$PWD"'
```

A more advanced function (Zsh):
```zsh
toxini=(./(../)#tox.ini(NY1))
if (( $#toxini )); then
  toxdir=${toxini:h}/.tox
else
  toxdir=
  echo 'WARN: no tox.ini found.' >&2
fi
if ! [[ -e $toxdir ]]; then
  if ! [[ -L $toxdir ]]; then
    ln -s "/tmp/tox$PWD" $toxdir
  fi
fi
command tox --workdir="$toxdir" "$@"
```

blueyed:
Closing this for now - the shell wrapper is fine for me now.

PR #691, opened 2017-11-29
Minor optimization for SectionReader.getpath

blueyed:


blueyed:
Yes, not really an "optimization", but I just came across this when reading the code.

gaborbernat:
Thanks for the contribution @blueyed  👍

Issue #292, opened 2016-09-17
Negated factor conditions.


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/292
- Originally reported by: [mrmachine](http://bitbucket.org/mrmachine)
- Originally created at: 2015-12-04T14:45:54.405

I'd like to do something like:

```
[testenv]
commands =
    cov: coverage run ... runtests.py
    cov: coveralls
    !cov: python runtests.py
```

Currently I have to declare factor conditions for `cov` and `nocov` instead, which means users must always specific either `cov` or `nocov` with `tox -e`. If they forget, `tox` just reports success without running any command:

```
[testenv]
commands =
    cov: coverage run ... runtests.py
    cov: coveralls
    nocov: python runtests.py
```

jurko-gospodnetic:
Added pull request jurko-gospodnetic#1 for discussion. Not added to this repo as it's based on still unmerged test cleanup pull request #687, but I'll recreate it here as soon as that one is merged.

For now you can test it on that repo and leave the comments there.

jurko-gospodnetic:
Pull request #694 on this repo now up for discussion and the original one closed.

Issue #688, opened 2017-11-28
Support for negative factor conditionals


jurko-gospodnetic:
When playing around with nicely configuring tox for a project I ran into a scenario where I'd like to have a setting for an environment unless its name has some specific factor like `dev`. That would make it simpler to effectively remove a setting for specifically marked environments instead of having to duplicate all the other values for it or have to extract those values into a separate section and then reference them from there explicitly in multiple places.

Here's a simplified scenario:

A project that has a `[dev]` setuptools extra defined that adds a pytest dependency, which you want to use in `dev` environments, but in which case you do not want to add the regular `pytest` dependency to the tox `deps` setting, e.g. to test that the extra definition actually works and is getting used.
```
[tox]
envlist = {py27,py36}

[base]
deps = some-generic-packages

[testenv]
commands = pytest
deps =
    {[base]deps}
    pytest
extras = dev: dev
usedevelop = dev: True

[testenv:py27-dev]
deps = {[base]deps}

[testenv:py36-dev]
deps = {[base]deps}
```
while with negative factor selector support using a `!` prefix it can be rewritten to just:
```
[tox]
envlist = {py27,py36}

[testenv]
commands = pytest
deps =
    some-generic-package
    !dev: pytest
extras = dev: dev
usedevelop = dev: True
```

jurko-gospodnetic:
Added pull request https://github.com/jurko-gospodnetic/tox/pull/1 for discussion. Not added to this repo as it's based on still unmerged test cleanup pull request #687, but I'll recreate it here as soon as that one is merged.

The-Compiler:
Closing this as duplicate of #292 - really looking forward to that pull request though! 👍

jurko-gospodnetic:
Well the pull request is up, but for now on my tox project fork until #687 gets merged.

jurko-gospodnetic:
Pull request #694 on this repo now up and the original one closed.

jurko-gospodnetic:
Pull request #694 on this repo now up and the original one closed.

PR #694, opened 2017-12-01
allow defining negative conditional factors with a leading !

jurko-gospodnetic:
Pull request created for discussion on tox-dev/tox issues #292 and #688.

Still missing a doc/changelog/contributors changes, but the implementation & tests should be ok.

Suggested discussion points:
- negated factor operator other than `!`?
- any missing tests?

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1376,11 +1376,20 @@ class TestConfigTestEnv:
                 a: dep-a
                 b: dep-b
                 x: dep-x
+                !a: dep-!a
+                !b: dep-!b
+                !x: dep-!x
         """
         conf = newconfig([], inisource)
         configs = conf.envconfigs
-        assert [dep.name for dep in configs['a-x'].deps] == ["dep-all", "dep-a", "dep-x"]
-        assert [dep.name for dep in configs['b'].deps] == ["dep-all", "dep-b"]
+        expected = ["dep-all", "dep-a", "dep-x", "dep-!b"]
+        assert [dep.name for dep in configs['a-x'].deps] == expected
+        expected = ["dep-all", "dep-b", "dep-!a", "dep-!x"]
+        assert [dep.name for dep in configs['b'].deps] == expected
+        expected = ["dep-all", "dep-a", "dep-x", "dep-!b"]
+        assert [dep.name for dep in configs['a-x'].deps] == expected
+        expected = ["dep-all", "dep-b", "dep-!a", "dep-!x"]
+        assert [dep.name for dep in configs['b'].deps] == expected
 
     def test_factor_ops(self, newconfig):
         inisource = """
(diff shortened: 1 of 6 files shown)


jurko-gospodnetic:
This is a replacement for the original pull request https://github.com/jurko-gospodnetic/tox/pull/1 created in the fix repo before all the requirement pull requests got merged.

gaborbernat:
@jurko-gospodnetic the code style check failed

jurko-gospodnetic:
@gaborbernat - heh, yeah, as the pull request description said at the time: `and there might be fix-lint failures to clean up` ;)

Should be fixed now.

gaborbernat:
@asottile not sure what the pyupgrade means, but it failed.

asottile:
in this case it's probably rewriting the python<2.7 set call to a python2.7+ set comprehension

tests/test_config.py
```diff
@@ -1376,11 +1376,16 @@ def test_factors(self, newconfig):
...
                 x: dep-x
+                !a: dep-!a
+                !b: dep-!b
+                !x: dep-!x
         """
         conf = newconfig([], inisource)
         configs = conf.envconfigs
-        assert [dep.name for dep in configs['a-x'].deps] == ["dep-all", "dep-a", "dep-x"]
-        assert [dep.name for dep in configs['b'].deps] == ["dep-all", "dep-b"]
+        assert [dep.name for dep in configs['a-x'].deps] == \
```
gaborbernat: I prefer avoiding using the line separator - instead consider setting out the right hand side to a local variable e.g. ``exp_deps``.

tox/config.py
```diff
@@ -913,8 +913,20 @@ def _split_env(env):
 
 
 def _split_factor_expr(expr):
+    def split_single(e):
+        raw = e.split('-')
+        included = {x for x in raw if x[:1] != '!'}
+        excluded = {x[1:] for x in raw if x[:1] == '!'}
```
gaborbernat: this is a bit cryptic, consider instead using ``startswith``

tox/config.py
```diff
@@ -913,8 +913,20 @@ def _split_env(env):
 
 
 def _split_factor_expr(expr):
+    def split_single(e):
+        raw = e.split('-')
+        included = {x for x in raw if x[:1] != '!'}
```
gaborbernat: this is a bit cryptic, consider instead using ``not startswith`` to make it more explicit

tox/config.py
```diff
@@ -913,8 +913,20 @@ def _split_env(env):
...
+        return included, excluded
+
+    partial_envs = _expand_envstr(expr)
+    return [split_single(e) for e in partial_envs]
+
+
+def _split_factor_expr_all(expr):
     partial_envs = _expand_envstr(expr)
-    return [set(e.split('-')) for e in partial_envs]
+    return [{x[1:] if x[:1] == '!' else x for x in e.split('-')}
```
gaborbernat: I feel like we should have a ``is_negated_factor`` function to avoid all this repetition of this check

gaborbernat: the variable x is a bit too cryptic, e.g. the ``partial_envs`` hints that ``e`` will be a ``partial env``, but once I iterate on that what that contains?

tests/test_config.py
jurko-gospodnetic: ok, changing it to:
```
        expected = ["dep-all", "dep-a", "dep-x", "dep-!b"]
        assert [dep.name for dep in configs['a-x'].deps] == expected
        expected = ["dep-all", "dep-b", "dep-!a", "dep-!x"]
        assert [dep.name for dep in configs['b'].deps] == expected
```
Though I'd personally prefer:
```
        assert [dep.name for dep in configs['a-x'].deps] == [
            "dep-all", "dep-a", "dep-x", "dep-!b"]
        assert [dep.name for dep in configs['b'].deps] == [
            "dep-all", "dep-b", "dep-!a", "dep-!x"]
```

jurko-gospodnetic:
@gaborbernat  - Ok, I updated the code with the changes you requested.

- take a look at `_is_negated_factor()` and `_base_factor_name()` functions and where they are used in `tox/config.py`
- added 2 introductory doc cleanup commits not directly related to the negated factor conditionals feature
- added negated factor conditional related changelog entry
- updated docs with negated factor conditional related info

I briefly played with implementing the included/excluded factor splitting using something like this:
```
        split = dict(itertools.groupby(raw, key=_is_negated_factor))
        included = set(split[False])
        excluded = set(split[True])
```
which makes the code more concise but potentially too cryptic. In the end kept it as two dict comprehensions. Feel free to tweak as you like it.

When merging this, I suggest keeping the initial doc cleanup commits separate, but I did not feel like opening a pull request just for those.

jurko-gospodnetic:
Btw. all this condition checking actually leads very nicely to a more expressive grammar supporting:
- `,` for `or`
- `-` for `and`
- `!` for `not`
- and `(`/`)` parenthesis for grouping

but that change is not precluded by just adding the ! operator support so I did not want to get into that in this pull request.

jurko-gospodnetic:
I'm not really sure what you want added to `config-v2.rst`. That document does not contain full specification of how factor conditions are checked, as exists in `config.rst`. It seems like there's a lot of technical debt there that would need to be paid before extra documentation for this feature can be added and I don't really think all that should be done as a part of this pull request.

gaborbernat:
Resolves #292.

PR #700, opened 2017-12-06
improved the `initproj` test fixture code

jurko-gospodnetic:
- now can be passed a None `src_root` meaning the same thing as the '.', i.e. referring to the `tmpdir` folder
- added a check to make certain the passed `src_root` path is either relative or equal to tmpdir or one of its direct or indirect subfolders
- fixed generating the project's `__init__.py` script in case a non '.' `src_root` value is passed in (now can be avoided by passing prebuilt `__init__.py` content under that `src_root` folder in the `filedefs` structure)
- condensed code constructing `MANIFEST.in` content using list comprehension

No tests written for this as this is just test support code, and there are no tests actually broken by the current implementation, but had we had a test specifying e.g. both `src_root=a` and `{'a': {'project': {'__init__.py': ...}}}` as its `filedefs`, then the given `__init__.py` file content would not be applied and an internally generated `__init__.py` module would be used instead.

No docs exist about our test code.

No changelog entry added as this is not an external user facing change.

### DIFF
--- a/tests/test_pytest_plugins.py
+++ b/tests/test_pytest_plugins.py
@@ -1,0 +1,99 @@
+"""
+Test utility tests, intended to cover use-cases not used in the current
+project test suite, e.g. as shown by the code coverage report.
+
+"""
+import py.path
+import pytest
+
+from tox._pytestplugin import _filedefs_contains
+from tox._pytestplugin import _path_parts
+
+
+class TestInitProj:
+    @pytest.mark.parametrize('kwargs', (
+        {},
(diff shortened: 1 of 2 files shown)


gaborbernat:
@jurko-gospodnetic that's some non Python 2 like code there

jurko-gospodnetic:
Yup, just saw it. On it.

jurko-gospodnetic:
@gaborbernat - seems to be passing now :-)

Issue #701, opened 2017-12-07
pip upgrade support + better support for using `python -m pip` as the installer command


jurko-gospodnetic:
Default installation command used by `tox` actually runs pip's setuptools based runner executable (`pip.exe` on Windows), which then can not be removed or replaced while actively running on Windows.

A way towards a solution to this is to use a different installer command - `{envpython} -m pip install {opts} {packages}`, which should have the same effect but not use any executables to be replaced by a pip upgrade.

The main part of this is a relatively straightforward change, with just a bit of tweaking done on the test suite to adjust it to this change.

However, that runs into a bit of a problem with a slight difference between running `pip.exe` and `python -m pip`. The former adds the script folder to the Python path (`venv/Scripts` on Windows), while the latter adds the current working folder.

Unfortunately, adding the current working folder to the Python path is currently a big no-no for `tox`, As `tox` typically does its work from the target project's root folder, i.e. folder containing the  project's `tox.ini` configuration file, that means that this places that project root folder on the Python path for any Python processes run by `tox` in any of its Python environments, And since `tox`
in a typical project configuration also creates the project packages' `.egg-info` folders in that same project folder, that causes Python processes run by `tox` to automatically consider those packages as already installed in the target Python environment. This in turn makes various `pip` operations related to those packages fail, even their basic installations to various `tox` managed Python environments.

One way to work around this is to not hold our package sources in the root folder, but move them deeper to some `src` folder, which would make the `.egg-info` folders also be generated down in that `src` folder. This does avoid the problem, and can possibly be used for projects like those set up inside `tox`'s test suite, but does not seem feasible for general projects using `tox` out there in the wild which might really want to store their package folders directly inside their root project folder (and that is the typical suggested project structure).

Another way to work around it is to make tox not run its commands from the folder containing the `tox.ini` configuration file, but some temporary folder, e.g. `.tox` or `.tox/run-from-here`.

I can prepare a pull request for it but I assume the issue has come up before and someone must have had a good reason for not doing something like this before, so I'd like to hear what others think about this latter aproach? Anyone see any potential issues with it?

gaborbernat:
I think this has been fixed for now via switching to python -m tox. If not please say and I'll reopen.

Issue #102, opened 2016-09-17
Should handle pip requirements with relative paths


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/102
- Originally reported by: @generalov
- Originally created at: 2013-06-07T11:38:02.850

tox executes 'pip install' with changing cwd to envlogdir (.tox/py27/log)

If tox.ini takes deps from requirements.txt file and it contains relative paths installation will failed. 

requirements.txt

```
-e lib/python-package
```

tox.ini

```
[testenv]
deps = -r{toxinidir}/requirements.txt
```

/tmp/prj/$ tox -e py27 -v  

```
 ....  
/tmp/prj/.tox/py27/log$ /tmp/prj/.tox/py27/bin/pip install --download-cache=/tmp/.pip/cache -r/tmp/prj/requirements.txt >/tmp/prj/.tox/py27/log/py27-1.log  
ERROR: invocation failed, logfile: /tmp/prj/.tox/py27/log/py27-1.log  
ERROR: actionid=py27  
..  
lib/python-package should either by a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+
```

When I run same command from console installation success:

/tmp/prj/$ /tmp/prj/.tox/py27/bin/pip install --download-cache=/tmp/.pip/cache django==1.4 -r/tmp/prj/requirements.txt

I suggest tox should not change cwd when runs `pip install`

PR #707, opened 2017-12-09
[WIP] Fixed pip upgrades on Windows

jurko-gospodnetic:
This is a fix for issue #701.

### run install_commands from a `{toxinidir}/install_from_here` folder

This avoids any adverse interactions between the current folder and the installation command, e.g. `python -m pip` command will add the current folder to the Python path, and thus fail to install any package located in that folder for which there also exists a `.egg-info` folder in that same folder.

### change default install command to `python -m pip`

This allows tox to upgrade pip on Windows, where this fails if you use the `pip.exe` runner script as Windows does not allow removing an executable that is currently being executed.

## Open tasks

- [x] add news fragment in [changelog folder](https://github.com/tox-dev/tox/tree/master/changelog)
- [ ] see if there is anything in the documentation that can be updated related to this
- [ ] consider how this impacts relative links give in `deps` settings or in included `requirements.txt` files referenced using something like a `-rrequirements.txt` `deps` setting (see related issue #102)

## Things to discuss:

- [ ] I found no clean way to test pip upgrade on Windows as a part of our test suite as any such attempt runs into the Windows path length limit
- [ ] should we change the `{toxinidir}/install_from_here` folder to just `{toxinidir}` or should we perhaps make it configurable, and if so, should it be a global or an environment specific configuration setting or both

### DIFF
--- a/changelog/701.bugfix.rst
+++ b/changelog/701.bugfix.rst
@@ -1,0 +1,12 @@
+fix performing pip upgrades on Windows using the tox's default install_command
+- by @jurko-gospodnetic
+
+Changes:
+
+- default `install_command` changed to ``python -m pip`` instead of the
+  setuptools based ``pip`` executable
+- now the install commands are run from the ``{toxinidir}/install_from_here``
+  folder to avoid install commands (e.g. ``python -m pip``) accidentally picking
+  up ``.egg-info`` folders in the current working folder as indications of
+  already installed packages, thus causing it to potentially not install those
+  packages into the target environment
--- a/tests/test_venv.py
+++ b/tests/test_venv.py
@@ -140,7 +140,8 @@ def test_install_deps_wildcard(newmocksession):
     tox_testenv_install_deps(action=action, venv=venv)
     assert len(pcalls) == 2
     args = pcalls[-1].args
-    assert pcalls[-1].cwd == venv.envconfig.config.toxinidir
+    expected_cwd = venv.envconfig.config.toxinidir.join('install_from_here')
+    assert pcalls[-1].cwd == expected_cwd
     assert "pip" in str(args[0])
     assert args[1] == "install"
     args = [arg for arg in args if str(arg).endswith("dep1-1.1.zip")]
@@ -683,6 +684,8 @@ def test_run_install_command(newmocksession):
     assert 'install' in pcalls[0].args
     env = pcalls[0].env
     assert env is not None
+    expected_cwd = venv.envconfig.config.toxinidir.join('install_from_here')
+    assert pcalls[-1].cwd == expected_cwd
 
 
 def test_run_custom_install_command(newmocksession):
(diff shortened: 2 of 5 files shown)


jurko-gospodnetic:
As for the pip upgrade test, the following should have worked, had it not been for the Windows path length limit:

```
    def test_default_install_command_pip_upgrade(self, cmd, initproj):
        # will fail if using the setuptools based `pip.exe` starter executable
        # on Windows as upgrading pip will not be able to remove/replace the
        # currently running executable
        def test(pip_version):
            initproj('e', filedefs={
                'tox.ini': '''
                    [tox]
                    envlist = py27

                    [testenv]
                    install_command = python -m pip install {opts} {packages}
                    deps = pip==%s''' % (pip_version,)
            })
            result = cmd.run("tox", "-vvv")
            assert result.ret == 0
        # set up an environment with one pip version installed (will actually
        # replace the already installed pip version if the version we are
        # installing is not the default one)
        test('9.0.0')
        # now upgrade the environment to a different pip version (will replace
        # the already installed pip version in case the first run did not)
        test('9.0.1')
```

gaborbernat:
Any reasons the coverage decreased?

jurko-gospodnetic:
Haven't got a clue. Will take a look at the report as soon as I get all the kinks causing Travis test failures out of the test code.

jurko-gospodnetic:
I think the coverage report got scr*w*d up. I can see no actual change in those links it gives us. Could you re-run it? Perhaps it got confused by me rebasing the branch while its original scan was already in progress?

jurko-gospodnetic:
And it seems it changed its story now and is actually showing improved test coverage even for this pull request, yay! :-)

gaborbernat:
In order for this to not become breaking change, can we consider what effect would this have on relative source tree installs? 

E.g. 
```
deps = module_1\\
            module_2\\
```

jurko-gospodnetic:
Could you explain a bit more what you mean by that? I'm not sure I understand completly what you mean by 'relative source tree installs'.

The closest thing I could find was an entry in the changelog for making installs be done from the `{toxinidir}` in the first place and its related issue #102 - is that what you were referring to?

gaborbernat:
Yes exactly👌

gaborbernat:
Also I think the folder name should be ``.tox/.pip``

jurko-gospodnetic:
well `pip` is not the only possible install command though... :-) `easy_install`? something else could come up like someone's custom wrapper script

gaborbernat:
``.install`` is okay, but I think #102 was the reason why we initially had this to toxworkdir. How can we work around that ?

gaborbernat:
@jurko-gospodnetic also please make this configurable, but the default being ``.tox/.install`` 👍 so users can change back if needed

jurko-gospodnetic:
@gaborbernat - as for the configuration - how do you want it? a single global option? an environment option? both with environment one overriding the global one?

Also, could you suggest appropriate option names?

gaborbernat:
I think this is an environment bound settings. ``install_work_dir`` seems appropriate (defaults to ``{toxworkdir}\.install``

gaborbernat:
@jurko-gospodnetic want to pick this up and finish it? thanks!

gaborbernat:
I'll close this now for lack of activity; feel free to re-open it if you pick up again.

PR #708, opened 2017-12-10
trivial interpreters py touchups and test coverage improvements

jurko-gospodnetic:
Trivial interpreters py touchups and test coverage improvements.

### DIFF
--- a/changelog/708.trivial.rst
+++ b/changelog/708.trivial.rst
@@ -1,0 +1,1 @@
+touched up ``interpreters.py`` code and added some missing tests for it - by @jurko-gospodnetic
--- a/tests/test_interpreters.py
+++ b/tests/test_interpreters.py
@@ -7,8 +7,13 @@ import py
 import pytest
 
 from tox.config import get_plugin_manager
+from tox.interpreters import ExecFailed
+from tox.interpreters import InterpreterInfo
 from tox.interpreters import Interpreters
+from tox.interpreters import NoInterpreterInfo
+from tox.interpreters import pyinfo
 from tox.interpreters import run_and_get_interpreter_info
+from tox.interpreters import sitepackagesdir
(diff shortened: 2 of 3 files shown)


gaborbernat:
hi, the lint fix failed 👍

jurko-gospodnetic:
doh... I hate that 'reorder imports' failure... wish it would tell me what it wants them to be reordered to...

I really have to get fix-lint to run on my machine...

gaborbernat:
@asottile I feel the same as @jurko-gospodnetic  the ``pre-commit`` often gives fairly cryptic errors; also i would prefer to group together imports (instead one per line)

jurko-gospodnetic:
Ok, somehow hacked together `fix-lint` to work on my machine (that whole thing is totally unfriendly to my setup... 😞) and finally found the culprit reordering... seems the ordering it wants is case-sensitive :-)

asottile:
The brevity of the error messages is intentional. The general guidance is : if it gets automatically fixed there's no need to spam a bunch of text, the filename is sufficient. But as it's apparent that you're not running them locally, I'm adding this so you can at least see the diff : https://github.com/tox-dev/tox/pull/710

As for the imports: it's impossible to agree on style when there are this many maintainers so we opted for a style that eliminates merge conflicts. If you're unhappy with this let's create an issue and discuss the trade offs of removing the import ordering entirely -- though I imagine this will just make things worse ;)

gaborbernat:
@asottile it does eliminate the probability of merge conflict, but increases the number of lines 👍 I'm not for removing it, just tweaking it. I would prefer to have something that works out of box, e.g. @jurko-gospodnetic struggle to get the fix-lint working is definitely what I was afraid of.

gaborbernat:
Thanks for the contribution @jurko-gospodnetic  👍

PR #709, opened 2017-12-10
pass the `PROGRAMDATA` envrionment variable when invoking fix-lint

jurko-gospodnetic:
Pass the `PROGRAMDATA` environment variable when invoking `fix-lint`, fixing `error setting certificate verify locations` errors when cloning a git repo using git for Windows.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,6 +28,9 @@ commands = sphinx-build -d {toxworkdir}/docs_doctree doc {toxworkdir}/docs_out -
 basepython = python3.6
 passenv = {[testenv]passenv}
           HOMEPATH
+          # without PROGRAMDATA cloning using git for Windows will fail with an
+          # `error setting certificate verify locations` error
+          PROGRAMDATA
 extras = lint
 description = run static analysis and style check using flake8
 commands = python -m flake8 --show-source tox setup.py {posargs}


jurko-gospodnetic:
Ok, this coverage report is getting ridiculous :-)

gaborbernat:
I think it's just you need to wait for all the envs to finish👌

asottile:
Environment variables are case insensitive so the general consensus is to use all-caps: PROGRAMDATA

changelog/709.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+pass the ``ProgramData`` environment variable when invoking the internal
+development ``fix-lint`` toolset, fixing ``error setting certificate verify
+locations`` errors when cloning a git repo using git for Windows - by
+@jurko-gospodnetic
```
asottile: Since this is an internal only change (and not a bug in tox itself) it should not have a bug fix changeling entry

The-Compiler:
> Environment variables are case insensitive so the general consensus is to use all-caps: PROGRAMDATA

I wouldn't say so, e.g. [this page](https://ss64.com/nt/syntax-variables.html) lists it as `ProgramData` as well.

gaborbernat:
Windows handles environment variables case insensitive, Linux does not. I would consider the windows behavior more a bug than a feature, so let's roll with the Linux behavior👌

changelog/709.bugfix.rst
gaborbernat: yeah, remove the changelog please 👍

jurko-gospodnetic:
@gaborbernat - Variable named `PROGRAMDATA` now.

changelog/709.misc.rst
```diff
@@ -0,0 +1,4 @@
+pass the ``PROGRAMDATA`` environment variable when invoking the internal
```
gaborbernat: this is a development environment change for tox, we don't need a changelog entry for this as servers no value for end user

jurko-gospodnetic:
@gaborbernat - changelog entry removed

PR #722, opened 2018-01-03
Upgrade pre-commit: better python detection on windows

asottile:


asottile:
see [the changelog](https://github.com/pre-commit/pre-commit/blob/master/CHANGELOG.md#143)

PR #726, opened 2018-01-05
Fix generation of JSON files with --skip-missing-interpreters option. Fixes #672 

### LINKED ISSUE #672: --result-json does not respect --skip-missing-interpreters-option
When executing tox with --result-json and --skip-missing-interpreters-option, the latter seems not to be reflected in the generated json file. Please find the sample project attached below:
[missing_interpreters.tar.gz](https://github.com/tox-dev/tox/files/1474831/missing_interpreters.tar.gz)

To reproduce the issue simply invoke:

`tox --skip-missing-interpreters --result-json toxreport.json`
Tox main process exits with return code 0, but in toxreport.json file user can see the following:
```
"python": {
  "setup": [
  {
    "command": [
    "setup virtualenv"
    ],
    "output": "InterpreterNotFound: xyz_unknown_interpreter",
    "retcode": "1"
   }
   ]
}
```
In my opinion retcode here should be changed to 0 if --skip-missing-interpreters option is explicitly used.

r2dan:
Fixes execution of tox with --result-json and --skip-missing-interpreters-option. The reported JSON file contains return code 0 if no interpreter is found. See example output below. Fixes #672

```
"python": {
  "setup": [
  {
    "command": [
    "setup virtualenv"
    ],
    "output": "InterpreterNotFound: xyz_unknown_interpreter",
    "retcode": "0"
   }
   ]
}
```

### DIFF
--- a/tox/session.py
+++ b/tox/session.py
@@ -490,9 +490,9 @@ class Session:
                         "not supported by virtualenv. Error details: %r" % e)
             except tox.exception.InvocationError as e:
                 status = (
-                    "Error creating virtualenv. Note that some special "
-                    "characters (e.g. ':' and unicode symbols) in paths are "
-                    "not supported by virtualenv. Error details: %r" % e)
+                        "Error creating virtualenv. Note that some special "
+                        "characters (e.g. ':' and unicode symbols) in paths are "
+                        "not supported by virtualenv. Error details: %r" % e)
             except tox.exception.InterpreterNotFound as e:
                 status = e
                 if self.config.option.skip_missing_interpreters:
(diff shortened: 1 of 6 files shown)


matthiasha:
This will make Devpi report those tests as "passed", right?

r2dan:
@matthiasha: Yes, setup of those test environments will be "passed" on devpi.

tox/session.py
```diff
@@ -489,14 +490,21 @@ def setupenv(self, venv):
                         "not supported by virtualenv. Error details: %r" % e)
             except tox.exception.InvocationError as e:
                 status = (
-                        "Error creating virtualenv. Note that some special "
-                        "characters (e.g. ':' and unicode symbols) in paths are "
-                        "not supported by virtualenv. Error details: %r" % e)
+                    "Error creating virtualenv. Note that some special "
```
gaborbernat: this change seems unrelated here

r2dan: My IDE made that change - I'm reverting it back.

Issue #725, opened 2018-01-05
[FEATURE] Add support for `envreport` action to be pluggable


tonybaloney:
Requesting (and willing to submit a PR) that the plugin system support a hookimpl for the envreport action.

I'm developing a plugin for `Pipenv` https://github.com/tonybaloney/tox-pipenv using the hook implementation API in 2.9.1. 

So far it's working nicely, however as stated in the docs, the hook implementation options are not complete.

Whilst the default option for list_dependencies_command can be overridden at runtime, I want the plugin to be self-contained. Also, setting the configuration option when the plugin ignores it seems a little obtuse.

https://github.com/tox-dev/tox/blob/master/tox/config.py#L595-L599

tonybaloney:
Raised PR with potential implementation

asottile:
I think you can do this without needing new apis.  For example, here's how `tox-pip-extensions` sets the `install_cmd`: https://github.com/tox-dev/tox-pip-extensions/blob/a989ec95ef8f3c5e63bb809308cd4260f0690e76/tox_pip_extensions.py#L71

PR #730, opened 2018-01-08
Package stat plugin

tonybaloney:
This pull request implements a plugin hook for reporting package versions as described in #725 

Implementation simply moves existing code into a static method and refactors existing calls to use hookimpl.

The expected result not a list, since `first_result` was used, I decided to match the behaviour of the installation of packages.

### DIFF
--- a/tox/hookspecs.py
+++ b/tox/hookspecs.py
@@ -103,3 +103,11 @@ def tox_runtest_post(venv):
 
     This could be used to have per-venv test reporting of pass/fail status.
     """
+
+
+@hookspec(firstresult=True)
+def tox_runenvreport(venv, action):
+    """ [experimental] Get the installed packages and versions in this venv
+
+    This could be used for alternative (ie non-pip) package managers
+    """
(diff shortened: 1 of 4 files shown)


gaborbernat:
@tonybaloney the CI fails sadly

tox/hookspecs.py
```diff
@@ -103,3 +103,11 @@ def tox_runtest_post(venv):
 
     This could be used to have per-venv test reporting of pass/fail status.
     """
+
+
+@hookspec(firstresult=True)
+def tox_runenvreport(venv, action):
+    """ [experimental] Get the installed packages and versions in this venv
+
+    This could be used for alternative (ie non-pip) package managers
```
nicoddemus: It would be nice to document that this hook should return a list of strings.

gaborbernat: indeed @tonybaloney can you add that part

tonybaloney: done in f72dd22

PR #738, opened 2018-01-10
Prevent test failure due to Travis pyenv misconfiguration

rpkilby:
Travis seems to occasionally misconfigure pyenv (see: travis-ci/travis-ci#8363). Previously, `python3.5` was misconfigured, but a more recent deployment seems to have broken `python3.6` instead. Per @asottile's [comment](https://github.com/tox-dev/tox-venv/pull/4), `python3` is a little more robust in this regard, and less dependent on one specific version of python being misconfigured.

### DIFF
--- a/tests/test_venv.py
+++ b/tests/test_venv.py
@@ -385,11 +385,11 @@ def test_install_command_not_installed_bash(newmocksession):
 
 
 def test_install_python3(tmpdir, newmocksession):
-    if not py.path.local.sysfind('python3.6'):
-        pytest.skip("needs python3.6")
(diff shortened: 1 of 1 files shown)


asottile:
should the patch be `python3` instead of `python3.6` (not sure if you committed the wrong patch!)

rpkilby:
Yes - I was submitting this patch first, to ensure that `python3.6` *does* fail. Will followup with `python3` once tests fail.

rpkilby:
Or if they don't fail, bang head on desk.

PR #743, opened 2018-01-14
Change the logo/fav-icon

hazalozturk:
Changed the logo based on the issue #639


gaborbernat:
@hazalozturk can you update these in our documentation too? thanks a lot!!!

hazalozturk:
@gaborbernat It looks like it is already updated. Do I need to do anything else?

gaborbernat:
@hazalozturk it seems it's working now, probably was just cached in locally for me, thanks!

Issue #752, opened 2018-01-27
test_alwayscopy fails on openSUSE


ederag:
On openSUSE 42.2 (python2.7 or python3.4), `test_alwayscopy` fails.
This seems due to a `virtualenv` issue (https://github.com/pypa/virtualenv/issues/565).
So this is more a report for anybody searching for this error + "tox".

Here is the log:
```
tox -e py34    
GLOB sdist-make: /home/ederag/share/prog/python/tox/setup.py
py34 inst-nodeps: /home/ederag/share/prog/python/tox/.tox/dist/tox-2.9.1.zip
py34 installed: apipkg==1.4,attrs==17.4.0,coverage==4.4.2,execnet==1.5.0,pluggy==0.6.0,py==1.5.2,pytest==3.3.2,pytest-cov==2.5.1,pytest-forked==0.2,pytest-timeout==1.2.1,pytest-xdist==1.22.0,six==1.11.0,tox==2.9.1,virtualenv==15.1.0
py34 runtests: PYTHONHASHSEED='1275035432'
py34 runtests: commands[0] | pytest tests --cov-config=/home/ederag/share/prog/python/tox/tox.ini --cov=/home/ederag/share/prog/python/tox/.tox/py34/lib/python3.4/site-packages/tox --timeout=180
========================================================= test session starts ==========================================================
platform linux -- Python 3.4.6, pytest-3.3.2, py-1.5.2, pluggy-0.6.0
tox comes from: '/home/ederag/share/prog/python/tox/.tox/py34/lib/python3.4/site-packages/tox/__init__.py'
rootdir: /home/ederag/share/prog/python/tox, inifile: tox.ini
plugins: xdist-1.22.0, timeout-1.2.1, forked-0.2, cov-2.5.1
timeout: 180.0s method: signal
collected 303 items                                                                                                                    

tests/test_config.py ........................................................................................................... [ 35%]
.................................................................x............s                                                  [ 61%] 
tests/test_interpreters.py s......                                                                                               [ 63%]
tests/test_quickstart.py ................                                                                                        [ 68%]
tests/test_result.py ....                                                                                                        [ 70%]
tests/test_venv.py ...................s..................                                                                        [ 82%]
tests/test_z_cmdline.py ..................................F.................                                                     [100%]

----------- coverage: platform linux, python 3.4.6-final-0 -----------
Name                                                         Stmts   Miss Branch BrPart  Cover
----------------------------------------------------------------------------------------------
[...]

PR #753, opened 2018-01-27
Skip alwayscopy tests for SuSE

ederag:
This is a fix to issue #752.
Now `tox` succeeds (with python versions: py27, py34, py36).
*a priori* the i686 arch should be able to run, so skip `test_alwayscopy`  only for the x86-64 arch.

### DIFF
--- a/tests/test_z_cmdline.py
+++ b/tests/test_z_cmdline.py
@@ -636,10 +636,8 @@ def _alwayscopy_not_supported():
     if hasattr(platform, 'linux_distribution'):
         _dist = platform.linux_distribution(full_distribution_name=False)
         (name, version, arch) = _dist
-        if any((
-                name == 'centos' and version[0] == '7',
-                name == 'SuSE' and arch == 'x86_64',
-               )):
+        if any(( name == 'centos' and version[0] == '7',
+                 name == 'SuSE' and arch == 'x86_64' )):
             return True
     return False
 
--- a/tests/test_z_cmdline.py
+++ b/tests/test_z_cmdline.py
@@ -636,8 +636,8 @@ def _alwayscopy_not_supported():
     if hasattr(platform, 'linux_distribution'):
         _dist = platform.linux_distribution(full_distribution_name=False)
         (name, version, arch) = _dist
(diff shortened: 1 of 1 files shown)


tests/test_z_cmdline.py
```diff
@@ -635,7 +635,11 @@ def _alwayscopy_not_supported():
     # see: https://github.com/pypa/virtualenv/issues/565
     if hasattr(platform, 'linux_distribution'):
         _dist = platform.linux_distribution(full_distribution_name=False)
-        if _dist[0] == 'centos' and _dist[1][0] == '7':
+        (name, version, arch) = _dist
+        if (
```
gaborbernat: this formatting looks akward, maybe instead use the ``any`` with two args, one per line :+1:

ederag: Nice trick !

tests/test_z_cmdline.py
```diff
@@ -635,7 +635,11 @@ def _alwayscopy_not_supported():
     # see: https://github.com/pypa/virtualenv/issues/565
     if hasattr(platform, 'linux_distribution'):
         _dist = platform.linux_distribution(full_distribution_name=False)
-        if _dist[0] == 'centos' and _dist[1][0] == '7':
+        (name, version, arch) = _dist
+        if any((
```
gaborbernat: what I meant 
``` python
        if any(( name == 'centos' and version[0] == '7',
                 name == 'SuSE' and arch == 'x86_64' )):
```

ederag: The current state is closer to [this pep8 recommendation](https://www.python.org/dev/peps/pep-0008/#id29) (the first example). This makes suppressions very clear to read in the logs. (1 suppression: 1 line).
But just confirm your preference and I'll update accordingly.

gaborbernat: I tend to disagree with that, so I would go to change it to be more readable :thinking:

ederag: Well, then this is a matter of taste ;-) 
So your code was copy/pasted, but failed `fix-lint`. Had to remove few spaces.

PR #757, opened 2018-01-30
Add example symptomatic output for the described issue

tom-dalton-fanduel:
Add an example of what this error might look like when running tox/pytest.

I had this problem but it took me a long time to find this page. Searching for the error message brings back a lot of results but they're mostly not tox-related and even the discussions of this issue wrt. pytest didn't make it super-obvious there are fairly straighforward fixes possible (as documented in the page I've amended).

My hope is that by adding the example error output, this page becomes a higher result for search terms for the error message along with 'tox'. It might also stand out more to people that have the error without necessarily realising that the solutions here apply to them.

### DIFF
--- a/doc/example/pytest.rst
+++ b/doc/example/pytest.rst
@@ -98,6 +98,7 @@ installed package.
 This issue may be characterised by pytest test-collection error messages, in python 3.x environments, that look like:
 
 .. code-block:: shell
+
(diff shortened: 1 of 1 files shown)


tom-dalton-fanduel:
Thanks to Travis for noticing my sloppy formatting, fixed.

Issue #290, opened 2016-09-17
Display exit code with InvocationErrors


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/290
- Originally reported by: @blueyed
- Originally created at: 2015-11-25T13:57:56.481

When a command fails, you currently only get `ERROR: InvocationError:
'/usr/bin/false'`, but the exitcode itself is useful/important, e.g. with
py.test using 5 for "no tests collected".

The following patch would provide that, by using a custom `__str__` method for
InvocationError - but it does not handle the case where that might be missing (not all instances / calls to it provide it).

Maybe a custom `__init__` should be used to set the arguments as properties explicitly, and then use them?!

```
diff -r e9569646da4f tox/__init__.py
--- a/tox/__init__.py   Wed Nov 25 12:27:48 2015 +0100
+++ b/tox/__init__.py   Wed Nov 25 13:23:26 2015 +0100
@@ -17,6 +17,9 @@
         "signals that an interpreter could not be found"
     class InvocationError(Error):
         """ an error while invoking a script. """
+        def __str__(self):
+            return "%s: %s (exitcode %d)" % (self.__class__.__name__,
+                                             self.args[0], self.args[1])
     class MissingFile(Error):
         """ an error while invoking a script. """
     class MissingDirectory(Error):
```

This is related / similar to #192.

obestwalter:
@blueyed - still interested to make this happen?

blueyed:
@obestwalter 
In general, yes - but not in the near future.
Feel free to pick up the patch(es) from here, of course.

obestwalter:
@blueyed thanks - I might try to crystalize your drops of genius in this issue in a PR then :)

ederag:
This is useful, thanks !
The patch https://github.com/tox-dev/tox/issues/290#issuecomment-247791000 did not apply to current repo. Here is an updated one (only the `@@` line changed).
```
diff --git a/tox/__init__.py b/tox/__init__.py
index b7a1631..d872b32 100644
--- a/tox/__init__.py
+++ b/tox/__init__.py
@@ -34,6 +34,19 @@ class exception:
 
     class InvocationError(Error):
         """ an error while invoking a script. """
+        def __init__(self, *args):
+            super(exception.Error, self).__init__(*args)
+            self.command = args[0]
+            try:
+                self.exitcode = args[1]
+            except IndexError:
+                self.exitcode = None
+
+        def __str__(self):
[...]

PR #760, opened 2018-02-02
Display exit code with InvocationErrors

ederag:
This is a PR for @blueyed patch (Issue #290), supplemented with tests. 
For exit codes larger than 128, a hint is given: on unix systems this often means a fatal error.
(Not always, since an application might exit with a large custom error code)

Not showing this hint for windows platform was considered, but discarded to keep code short and
simplify tests.

Before:

> ERROR: InvocationError: '/home/ederag/share/prog/python/tox/.tox/exitcode/bin/python3.6 -c import sys; sys.exit(5)'

After:

> ERROR: InvocationError for command '/home/ederag/share/prog/python/tox/.tox/exitcode/bin/python3.6 -c import sys; sys.exit(5)' (exited with code 5)

or, for an exit code larger than 128:

> ERROR: InvocationError for command '/home/ederag/share/prog/python/tox/.tox/exitcode/bin/python3.6 -c import sys; sys.exit(129)' (exited with code 129)
> Note: On unix systems, an exit code larger than 128 often means a fatal error (e.g. 139=128+11: segmentation fault)

To generate these outputs, a section was added to `tox.ini`:
```
[testenv:exitcode]
basepython = python3.6
description = commands with several exit codes
skip_install = True
commands = python3.6 -c "import sys; sys.exit(5)"
```
and run with `python3 -m tox -e exitcode`.
The modified section is too specific to be included in the PR, isn't it ?

Tests have been added to `test_z_cmdline.py`.
Safely replacing these tests with faster ones (mock) would require someone experienced with this technique.

The documentation might be completed (still have to think about it).
Probably add a section "Understanding InvocationError exit codes"
in [example/general.html](https://tox.readthedocs.io/en/latest/example/general.html)
with links to `pytest` exit codes, to http://www.faqs.org/docs/abs/HTML/exitcodes.html, 
and explaining that no exit code is shown when the process has been aborted (like Qt does).
And for good measure, giving a link to [ignoring-a-command-exit-code](https://tox.readthedocs.io/en/latest/example/basic.html#ignoring-a-command-exit-code).

## Contribution checklist:

- [x] added/updated test(s)
      in message body
- [x] added news fragment in [changelog folder](https://github.com/tox-dev/tox/tree/master/changelog)

### DIFF
--- a/tox/__init__.py
+++ b/tox/__init__.py
@@ -34,13 +34,10 @@ class exception:
 
     class InvocationError(Error):
         """ an error while invoking a script. """
-        def __init__(self, *args):
-            super(exception.Error, self).__init__(*args)
-            self.command = args[0]
-            try:
-                self.exitcode = args[1]
-            except IndexError:
-                self.exitcode = None
+        def __init__(self, command, exitcode=None):
+            super(exception.Error, self).__init__(command, exitcode)
+            self.command = command
+            self.exitcode = exitcode
 
         def __str__(self):
             str_ = "%s for command %s" % (self.__class__.__name__, self.command)
--- a/tox/__init__.py
+++ b/tox/__init__.py
@@ -43,10 +43,13 @@ class exception:
                 self.exitcode = None
 
         def __str__(self):
+            str_ = "%s for command %s" % (self.__class__.__name__, self.command)
             if self.exitcode:
-                return "%s: %s (exitcode %d)" % (self.__class__.__name__,
-                                                 self.command, self.exitcode)
-            return "%s: %s" % (self.__class__.__name__, self.command)
+                str_ += " (exited with code %d)" % (self.exitcode)
+                if self.exitcode > 128:
+                    str_ += ("\nNote: On unix systems, an exit code larger than 128 "
+                             "often means a fatal error (e.g. 139=128+11: segmentation fault)")
+            return str_
 
     class MissingFile(Error):
         """ an error while invoking a script. """
--- a/tox/__init__.py
+++ b/tox/__init__.py
@@ -34,6 +34,19 @@ class exception:
 
     class InvocationError(Error):
         """ an error while invoking a script. """
+        def __init__(self, *args):
+            super(exception.Error, self).__init__(*args)
+            self.command = args[0]
+            try:
+                self.exitcode = args[1]
+            except IndexError:
+                self.exitcode = None
+
+        def __str__(self):
+            if self.exitcode:
+                return "%s: %s (exitcode %d)" % (self.__class__.__name__,
+                                                 self.command, self.exitcode)
+            return "%s: %s" % (self.__class__.__name__, self.command)
 
     class MissingFile(Error):
         """ an error while invoking a script. """
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -16,6 +16,7 @@ Chris Jerdonek
 Chris Rose
 Clark Boylan
 Cyril Roelandt
+Ederag
 Eli Collins
 Eugene Yunak
 Fernando L. Pereira
--- a/changelog/290.feature.rst
+++ b/changelog/290.feature.rst
@@ -1,0 +1,1 @@
+``tox`` displays exit code together with ``InvocationError`` - by @blueyed and @ederag.
--- a/doc/example/basic.rst
+++ b/doc/example/basic.rst
@@ -289,6 +289,7 @@ Integration with "setup.py test" command
   point it's maybe best to not go for any ``setup.py test`` integration.
 
 
+.. _`ignoring exit code`:
 
 Ignoring a command exit code
 ----------------------------
--- a/doc/example/general.rst
+++ b/doc/example/general.rst
@@ -204,3 +204,49 @@ There is an optimization coded in to not bother re-running the command if
 ``$projectname.egg-info`` is newer than ``setup.py`` or ``setup.cfg``.
 
 .. include:: ../links.rst
+
+
+Understanding ``InvocationError`` exit codes
+--------------------------------------------
+
+When a command (defined by ``commands =`` in ``tox.ini``) fails,
+it has a non-zero exit code,
+and an ``InvocationError`` exception is raised by ``tox``:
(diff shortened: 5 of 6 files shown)


tox/__init__.py
```diff
@@ -34,6 +34,22 @@ class InterpreterNotFound(Error):
 
     class InvocationError(Error):
         """ an error while invoking a script. """
+        def __init__(self, *args):
+            super(exception.Error, self).__init__(*args)
+            self.command = args[0]
+            try:
+                self.exitcode = args[1]
+            except IndexError:
+                self.exitcode = None
```
asottile: I think this would be a little more clear as:

```python
        def __init__(self, command, exitcode=None):
            super(Error, self).__init__(command, exitcode)
            self.command = command
            self.exitcode = exitcode
```

tox/__init__.py
```diff
@@ -34,6 +34,22 @@ class InterpreterNotFound(Error):
...
+
+        def __str__(self):
+            str_ = "%s for command %s" % (self.__class__.__name__, self.command)
+            if self.exitcode:
+                str_ += " (exited with code %d)" % (self.exitcode)
+                if self.exitcode > 128:
+                    str_ += ("\nNote: On unix systems, an exit code larger than 128 "
+                             "often means a fatal error (e.g. 139=128+11: segmentation fault)")
```
asottile: >128 means it exited due to a fatal signal

You can see the various signals that might trigger this in the table at the bottom of [this](https://en.wikipedia.org/wiki/Signal_(IPC)#POSIX_signals)

We could also compute this and show it to the user:

```console
$ python -c 'import time; time.sleep(100)'
^\Quit (core dumped)
$ echo $?
131
$ python -c '
> import signal
> sigs = {n: s for s, n in vars(signal).items() if s.startswith("SIG")}
> print(sigs[131 - 128])
> '
SIGQUIT
```

A `KeyError` on `sigs` (in my example) means it's probably not a fatal signal that killed it (also note that although `^C` sends SIGINT -- python sets a signal handler for this and reraises it as `KeyboardInterrupt` leading to an exit code of `1` instead of the expected `130`)

tox/__init__.py
ederag: Thanks a lot. Just had to use `exception.Error`.

asottile: oops yeah, sorry my sample wasn't completely copy-pastable 😆

tox/__init__.py
ederag: Thanks for the link, added to the documentation that was in preparation.
Giving the signal value/name could be nice, since python uses directly the relevant `signals.h`; got to think about it.

asottile:
@gaborbernat I had an outstanding issue, I don't think this was ready for merge

gaborbernat:
@asottile which one?

asottile:
I suggested an alternate wording and some additional helpful output [here](https://github.com/tox-dev/tox/pull/760#discussion_r165831005)

ederag:
TL;DR: done, works, but could be better.
The [commit](https://github.com/ederag/tox/commit/bf9202ac267d461d9e8890e93605023b1ad95a4f) does not appear. Is this PR broken ?

About the proposed wording "means it exited due to a fatal signal", it is true that "signal" is more precise than "error". Finally, the wording "fatal error signal" of [faqs.org](http://www.faqs.org/docs/abs/HTML/exitcodes.html), seems good (especially for segmentation fault, since it originates from an error in the program, in the first place).

The note now gives the corresponding signal if one is found, hopefully matching your suggestion.
Issuing that message only for unix would require mocking `sys.platform` in tests.
Note sure it is worth the trouble though. And it would require [`pytest.mock`](https://pypi.python.org/pypi/pytest-mock).

Now testing the `InvocationError` itself is done more extensively, and much faster, in `test_result.py`. Obviously, it is not a `tox.result`, but kind of a result. Moreover, no other file (except the slow `test_z_cmdline.py`) seemed adequate.
That `InvocationError ` is actually called or not, is still tested in `test_z_cmdline.py`. These tests look for a specific string in the tox output.
Two tests - exit code 0 or non-zero - remain. To be faster, it would be possible to use both commands in the same test. But it seemed safer to keep these two separate.
One fast and safe way would have been to use `mock`, but it did not work because [exceptions can not be mocked like that](https://stackoverflow.com/a/31873937/3565696). One way to circumvent that would be to create a separate *function* `tox.exceptions.InvocationError._str(command, exitcode)`, called from `tox.exceptions.InvocationError.__str__`, and use the [`pytest.mock`](https://pypi.python.org/pypi/pytest-mock) `spy` on the `_str` function only.
This would allow to gain about 3 second of test time for each python version tested.

Questions:
Just in case, would it be acceptable to add [`pytest.mock`](https://pypi.python.org/pypi/pytest-mock) to the `extra_require['testing']` in `setup.py` ?

Finally, a minor question from the initial post is still pending: 

> The modified section is too specific to be included in the PR, isn't it ?

gaborbernat:
@ederag  I think it's because the PR is closed, please open a new one :+1:

ederag:
Waiting for revert-760-exitcode to be merged into master. Glad I checked before pulling. This is a bit weird. Shouldn't it be automatic when the revert button is used ? Why can't we keep the current PR ? A github limitation or bug ?

gaborbernat:
no need to have the merge, just create a PR against the master.

ederag:
Strange. If you won't merge `revert-760-exitcode`, then why revert the PR in the first place ?

gaborbernat:
Was a miss click creation of that branch,  sorry if it caused confusion, the revert was never merged on master https://github.com/tox-dev/tox/commits/master, I consider the extra changes here an improvement over what we have here. Hence  why saying just  create a new PR 👍

ederag:
OK, then the simplest way seems to make a brand new fork and paste the modified files in it. Will file a new PR later.

tox/__init__.py
blueyed: For reference: subprocess actually returns negative exit codes for signals.
See https://github.com/tox-dev/tox/pull/1401 for a followup/fix.

PR #776, opened 2018-03-18
remove unused hookimpl marker

hpk42:
it's confusing to have the hookimpl marker in the hookspec module.


obestwalter:
This looks like a no brainer then :)

Issue #773, opened 2018-03-12
tox.cmdline() fails on 3.0.0rc2 but works in 2.9.1


Jerhaad:
We use this functionality in a lot of our tools and I'm not seeing anything about this in the changelog/documentation.

```bash
$ python
Python 3.6.3 (v3.6.3:2c5fed86e0, Oct  3 2017, 00:32:08)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import tox
>>> tox.cmdline()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: main() missing 1 required positional argument: 'args'
>>>
```

```bash
$ pip list
pip (9.0.1)
pluggy (0.6.0)
py (1.5.2)
setuptools (38.5.2)
six (1.11.0)
tox (3.0.0rc2)
virtualenv (15.1.0)
wheel (0.30.0)
```

Mac OS X 10.12.6

gaborbernat:
@Jerhaad thanks for the report, seems we should import ``run_main`` instead of ``main`` here https://github.com/tox-dev/tox/blob/master/tox/__init__.py#L87

obestwalter:
Thanks @Jerhaad - good, I start to believe in the power of the rc :)

@gaborbernat - you assigned yourself, so I'll check when this is fixed and release the hopefully last rc. I would also suggest to not merge anything else then anymore until 3.0 is released.

Issue #728, opened 2018-01-05
Preparing 3.0 release


obestwalter:
Looking at what is going on here, I think it is high time to get a new release out :)

It would be great if we could coordinate here what needs to be done, before we can cut the next release.

The one thing that is on my radar is the question of the logo and the docs. There were several people that think that the middle part of the logo looks a bit too much like the male gender sign. Maybe we can  find a solution for that, before the next release?

What else needs to be done before we can make a first rc?

I'll read my way into what happened over the last months and update https://github.com/tox-dev/tox/projects/7

vlcinsky:
Regarding the logo resembling male gender sign: I propose to call it [the perpen-dick-dick-ularity problem](https://youtu.be/BKorP55Aqvg?t=274) and "just ignore it", because:
- any logo will suffer from similar kind of problems
- in this case the geometry is not really an issue

Looking forward to all really important features in new tox release.

With best regards from my command line.

sigmavirus24:
> any logo will suffer from similar kind of problems

That's horribly dismissive and definitely false. Any good graphic designer could avoid that while still making an awesome logo.

obestwalter:
I am not a fan of the current state of the logo msyelf, but please let's keep the discussion about this here: https://github.com/tox-dev/tox/issues/639.

I definitely don't see the decision about a logo as a blocker for the next release though.

vlcinsky:
@sigmavirus24 sorry, I did not intend to be dismissive. My apologize. I am just looking forward to 2.10 release regardless of the logo.

fschulze:
I think #734 and #426 should go into the release.

obestwalter:
Life got in the way pretty hard, but I have some time (and mindspace) today and tomorrow, which I hope will be enough to get the rc out.

Next thing would be to remove me as a bottleneck for releases that I seem to have turned into now, as @hpk42 is busy doing other [cool stuff](https://github.com/hpk42/muacrypt) and I effectively have made the release process more complex in my attempts tzo simplify it (D'OH), so I will use the little time I have over the next weeks to try to fix that.

obestwalter:
I had some time to look through the issues and merged PRs. A lot has happened and it's really nice to see more contributors and the flurry of activity in the last months.

I will need some more time to wade through all of this and I also need to solve some interesting new problems I get, when trying to test and run things locally.

[...]

PR #777, opened 2018-03-23
#773 allow cmdline to be called bot with and without args, make args use sys.args if not specified

gaborbernat:
Take two on fixing this :face_with_head_bandage:


gaborbernat:
to fix report from https://github.com/tox-dev/tox/issues/728

PR #778, opened 2018-03-24
reintroduce hookimpl (revert 06e4877ab672f2ab264677ef53505f09fa0fec10)

hpk42:
it is used after all ...

### DIFF
--- a/tox/hookspecs.py
+++ b/tox/hookspecs.py
@@ -1,9 +1,11 @@
 """ Hook specifications for tox.
 
 """
+from pluggy import HookimplMarker
 from pluggy import HookspecMarker
 
 hookspec = HookspecMarker("tox")
+hookimpl = HookimplMarker("tox")
 
 
 @hookspec


obestwalter:
oh right. I just merged it to master yesterday while doing the last rc without really thinking about it. Having the release branch now means it doesn't affect ongoing releases. Turns out it being a no brainer means that I didn't use my brain, before merging it :D 

Thinking about it, it makes sense that tox implements some hooks itself. This will actually get more in the future, I guess. What is confusing though and what also might have thrown you, is the place where it is instantiated? Because that's the file where the specs live - not the implementations. 

I also had an actual look at the code and we are creating two identical `hookimpl` objects. One in `config.py` and one in `hookspecs.py`. I have to admit I never yet looked properly how the hook system works. I only know it exists and have a vague idea how to use it in plugins, so there might be some etra magic I don't get, but to my mind there should be only one hookimpl object per project. So if I am correct in that assumpton, it would be less confusing if it is instantiated once directly in `__init__.py` (where it is imported and added to `__all__` already).  I would then also add a comment there along the lines of: "tox implements some of its own hooks that can be overriden or enhanced upon by plugins".

Anyway just some thoughts. I'll merge the revert for now.

Issue #779, opened 2018-03-24
Migrate contributor docs into actual docs


obestwalter:
There is a lot of valuable contributor documentation in https://github.com/tox-dev/tox/blob/master/CONTRIBUTING.rst - I feel most of this should rather live in the actual docs and only be references in that file to improve visibility.

obestwalter:
Also as @The-Compiler pointed out, GH added a feature recently to set label descriptions, so those should go there:

Team members can review code and merge PRs. They can also label issues and change their status.

Issue labels
^^^^^^^^^^^^

Short overview about how labels should be used:

**bugs:**

* bug:minor -> does not affect many people or has no big impact
* bug:normal -> affects many people or has quite an impact
* bug:critical -> bad, bad thing (e.g. security or can cause data loss). Should be fixed as soon as possible
* bug:upstream -> something does not behave as it should, but can't or shouldn't be fixed in tox, but in a project tox depends on (e.g. pluggy, virtualenv, pip). Will be left open until the upstream bug is fixed for better visibility. Ideally we have an xfailing reproducer so we noticed automatically noticed if the bug is fixed (e.g. see https://github.com/tox-dev/tox/pull/553)

**features:**

* feature:new -> something does not exist yet, but should
* feature:change -> something exists already but should behave differently
* feature:deprecate -> something already exists, but shouldn't really :)

**types:**

* type:internal -> should have no impact on the user (refactoring, infrastructure, tools, etc.)
* type:organization -> has to do with the organization of the project (process)
* type:invalid -> not applicable - e.g. a misunderstanging about how tox works or a "bug" that is actually a feature. This could cleared up right in the issue.
* type:question -> not a bug or feature request, more of a question about how things work or if something is a bug or a feature

**levels:**

easy, medium and hard

A completely from the hip gut feeling estimation of how hard something is to fix or implement, basing on size of necessary changes and necessary knowledge of the tox code base and its surrounding ecosystem.

gaborbernat:
This has been done with tox 4.

Issue #555, opened 2017-07-15
Improving project organization


obestwalter:
Feedback highly apreciated. If there are no objections, I will phase these changes in over time, when wading through the issues.

## Use the project boards more

[Project boards](https://help.github.com/articles/about-project-boards/) are the virtual form of the "post-its on a whiteboard" approach to organization. If you know Trello or have a vague idea what a Kanban board is, you basically know all you need to start using them.

Issues can be pulled onto the project boards and moved around. There is also the possibility to add additional notes to the boards to give further guidance.

The issues move on so called swimlanes from the left to the right representing the phase they are in. The vertical position on the swimlane can be used for prioritization. The swimlanes can be freely configured per project and usually follow some form of todo/doing/done scheme.

Projects can be used for these purposes in the tox project:

* Group issues which belong to the same topic, give some order in which they might be tackled and visualize their status.

* Provide a different view into the issues to help getting an overview in certain areas.

### Project examples:

* [improve interpreter discovery](https://github.com/tox-dev/tox/projects/3) there is also a project that
* [group all bugs](https://github.com/tox-dev/tox/projects/1) and one to [plan the next release](https://github.com/tox-dev/tox/projects/6).
* there are other areas that have a lot of issues but somehow belong together like configuration handling, caching dependency problems, which could make it easier to get an overview what is going on, if they were grouped into project baords

### Stages

**Reference (optional)**

Notes and issues that are not actionable but provide more information about the project

**Open**

The issue belongs to the project but its status has not been determined yet. There might be open questions or the implementation needs to be discussed or it is not clear yet, if this should be worked upon at all.

**Ready**

This could be picked up

**In Progress**

Someone is working on this

**Done**

Issue is resolved but not yet part of a release

**Released/Wontfix**
Issue is part of a release or is not going to be resolved (could be reconsidered in the future though). Having an extra lane for wontfix issues, seeems a bit like overkill, so they will carry an extra label (wontfix) instead to mark them appropriately.

## Use better labels

I tried to do with the existing labels but I find them lacking. If we use projects to track issues we do not need prio and status labels because that will be presented by the position on the project board (which can also be seen in the issue itself). What IMO is needed though is to categorize them a bit more finegrained and make it easier to search for certain qualities of an issue.

I would propose the following groups of labels (to completely replace the existing ones):

bugs:

* bug:minor -> does not affect many people or has no big impact
* bug:normal -> affects many people or has quite an impact
[...]

PR #780, opened 2018-03-24
Add label info from #555 to contributor docs.

obestwalter:
Salvaged the explanation of label meanings and added it to the contributor docs. I think a lot of that should actually live in the actual docs, but for that I opened #779.


The-Compiler:
Now that GitHub [supports label descriptions](https://blog.github.com/2018-02-22-label-improvements-emoji-descriptions-and-more/), why not put them there (where people would probably expect them) instead of a document separate from the labels?

obestwalter:
He< @The-Compiler - thanks! Didn't know about that - will do :)

Closing this then.

nicoddemus:
That's an excellent idea, we should do that for pytest as well.

Em sáb, 24 de mar de 2018 11:49, Oliver Bestwalter <notifications@github.com>
escreveu:

PR #813, opened 2018-05-12
add automatic code format enforcer - black

gaborbernat:
``Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.`` - https://github.com/ambv/black

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,10 @@
 repos:
+-   repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+    - id: black
+      args: [--line-length=99, --safe]
+      python_version: python3.6
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v1.2.3
     hooks:
@@ -7,12 +13,6 @@ repos:
     -   id: check-yaml
     -   id: debug-statements
     -   id: flake8
--   repo: https://github.com/ambv/black
-    rev: stable
-    hooks:
-    - id: black
-      args: [--line-length=99, --safe]
-      python_version: python3.6
 -   repo: https://github.com/asottile/pyupgrade
     rev: v1.2.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,8 +8,6 @@ repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v1.2.3
     hooks:
-    -   id: trailing-whitespace
-    -   id: end-of-file-fixer
     -   id: check-yaml
     -   id: debug-statements
     -   id: flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: stable
+    rev: 18.4a4
     hooks:
     - id: black
       args: [--line-length=99, --safe]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,19 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    sha: v1.1.1
+    rev: v1.2.3
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-yaml
     -   id: debug-statements
     -   id: flake8
+-   repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+    - id: black
+      args: [--line-length=99, --safe]
+      python_version: python3.6
 -   repo: https://github.com/asottile/pyupgrade
-    sha: v1.2.0
+    rev: v1.2.0
     hooks:
     -   id: pyupgrade
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,9 @@
 .. image:: https://readthedocs.org/projects/tox/badge/?version=latest
   :target: http://tox.readthedocs.io/en/latest/?badge=latest
   :alt: Documentation status
+.. image:: https://img.shields.io/badge/code%20style-black-000000.svg
+  :target: https://github.com/ambv/black
+  :alt: Code style: black
 
 tox automation project
 ======================
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -8,16 +8,9 @@ import pytest
 from pluggy import PluginManager
 
 import tox
-from tox.config import (
-    CommandParser,
-    DepOption,
-    SectionReader,
-    get_homedir,
-    get_version_info,
-    getcontextname,
-    is_section_substitution,
-    parseconfig,
-)
+from tox.config import CommandParser, DepOption, SectionReader, get_homedir
+from tox.config import get_version_info, getcontextname, is_section_substitution, parseconfig
+
 from tox.venv import VirtualEnv
 
 
@@ -41,7 +34,7 @@ class TestVenvConfig:
             [],
             """
             [tox]
-            toxworkdir = %s
+            toxworkdir = {}
             indexserver =
                 xyz = xyz_repo
             [testenv:py1]
@@ -50,8 +43,9 @@ class TestVenvConfig:
             deps=
                 world1
                 :xyz:http://hello/world
-        """
-            % (tmpdir,),
+        """.format(
+                tmpdir
+            ),
         )
         assert config.toxworkdir == tmpdir
         assert len(config.envconfigs) == 2
@@ -252,9 +246,10 @@ class TestConfigPackage:
         config = newconfig(
             """
             [tox]
-            toxworkdir=%s
-        """
-            % tmpdir
+            toxworkdir={}
+        """.format(
+                tmpdir
+            )
         )
         assert config.toxworkdir == tmpdir
 
@@ -1009,9 +1004,10 @@ class TestConfigTestEnv:
         config = newconfig(
             """
             [testenv]
-            basepython=%s
-        """
-            % bp
+            basepython={}
+        """.format(
+                bp
+            )
         )
         assert len(config.envconfigs) == 1
         envconfig = config.envconfigs["python"]
@@ -1196,7 +1192,11 @@ class TestConfigTestEnv:
         assert (
             envconfig.install_command
             == [
-                "some_install", "--arg=%s/foo" % config.toxinidir, "python", "{opts}", "{packages}"
+                "some_install",
+                "--arg={}/foo".format(config.toxinidir),
+                "python",
+                "{opts}",
+                "{packages}",
             ]
         )
 
@@ -1655,7 +1655,7 @@ class TestConfigTestEnv:
         conf = newconfig([], inisource)
         configs = conf.envconfigs
         for name, config in configs.items():
-            assert config.basepython == "python%s.%s" % (name[2], name[3])
+            assert config.basepython == "python{}.{}".format(name[2], name[3])
 
     @pytest.mark.issue188
     def test_factors_in_boolean(self, newconfig):
@@ -1829,7 +1829,7 @@ class TestGlobalOptions:
 
     def test_correct_basepython_chosen_from_default_factors(self, newconfig):
         envlist = list(tox.PYTHON.DEFAULT_FACTORS.keys())
-        config = newconfig([], "[tox]\nenvlist=%s" % ", ".join(envlist))
+        config = newconfig([], "[tox]\nenvlist={}".format(", ".join(envlist)))
         assert config.envlist == envlist
         for name in config.envlist:
             basepython = config.envconfigs[name].basepython
@@ -2253,7 +2253,7 @@ class TestIndexServer:
                 pypi    = http://pypi.org/simple
         """
         config = newconfig([], inisource)
-        expected = "file://%s/.pip/downloads/simple" % config.homedir
+        expected = "file://{}/.pip/downloads/simple".format(config.homedir)
         assert config.indexserver["default"].url == expected
         assert config.indexserver["local1"].url == config.indexserver["default"].url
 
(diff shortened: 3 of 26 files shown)


obestwalter:
Hi @gaborbernat, are you using this project already somewhere else and can share some experiences with it? A first glimpse over the diff, looks like it is doing a good job. I also think I could get used to that style. Is it stable?

obestwalter:
I can't say I like the style it enforces, especially how parentheses are always on their own lines, but I get the argument that it produces the smallest possible difffs and most of that is a matter of getting used to anyway.

http://black.readthedocs.io/en/stable/installation_and_usage.html#note-this-is-an-early-pre-release

It's marked alpha.

> Also, as a temporary safety measure, Black will check that the reformatted code still produces a valid AST that is equivalent to the original. This slows it down. If you’re feeling confident, use --fast.

... but that check puts at my worries to rest.

 I will start using it in my own code that I am working with for a while.

gaborbernat:
It's not perfect, but I think it still provides much value.

obestwalter:
I just read the https://github.com/ambv/black#the-black-code-style and I have to say that I am convinced.

I also agree with the thoughts about line length:

> You probably noticed the peculiar default line length. Black defaults to 88 characters per line, which happens to be 10% over 80. This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, 90-ish seems like the wise choice.

So I would also propose to keep the black default for that.

obestwalter:
Not a bug - missing (tricky) feature (string wrapping): https://github.com/ambv/black/issues/182

gaborbernat:
Eh I like what we have with 100 now myself. As for the later, maybe down the line it will be fixed and then the formatter will fix that too automatically.

obestwalter:
Irt does some really weird things with strings though ...

It turns

    fmt = (
        "%(asctime)s %(filename)s.%(funcName)s:%(lineno)d %(levelname)s: "
        "%(message)s"
    )

into 

    fmt = (
        "%(asctime)s %(filename)s.%(funcName)s:%(lineno)d %(levelname)s: " "%(message)s"
    )

It does not concatenate the strings again ...

gaborbernat:
Hmmm may require some manual one time mangling then. Probably fill in an issue for it though.

obestwalter:
See https://github.com/ambv/black/issues/205

.pre-commit-config.yaml
```diff
@@ -1,13 +1,19 @@
...
-    sha: v1.1.1
+    rev: v1.2.3
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-yaml
     -   id: debug-statements
     -   id: flake8
+-   repo: https://github.com/ambv/black
+    rev: stable
```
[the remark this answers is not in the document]
gaborbernat: fixed, and yeah I agree about the quotes; but at the end of day I prefer consistency over which of the two :-)

gaborbernat:
@obestwalter take another look cleaned up manually few of the oddities, if there are any more you spot can do similarly

asottile:
Another option would be to use autopep8 + double-quote-string-fixer (converts strings in the opposite directing of black) + a few other one off formatters to get basically the same deal :man_shrugging:

gaborbernat:
I'm personally not that pedantic about the single quote to consider worth the effort there.

gaborbernat:
@obestwalter it now passed tests, let me know if you spot anything else :smile:

obestwalter:
> I wish it was single quotes instead :/

Hey @asottile, if you are interested how this came to pass: I just stumbled over the issue dealing with this: https://github.com/ambv/black/issues/51 - there are a few very strong arguments why double quotes are the better choice. The TL;DR of the tipping point is: apostrophes (https://github.com/ambv/black/pull/75#issuecomment-376193551).

Regarding other tools: the more I look into black the more I get the feeling that this could become a standard that will save the community a lot of bike shedding.

I really like the idea to approach the problem from the angle that code should be formatted in a way that maintainability and preotection from stupid bugs is prioritized. For that goal it is doing stuff that "traditional" formatters won't touch and IMO that is a GoodThing. Prominently protecting us from the infamous: "Whoops there should actually have been a comma between those two strings in a list, where I moved an element from the end to the middle". 

@ambv also has a very clear vision, but is open to change his opinion if confronted with strong arguments (e.g. he also preferred single quotes first), so this looks like a very good project to go with and thanks to the AST checking safeguard, we do not have to be too worried that this first big reformatting PR will introduce some subtle bugs.

The only problems I see atm are around the topic of strings, but they seem to be not too painful and will likely get better. I noticed that our way of creating these mini configs with `initproj` gets totally shot, which IMO is another black bug, as it actually changes the formatting of the string that will be outputted. In our case it is only an annoyance as the change in formatting does not affect the parsed config. It just looks really ugly now, but they can be formatted in normal strings anyway (I don't need a triple quoted string to create a test config with one or two line breaks), so I would suggest to reformat simple configs like:

        config = newconfig(
            [],
            """
            [testenv:py1]
            deps=pytest
        """,
        )

to:

config = newconfig([], "[testenv:py1]\ndeps=pytest")

Which is much more readable than even the default formatting anyway.

I am going through the PR today and if there are no objections, I think we should whip this into shape together quickly as long as @gaborbernat has still his PyCon wings on :)

asottile:
> the more I look into black the more I get the feeling that this could
become a standard that will save the community a lot of bike shedding.

yapf started with exactly the same mentality, and is much more mature.

> For that goal it is doing stuff that "traditional" formatters won't touch

true yapf / autopep8 don't even try here but that isn't to say that all
formatters don't:

- https://github.com/asottile/add-trailing-comma
- https://github.com/asottile/pyupgrade

> The TL;DR of the tipping point is: apostrophes

For me it's less of a personal preference thing and more of an annoyance.
A good chunk of the python I write is for web development where double
quotes in strings are extremely common (html!).  The linked comment above
happens way more frequently in the _other direction_ for me.  It's also
contrary to nearly every style guide I've worked with (again, because web
development, but also because pressing the shift key is a slow and `"ERorr
prone"` operation).

> config = newconfig([], "[testenv:py1]\ndeps=pytest")

> Which is much more readable than even the default formatting anyway.

Having used this corner-cut myself before, I find it all too easy to typo
`\b` instead of `\n` whereas with multiline strings this isn't a problem.
Another note in your example is the two aren't equivalent -- there's a
whitespace difference (`textrwap.dedent` comes to mind).

Anyway, I'm not out here to change minds -- code formatters are some of my
favorite things and I think using more of them is a great step in the right
direction.  Just wish @ambv's black didn't strongly prefer (in my mind) the
less convenient quotation style.



On Sun, May 13, 2018 at 3:04 AM, Oliver Bestwalter <notifications@github.com

gaborbernat:
Given ``tox`` is definitely not a web tool, I can get along with having ``"``. As for the initproj we already dedent by default the code, so I think that's just fine.

RonnyPfannschmidt:
i just finished applying black to 2 of my projects along with pre-commit hooks

while my quoting preference is also different, i consider it a straw-man argument, you fix it once, and then keep it consistent -> and consistency is more important there imho

obestwalter:
I understand your annoyance @asottile - it is very dependent on in which kind of universe your code lives - maybe that should really be something configurable in black and maybe it will be at some point to get those kind of projects on board (line lenght is configurable already after all). 

The way how black deals with quotes is definitely quite sane IMO - independent of preferences. It always tries to reformat the string in a way that you need the least amount of escapes. e.g. It and leaves strings like `bla = 'hello "world"'` alone and reformats strings like `blu = "hello \"world\""` to blu = 'hello "world"'

As you seem o.k. with using black anyway as a step in the right direction and we have an implicit +1 from @ronnypfannschmidt also, I would say let's just run with this for now. 

I'll finish reviewing this tonight then and merge it, so that new PRs coming in tomorrow at the sprint can already base on this, o.k.?

tox/session.py
```diff
@@ -59,24 +59,29 @@ def show_help(config):
     tw.write(config._parser._format_help())
     tw.line()
     tw.line("Environment variables", bold=True)
-    tw.line("TOXENV: comma separated list of environments "
-            "(overridable by '-e')")
-    tw.line("TOX_TESTENV_PASSENV: space-separated list of extra "
-            "environment variables to be passed into test command "
-            "environments")
+    tw.line("TOXENV: comma separated list of environments " "(overridable by '-e')")
```
obestwalter: needs manual concat.

tox/venv.py
```diff
@@ -375,13 +377,21 @@ def test(self, redirect=False):
...
+                        ignore_ret=ignore_ret,
+                        testcommand=True,
+                    )
                 except tox.exception.InvocationError as err:
                     if self.envconfig.ignore_outcome:
                         self.session.report.warning(
-                            "command failed but result from testenv is ignored\n"
-                            "  cmd: %s" % (str(err),))
+                            "command failed but result from testenv is ignored\ncmd: {}".format(
+                                str(err)
```
obestwalter: loose the str() here.

.pre-commit-config.yaml
```diff
@@ -1,13 +1,19 @@
...
+    rev: 18.4a4
+    hooks:
+    - id: black
+      args: [--line-length=99, --safe]
+      python_version: python3.6
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    sha: v1.1.1
+    rev: v1.2.3
     hooks:
     -   id: trailing-whitespace
```
obestwalter: Not necessary with black anymore, I guess.

.pre-commit-config.yaml
```diff
@@ -1,13 +1,19 @@
...
     -   id: end-of-file-fixer
```
obestwalter: see above

README.rst
```diff
@@ -16,6 +16,9 @@
 .. image:: https://readthedocs.org/projects/tox/badge/?version=latest
   :target: http://tox.readthedocs.io/en/latest/?badge=latest
   :alt: Documentation status
+.. image:: https://img.shields.io/badge/code%20style-black-000000.svg
+  :target: https://github.com/ambv/black
+  :alt: Code style: black
```
obestwalter: Yeah, I guess that is the case now :)

tests/test_config.py
```diff
@@ -8,27 +8,33 @@
...
+
     def test_config_parsing_minimal(self, tmpdir, newconfig):
-        config = newconfig([], """
+        config = newconfig(
+            [],
+            """
             [testenv:py1]
-        """)
+        """,
+        )
```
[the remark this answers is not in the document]
gaborbernat: Yeah, the comma in this case is optional, but never changes the meaning.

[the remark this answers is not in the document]
obestwalter: I was rather thinking that the line break introduced after the `"""` would introduce changes in formatting (in some cases - not this one, so the comment was placed not in a good example), but they don't.

tests/test_pytest_plugins.py
```diff
@@ -10,89 +10,97 @@
...
+        "input, expected",
+        (
+            ("", []),
+            ("/", ["/"]),
+            ("//", ["//"]),
+            ("/a", ["/", "a"]),
+            ("/a/", ["/", "a"]),
+            ("/a/b", ["/", "a", "b"]),
+            ("a", ["a"]),
+            ("a/b", ["a", "b"]),
```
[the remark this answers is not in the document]
gaborbernat: indeed :laughing:

tests/test_result.py
```diff
@@ -65,31 +69,34 @@ def test_get_commandlog(pkg):
     assert "setup" not in envlog.dict
     setuplog = envlog.get_commandlog("setup")
     setuplog.add_command(["virtualenv", "..."], "venv created", 0)
-    assert setuplog.list == [{"command": ["virtualenv", "..."],
-                              "output": "venv created",
-                              "retcode": "0"}]
+    assert (
+        setuplog.list
+        == [{"command": ["virtualenv", "..."], "output": "venv created", "retcode": "0"}]
```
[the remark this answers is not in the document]
gaborbernat: removed all these by extracting the constant to a local variable

obestwalter:
@asottile said:
>> config = newconfig([], "[testenv:py1]\ndeps=pytest")
>> Which is much more readable than even the default formatting anyway.
> Having used this corner-cut myself before, I find it all too easy to typo
> `\b` instead of `\n` whereas with multiline strings this isn't a problem.
> Another note in your example is the two aren't equivalent -- there's a
> whitespace difference (`textrwap.dedent` comes to mind).

In general I agree, but I think in the context of a tox config test this is not a problem. If you accidentally type `\b` your test will fail and you'll fix the typo. The whitespace difference is also irrelevant in the context of those particular tests.

gaborbernat:
@obestwalter should be ready for merge now, :smile: as for doing the ``%`` to ``str.format`` I did it mostly by search, and pycharm suggestions

obestwalter:
> as for doing the % to str.format I did it mostly by search, and pycharm suggestions

Wow. You've been busy then :) thanks for all the work you put in to whip that into shape.

tests/test_config.py
```diff
@@ -8,9 +8,16 @@
...
+from tox.config import (
+    CommandParser,
+    DepOption,
+    SectionReader,
+    get_homedir,
+    get_version_info,
+    getcontextname,
+    is_section_substitution,
+    parseconfig,
+)
```
obestwalter: argh, now we're back to that thing, but tbh atm I don't care anymore. From now on it's just a question of reconfiguring whatever formatter we use.

tests/test_config.py
```diff
@@ -2659,28 +2647,17 @@ def test_command_with_split_line_in_subst_arguments(self):
...
-                " ",
-                "!deferred",
-                " ",
-                "--with-doctest",
-                " ",
-                "[]",
-            ]
-        )
+        expected = [
+            "nosetests", " ", "-v", " ", "-a", " ", "!deferred", " ", "--with-doctest", " ", "[]"
```
obestwalter: phew - thanks.

tests/test_config.py
gaborbernat: yeah, maybe things will change in future with https://github.com/ambv/black/issues/204

gaborbernat: stays as it is then :+1:

PR #817, opened 2018-05-14
enforce and automate import ordering

gaborbernat:
Removing the final source of possible style debates.

thanks @asottile for the help and for creating https://github.com/asottile/seed-isort-config

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ def main():
                 "pytest >= 3.0.0", "pytest-cov", "pytest-mock", "pytest-timeout", "pytest-xdist"
             ],
             "docs": ["sphinx >= 1.6.3, < 2", "towncrier >= 17.8.0"],
-            "lint": ["flake8 == 3.5.0", "flake8-bugbear == 18.2.0", "pre-commit == 1.8.2"],
+            "lint": ["pre-commit == 1.8.2"],
             "publish": ["devpi", "twine"],
         },
         classifiers=[
--- a/tox.ini
+++ b/tox.ini
@@ -33,7 +33,7 @@ passenv = {[testenv]passenv}
           # without PROGRAMDATA cloning using git for Windows will fail with an
           # `error setting certificate verify locations` error
           PROGRAMDATA
-extras = lint, testing, docs, publish
+extras = lint
 description = run static analysis and style check using flake8
 commands = pre-commit run --all-files --show-diff-on-failure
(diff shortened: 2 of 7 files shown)


tox.ini
```diff
@@ -33,11 +33,9 @@ passenv = {[testenv]passenv}
           # without PROGRAMDATA cloning using git for Windows will fail with an
           # `error setting certificate verify locations` error
           PROGRAMDATA
-extras = lint
+extras = lint, testing, docs, publish
```
asottile: I think you only need `lint` here

setup.py
```diff
@@ -55,7 +55,7 @@ def main():
                 "pytest >= 3.0.0", "pytest-cov", "pytest-mock", "pytest-timeout", "pytest-xdist"
             ],
             "docs": ["sphinx >= 1.6.3, < 2", "towncrier >= 17.8.0"],
-            "lint": ["flake8 == 3.5.0", "flake8-bugbear == 18.2.0", "pre-commit == 1.8.2"],
```
asottile: probably want to add `flake8-bugbear==...` to the `additional_dependencies` of the `id: flake8` hook so that we continue to lint for these things

asottile: also while doing that, probably need to limit `flake8` to `language_version: python3.6` since if I recall correctly `flake8-bugbear` only supports python3

tox.ini
gaborbernat: needed all these I believe to seed isort, testing to have pytest, docs to have sphinx, publish for stuff used in setup.py

asottile: `seed-isort-config` runs in an isolated environment -- it doesn't actually need any dependencies :D

gaborbernat: how does it then differentiate between local/thirdparty/stdlib?

setup.py
gaborbernat: yeah, ammended

tox.ini
asottile: `aspy.refactor_imports` is clever: https://github.com/asottile/aspy.refactor_imports/blob/9acde162c907d0e28b665dd906ac032052ba300c/aspy/refactor_imports/classify.py#L121-L142

tl;dr:

1. try to use the import machinery (it'll probably fail), if it succeeds classify based on whether it is from a `X-packages` directory => (system or third party)
2. look on the local filesystem for the first import segment => (first party)
3. otherwise => (third party)

gaborbernat: sounds great

PR #821, opened 2018-05-14
allow popen of action to take a shell argument

gaborbernat:
this allows plugin users to use our framework to handle shell calls :-)

### DIFF
--- a/tox/session.py
+++ b/tox/session.py
@@ -14,6 +14,7 @@ import sys
 import time
 
 import py
+import six
 
 import tox
 from tox._verlib import IrrationalVersionError, NormalizedVersion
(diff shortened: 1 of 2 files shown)


obestwalter:
Can you give a quick example in which situation this would be necessary? Generally I get a bit scared when seeing `shell=True` being used for anything, so there should be a good reason for adding this.

tox/session.py
```diff
@@ -133,7 +134,16 @@ def _initlogpath(self, actionid):
...
-    def popen(self, args, cwd=None, env=None, redirect=True, returnout=False, ignore_ret=False):
+    def popen(
+        self,
+        args,
+        cwd=None,
+        env=None,
+        redirect=True,
+        returnout=False,
+        ignore_ret=False,
+        shell=False,
```
asottile: I'm actually pretty `-1` on this.  shells are a great way to expose a CVE.  If someone wants a shell they can already `['bash', '-c', ...]`

gaborbernat:
conda has some machinery (conda activate) that requires running in shell mode; here we don't run things ourselves in shell, but allow plugin users to do so without re-writing themselves this whole logic

obestwalter:
And there no other way to accomplish this?

asottile:
is `conda activate` more than just setting some environment variables `PATH`, `LD_..`, etc.?

gaborbernat:
@asottile seems it's much more, it can run arbitrary user provided scripts

gaborbernat:
any better ideas? :-)

gaborbernat:
``['bash', '-c', ...]`` it is then :+1:

PR #822, opened 2018-05-14
prefer using with as for action

gaborbernat:
this is much more Pythonic


obestwalter:
And pythonic we like :) - will be merged automatically, when tests are done.

Issue #826, opened 2018-05-17
Switch to src/ layout?


obestwalter:
While we are in tidy up mode: How about switching the project to src/ layout (https://github.com/pypa/python-packaging-user-guide/issues/320 - especially: https://github.com/pypa/python-packaging-user-guide/issues/320#issuecomment-389938850) - we have that kind of layout.

Apart from avoiding potential gotchas, this would make it trivial to run the tests of a checkout against a diffferent version of tox that is installed.

gaborbernat:
Yeah definitely this needs to happen.

obestwalter:
Some additional remarks on this, even when I seem to have pushed at an open door with this proposal here. 

I think using this layout specifically in the context of projects using pytest and/or tox should be generally recommended. The reasoing behinfd this is laed out very well here in the [pytest docs](https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code). Now, if you actually **are** pytest or tox and therefore testing yourself with yourself (how wonderfully meta is that?)), this should be the obvious layout choice. 

Also: the vision of tox is to support and promote good testing and packaging practices, so the way how the project itself is structured should follow its own advice.

gaborbernat:
@obestwalter this is now fixed

obestwalter:
Thanks @gaborbernat!

PR #827, opened 2018-05-17
[#826] switch to source layout

gaborbernat:
a POC of it at least


tox.ini
```diff
@@ -115,4 +115,4 @@ include_trailing_comma = True
 force_grid_wrap = 0
 line_length = 99
 known_first_party = tox
-known_third_party = pkg_resources,pluggy,py,pytest,setuptools,six
```
gaborbernat: @asottile I think this is a false positive of the isort plugin

asottile: yeah, it assumes a specific layout.  PRs welcome, the missing bits are to set the application directories:

For example, here's how `reorder_python_imports` handles this:
https://github.com/asottile/reorder_python_imports/blob/794f0498340ab3adbef443431993f8a8b2046da7/reorder_python_imports.py#L405-L411

You'll want:

```yaml
    args: [--application-directories, src]
```

once `seed-isort-config` supports that

gaborbernat: I'm caught up now with other stuff, so if you can fix it that's cool :+1:

asottile:
Here ya go:

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 8a1d348..3469cf6 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,9 +6,10 @@ repos:
       args: [--line-length=99, --safe]
       python_version: python3.6
 -   repo: https://github.com/asottile/seed-isort-config
-    rev: v0.1.0
+    rev: v1.0.0
     hooks:
     -   id: seed-isort-config
+        args: [--application-directories, src]
 -   repo: https://github.com/pre-commit/mirrors-isort
     rev: v4.3.4
     hooks:
diff --git a/tox.ini b/tox.ini
index 548f602..9acd122 100644
--- a/tox.ini
+++ b/tox.ini
@@ -115,4 +115,4 @@ include_trailing_comma = True
 force_grid_wrap = 0
 line_length = 99
 known_first_party = tox
-known_third_party = pkg_resources,pluggy,py,pytest,setuptools,six,tox
+known_third_party = pkg_resources,pluggy,py,pytest,setuptools,six
```

tox.ini
```diff
@@ -103,7 +103,7 @@ source = tox
          {toxworkdir}/pypy*/site-packages/tox
 
 [pytest]
-addopts = -rsxX -vvv --showlocals
+addopts = -rsxX --showlocals
```
obestwalter: ok :)

setup.py
```diff
@@ -1,7 +1,8 @@
 import io
 import sys
 
-import setuptools
+from setuptools import __version__ as setuptools_version
```
obestwalter: What is the connection to switching to src/ layout here? Is that some name shadowing thing?

nicoddemus:
Rá, pytest wins the race! https://github.com/pytest-dev/pytest/pull/3513 😸

gaborbernat:
@nicoddemus the reason that's the case is cause I did not figure out yet how to merge coverage files in between interpreters, will probably do it tomorrow when I have some time for that. As I see pytest does not do that 🤔

anthrotype:
you need to run coverage with `--parallel-mode` option and use `[paths] source = ...` in `.coveragerc to list the file paths that should be considered equivalent when combining data:

https://coverage.readthedocs.io/en/coverage-4.5.1/config.html#paths

E.g.:
https://github.com/pyca/cryptography/blob/afdbfb13780fb78e7b277b9de07e7636ba9c5119/tox.ini#L18
https://github.com/pyca/cryptography/blob/afdbfb13780fb78e7b277b9de07e7636ba9c5119/.coveragerc#L7-L11

anthrotype:
@gaborbernat I see. I remember I also tried once to do it with pytest-cov but then gave up and decided to run pytest as `coverage run -m pytest`.

gaborbernat:
@nicoddemus managed to nail down things now; tox is not too far behind! @obestwalter @asottile please review and we can merge :+1:

tox.ini
```diff
@@ -12,47 +12,60 @@ skip_missing_interpreters = true
 
 [testenv]
 description = run the tests with pytest under {basepython}
-setenv = COVERAGE_FILE={toxworkdir}/.coverage.{envname}
+setenv = VIRTUALENV_DOWNLOAD = 0
```
asottile: I've usually seen this written as `VIRTUALENV_NO_DOWNLOAD=1`

tox.ini
```diff
@@ -12,47 +12,60 @@ skip_missing_interpreters = true
...
+         PIP_DISABLE_VERSION_CHECK = 1
+         COVERAGE_FILE = {toxworkdir}/.coverage.{envname}
 passenv = http_proxy https_proxy no_proxy SSL_CERT_FILE TOXENV CI TRAVIS TRAVIS_* APPVEYOR APPVEYOR_* CODECOV_*
+deps = setuptools >= 29.0.0
+       setuptools_scm >= 2.0.0, < 3
+       wheel >= 0.31.0
```
asottile: you set `--no-download,`, but then ask for an upgrade of `wheel` / `setuptools`? (why not just allow virtualenv to do that for you at this point?)

tox.ini
```diff
@@ -12,47 +12,60 @@ skip_missing_interpreters = true
...
 description = invoke sphinx-build to build the HTML docs and check that all links are valid
 whitelist_externals = sphinx-build
 basepython = python3.6
 extras = docs
+changedir = {toxinidir}
 commands = sphinx-build -d "{toxworkdir}/docs_doctree" doc "{toxworkdir}/docs_out" --color -W -bhtml
 
 
 [testenv:fix-lint]
+description = run static analysis and style check using flake8
```
asottile: heh, description isn't exactly right

tox.ini
```diff
@@ -12,47 +12,60 @@ skip_missing_interpreters = true
...
+changedir = {toxinidir}
 commands = pre-commit run --all-files --show-diff-on-failure
            python -c 'print("hint: run {envdir}/bin/pre-commit install to add checks as pre-commit hook")'
 
 
 [testenv:coverage]
-description = combine coverage data and create reports
-deps = coverage
+description = [run locally after tests]: combine coverage data and create report;
+              generates a diff coverage against origin/master (can be changed by setting DIFF_AGAINST env var)
```
asottile: should `DIFF_AGAINST` be in `passenv`?

tox.ini
```diff
@@ -12,47 +12,60 @@ skip_missing_interpreters = true
...
+description = run static analysis and style check using flake8
 basepython = python3.6
 passenv = {[testenv]passenv}
           HOMEPATH
           # without PROGRAMDATA cloning using git for Windows will fail with an
           # `error setting certificate verify locations` error
           PROGRAMDATA
 extras = lint
-description = run static analysis and style check using flake8
+changedir = {toxinidir}
```
asottile: heh, `pre-comit` will `cd` right back to the repository root, not sure this is useful

tox.ini
```diff
@@ -90,21 +106,21 @@ max-line-length = 99
 ignore = E203, W503
 
 [coverage:run]
-omit = tox/__main__.py
+branch = false
```
asottile: branch coverage is useful! please don't turn it off!

tox.ini
```diff
@@ -90,21 +106,21 @@ max-line-length = 99
...
 
 [coverage:paths]
-source = tox
-         {toxworkdir}/*/lib/python*/site-packages/tox
-         {toxworkdir}/*/Lib/site-packages/tox
-         {toxworkdir}/pypy*/site-packages/tox
+source = src/tox
+         .tox/*/lib/python*/site-packages/tox
+         .tox/*/Lib/site-packages/tox
+         .tox/pypy*/site-packages/tox
```
asottile: this is by far my least favorite part of moving to a `src` directory, the coverage reports become unreadable :(

we've used [this script](https://github.com/Yelp/venv-update/blob/master/tests/testing/fix_coverage.py) or a derivative of it to try and make this more sane (it combines the coverage data from site-packages over top the local directory)

tox.ini
gaborbernat: It has to do with the index URL used. By default virtualenv uses PyPi. This way the tox config adheres too whatever the user has configured; e.g. value from ``PIP_INDEX_URL``.

tox.ini
```diff
@@ -12,47 +12,58 @@ skip_missing_interpreters = true
...
 
 [testenv:fix-lint]
+description = format the code base to adhere to our styles, and complain about what we cannot do automatically
 basepython = python3.6
 passenv = {[testenv]passenv}
           HOMEPATH
           # without PROGRAMDATA cloning using git for Windows will fail with an
           # `error setting certificate verify locations` error
           PROGRAMDATA
+          DIFF_AGAINST
```
asottile: oops, added this to the wrong testenv, needs to be in the `coverage` testenv

gaborbernat:
things now looks great :+1: so I'll go ahead and merge it. :speak_no_evil: sorry for people with PRs for the merge conflicts

PR #832, opened 2018-05-26
[wip] experimenting with type hint generation (pyannotate)

obestwalter:
I did not do much with this yet, just added the recommended boilerplate for pytest and ran the tests. On first view it only generated hints for the tests itself (where we definitely not would want them, I guess), but not for the production code.

Also see http://mypy-lang.blogspot.de/2017/11/dropbox-releases-pyannotate-auto.html

related issue: #831

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,12 @@ def main():
         install_requires=["py>=1.4.17", "pluggy>=0.3.0,<1.0", "six", "virtualenv>=1.11.2"],
         extras_require={
             "testing": [
-                "pytest >= 3.0.0", "pytest-cov", "pytest-mock", "pytest-timeout", "pytest-xdist"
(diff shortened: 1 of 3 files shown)


gaborbernat:
I have a branch with stubs where I have used mypy stubgen and hand
corrected it, so this seems a bit less reliable and duplication of work I
think.

gaborbernat:
See https://github.com/gaborbernat/tox/tree/type-hint-stubs/tox for details :+1: requires https://github.com/python/mypy/issues/5028 before creating a PR though.

obestwalter:
I did not put much work in yet, I just played with it for a bit and thought I push that boilerplate up, if anyone wants to experiment. looking forward to see what's coming out of this then.

gaborbernat:
This now is getting out of date, so I'll close.

PR #834, opened 2018-06-03
tox is always represented as lower case

asottile:
Committed via https://github.com/asottile/all-repos


asottile:
context: https://twitter.com/obestwalter/status/1003068449315278849

PR #835, opened 2018-06-03
blacken-docs

asottile:
Hi friends, I made a terrific (terrible? terrifying?) thing.

I needed one manual `,` addition to the `setup.py` example so it could be parsed -- otherwise this is fully automated!

Using the new thing I made today: https://github.com/asottile/blacken-docs

Heavily inspired by pytest-dev/pytest#3525.  See also ambv/black#292, ambv/black#294.


obestwalter:
Very cool - automate all the things :)

Issue #477, opened 2017-03-12
basepython + shared envdir + default environments = confusion


obestwalter:
This came up in #425:

`tox.ini`

    [tox]
    envlist = py36,py27
    [testenv]
    basepython=python3.6
    envdir = {toxworkdir}/sharedenvdir
    commands = python --version

Leads to the py27 environment also be run with python3.6 which is very likely not intended by the user and a mistake in the configuration. 

This might lead to unnecessary confusion and I wonder if we should catch that and throw an appropriate error, when we encounter configurations like that?

asottile:
A bug +1 on this one, even without `envdir` this is pretty confusing and a problem I've seen a lot

```ini
[tox]
envlist = py27,py36
[testenv]
basepython=python3.6
commands = python --version
```

Surprise! no longer testing against python2.7

I wrote a tool at my last employment which scanned all of the repositories and checked for `pyXX` environments that had a mismatched interpreter version, perhaps something of the sort could be baked into tox itself? (if it sees for instance a `py27` environment but the interpreter endswith `python3.6` it could error?).  Unfortunately the tox-specific tool I wrote is lost to closed source :( (though I've [rewritten the multi-repository scanner bit](https://github.com/asottile/all-repos))

alvations:
We're having the same problem. Somehow we can find the Python version on the terminal but `tox` can't https://stackoverflow.com/questions/48274339/where-did-tox-look-when-searching-for-basepython. And we've no idea where exactly is `tox` searching. 

Is it possible to pass in information to tell `tox` to specifically look for the Python binaries at a specific location?

gaborbernat:
@alvations tox uses the PATH environment variable, so try to set it to where to look for :+1:

asottile:
@alvations seems this is probably too complicated to solve over stack overflow comment thread :)  mind shooting me an email and we can try and work out iteratively to debug this (or if I can get ssh access -- I can probably find the bug directly).  My initial suspect is the `eval` + `inspect`:
https://github.com/tox-dev/tox/blob/336f4f6bd8b53223f940fc5cfc43b1bbd78d4699/tox/interpreters.py#L61
https://github.com/tox-dev/tox/blob/336f4f6bd8b53223f940fc5cfc43b1bbd78d4699/tox/interpreters.py#L81

(we can most likely replace this with `json` since py25 is long since dead -- but I want to confirm that's the problem)

If this seems ok, shoot me an email: asottile@umich.edu

[...]

PR #841, opened 2018-06-08
Ignore basepython for default factors (#477)

stephenfin:
tox provides a number of default factors - py27, py34, py35 etc. - that
are tied to particular interpreter versions. It is possible to override
these through individual sections or the global [testenv] section. For
example, consider the following 'tox.ini' file:

	[tox]
	skipsdist = True
	minversion = 2.0
	distribute = False
	envlist = py35,py27,pep8,py34-test

	[testenv]
	basepython = python3
	install_command = pip install {opts} {packages}
	commands =
	python --version

	[testenv:py27]
	basepython = python2.7

Running any target except for 'py27' will result in the same interpreter
being used. On Fedora 28 with the 'python3-tox' package:

	$ tox -qq -e py27
	Python 2.7.15
	$ tox -qq -e py35
	Python 3.6.5
	$ tox -qq -e py34-test
	Python 3.6.5

This is broken by design. Overriding these makes no sense and is a
source of common misconfigurations, as noted in #477. The only sane
thing to do here is ignore the request and use the correct interpreter
so this is what's done.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -47,7 +47,7 @@ intersphinx_mapping = {"https://docs.python.org/": None}
 def setup(app):
     # from sphinx.ext.autodoc import cut_lines
     # app.connect('autodoc-process-docstring', cut_lines(4, what=['module']))
-    app.add_description_unit(
+    app.add_object_type(
         "confval",
         "confval",
         objname="configuration value",
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -58,6 +58,18 @@ and will first lookup global tox settings in this section:
     * environment variable ``TOXENV``
     * ``tox.ini`` file's ``envlist``
 
+.. confval:: ignore_basepython_conflict=True|False(default)
+
+    .. versionadded:: 3.1.0
+
+    If ``True``, :confval:`basepython` settings that conflict with the Python
+    variant for a environments using default factors, such as ``py27`` or
+    ``py35``, will be ignored. This allows you to configure
+    :confval:`basepython` in the global testenv without affecting these
+    factors. If ``False``, the default, a warning will be emitted if a conflict
+    is identified. In a future version of tox, this warning will become an
+    error.
+
 
 Virtualenv test environment settings
 ------------------------------------
@@ -81,8 +93,15 @@ Complete list of settings that you can put into ``testenv*`` sections:
 
 .. confval:: basepython=NAME-OR-PATH
 
-    name or path to a Python interpreter which will be used for creating
-    the virtual environment. **default**: interpreter used for tox invocation.
+    Name or path to a Python interpreter which will be used for creating
+    the virtual environment; if the environment name contains a :ref:`default
+    factor <factors>`, this value will be ignored. **default**: interpreter
+    used for tox invocation.
+
+    .. versionchanged:: 3.1
+
+       Environments that use a :ref:`default factor <factors>` now ignore this
+       value, defaulting to the interpreter defined for that factor.
 
 .. confval:: commands=ARGVLIST
 
@@ -546,9 +565,6 @@ However, a better approach looks like this:
     envlist = {py27,py36}-django{15,16}
 
     [testenv]
-    basepython =
-        py27: python2.7
-        py36: python3.6
     deps =
         pytest
         django15: Django>=1.5,<1.6
@@ -612,23 +628,12 @@ Factors and factor-conditional settings
 ++++++++++++++++++++++++++++++++++++++++
 
 Parts of an environment name delimited by hyphens are called factors and can
-be used to set values conditionally:
+be used to set values conditionally. In list settings such as ``deps`` or
+``commands`` you can freely intermix optional lines with unconditional ones:
 
 .. code-block:: ini
 
-    basepython =
-        py27: python2.7
-        py36: python3.6
-
-This conditional setting will lead to either ``python3.6`` or
-``python2.7`` used as base python, e.g. ``python3.6`` is selected if current
-environment contains ``py36`` factor.
-
-In list settings such as ``deps`` or ``commands`` you can freely intermix
-optional lines with unconditional ones:
-
-.. code-block:: ini
-
+    [testenv]
     deps =
         pytest
         django15: Django>=1.5,<1.6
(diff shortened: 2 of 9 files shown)


gaborbernat:
@stephenfin thank you very much for the configuration. I would prefer to raise an error for invalid configuration though; instead of ignoring it. @obestwalter agree?

stephenfin:
> @stephenfin thank you very much for the configuration. I would prefer to raise an error for invalid configuration though; instead of ignoring it. @obestwalter agree?

I considered that but decided not to because I wanted to allow configurations like the following:

```
[tox]
envlist = py27,py35,py36,docs,style,coverage

[testenv]
basepython = python3
commands = pytest ...

[testenv:docs]
commands = sphinx-build ...

[testenv:coverage]
commands = coverage ...

[testenv:style]
commands = flake8 ...
```

e.g. I would be able to ensure everything *except* the Python-versioned environments would use Python 3, to assist in migrations to Python 3. This has come up repeatedly in OpenStack, as seen [here](https://review.openstack.org/#/c/571207/1/tox.ini) and [here](https://review.openstack.org/#/q/topic:python3-first), and simply overriding things allows us to keep our `tox.ini` smaller. I would have added a DEBUG-level log, but it seems you're not using any logging infrastructure so I can't.

Your call.

gaborbernat:
@stephenfin I see your point; I say let's follow the zen of Python: explicit better than implicit. So instead of blindly ignoring things let's have a way to opt-in into such decisions. 

```
basepython_env_name_missmatch = raise | ignore | override
```

Where raise aborts with a config error; ignore has the current strategy, and override can be what you're proposing here. Let's leave ``ignore`` as default for now (we can change it when we do a major release), and we should also expose this option as a command line argument to facilitate with testing/debugging. Does that seem reasonable to you?

stephenfin:
> @stephenfin I see your point; I say let's follow the zen of Python: explicit better than implicit. So instead of blindly ignoring things let's have a way to opt-in into such decisions.
> 
>     basepython_env_name_missmatch = raise | ignore | override
> 
> Where raise aborts with a config error; ignore has the current strategy, and override can be what you're proposing here. Let's leave ignore as default for now (we can change it when we do a major release), and we should also expose this option as a command line argument to facilitate with testing/debugging. Does that seem reasonable to you?

That sounds mostly OK but I'd like to propose a slight tweak.

    ignore_basepython_conflict = True | False (default)

Where `False` preserves the current behaviour in this version but raises a warning and `True` gives me the behaviour I described above. Once tox 4.0 is released, we could bump this warning to an exception. What do you think?

gaborbernat:
Ok. Make the changes and we can roll with it.

doc/conf.py
```diff
@@ -47,7 +47,7 @@
 def setup(app):
     # from sphinx.ext.autodoc import cut_lines
     # app.connect('autodoc-process-docstring', cut_lines(4, what=['module']))
-    app.add_description_unit(
+    app.add_object_type(
```
gaborbernat: please explain this change :+1:

doc/config.rst
```diff
@@ -81,12 +93,19 @@ Complete list of settings that you can put into ``testenv*`` sections:
...
+
+    .. versionchanged:: 3.1
+
+       Environments that use a :ref:`default factor <factors>` now ignore this
+       value, defaulting to the interpreter defined for that factor.
 
 .. confval:: commands=ARGVLIST
 
-    the commands to be called for testing. Each command is defined
+    The commands to be called for testing. Each command is defined
```
gaborbernat: please explain this change :+1:

doc/config.rst
```diff
@@ -114,38 +133,36 @@ Complete list of settings that you can put into ``testenv*`` sections:
...
-
 .. confval:: ignore_errors=True|False(default)
 
     .. versionadded:: 2.0
 
-      If ``True``, a non-zero exit code from one command will be ignored and
-      further commands will be executed (which was the default behavior in tox <
-      2.0).  If ``False`` (the default), then a non-zero exit code from one command
-      will abort execution of commands for that environment.
+    If ``True``, a non-zero exit code from one command will be ignored and
```
gaborbernat: please explain this change :+1:

doc/config.rst
```diff
@@ -205,51 +223,53 @@ Complete list of settings that you can put into ``testenv*`` sections:
...
-   .. versionadded:: 2.0
+    .. versionadded:: 2.0
 
-   A list of wildcard environment variable names which
-   shall be copied from the tox invocation environment to the test
-   environment when executing test commands.  If a specified environment
-   variable doesn't exist in the tox invocation environment it is ignored.
-   You can use ``*`` and ``?`` to match multiple environment variables with
-   one name.
+    A list of wildcard environment variable names which
```
gaborbernat: please explain this change :+1:

doc/config.rst
```diff
@@ -205,51 +223,53 @@ Complete list of settings that you can put into ``testenv*`` sections:
...
+        in ``passenv`` a warning will be emitted.
 
 .. confval:: recreate=True|False(default)
 
     Always recreate virtual environment if this option is True.
 
 .. confval:: downloadcache=path
 
-    **IGNORED** -- Since pip-8 has caching by default this option is now ignored.  Please remove it from your configs as a future tox version might bark on it.
+    **IGNORED** -- Since pip-8 has caching by default this option is now
```
gaborbernat: please explain this change :+1:

src/tox/config.py
```diff
@@ -505,11 +505,32 @@ def setenv(testenv_config, value):
...
+        for factor in testenv_config.factors:
+            if factor in tox.PYTHON.DEFAULT_FACTORS:
+                default = tox.PYTHON.DEFAULT_FACTORS[factor]
+
+                if value is None or testenv_config.config.ignore_basepython_conflict:
+                    return default
+
+                if str(value) != default:
+                    # TODO(stephenfin): Raise an exception here in tox 4.0
+                    print(
```
gaborbernat: let's use the warning system instead, https://docs.python.org/2/library/warnings.html; as the user can turn off that if they want

tox.ini
```diff
@@ -25,9 +25,7 @@ description = invoke sphinx-build to build the HTML docs and check that all link
 whitelist_externals = sphinx-build
 basepython = python3.6
 extras = docs
-changedir = {toxinidir}
```
gaborbernat: why did you remove this, I think that's why things fail now

src/tox/config.py
stephenfin: Sounds good. I wasn't able to find any other examples of warnings in use so I assumed you didn't use them.

stephenfin:
@gaborbernat I have a bad habit of using GitHub like I use mailing lists and Gerrit (including force pushes) :smile: The answer to all your questions above is in the [commit messages](https://github.com/tox-dev/tox/pull/841/commits). Is this what you're looking for?

doc/conf.py
stephenfin: See https://github.com/tox-dev/tox/pull/841/commits/116afb14d12fc16e2f7a98b685bd7c7bab2e125c

doc/config.rst
stephenfin: See https://github.com/tox-dev/tox/pull/841/commits/f52eea73ba00094c46979851551c74fc3e7de9ec

gaborbernat:
Yeah, migrate to the warnings api and we're good.

stephenfin:
> Yeah, migrate to the warnings api and we're good.

This is done now.

gaborbernat:
@stephenfin thank you very much!

PR #842, opened 2018-06-09
upgrade linters

gaborbernat:
just a small change, powered by the awesome ``pre-commit autoupdate``; we need to automate this even more via a GH bot :laughing:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     rev: v0.2.0
     hooks:
     -   id: blacken-docs
-        additional_dependencies: [black==18.6b1]
+        additional_dependencies: [black==18.6b2]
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
     rev: v0.2.0
     hooks:
     -   id: blacken-docs
-        additional_dependencies: [black==18.6b1]
+        additional_dependencies: [black==18.6b2]
```
gaborbernat: @asottile can we make this additional dependency to be also auto-updated somehow?

asottile: `pre-commit` currently doesn't know much about the contents of `additional_dependencies`, it's certainly _possible_ though I imagine pretty difficult

PR #857, opened 2018-06-23
docs: cleanup

blueyed:
While messing with Sphinx and tox's docs, this is something that makes sense already.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -104,7 +104,6 @@ Current features
    config
    support
    changelog
-   links
    plugins
    developers
    example/result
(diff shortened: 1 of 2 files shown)


doc/index.rst
```diff
@@ -104,7 +104,6 @@ Current features
    config
    support
    changelog
-   links
```
asottile: kinda funny, you can navigate to this "page" too: http://tox.readthedocs.io/en/latest/links.html

I wonder if we can trick rtfd by underscore prefixing it

blueyed: We should just remove it, no?

http://tox.readthedocs.io/en/latest/_links.html does not work (404).

asottile: it's a bit of a _hack_ to DRY links I think? see the `include` lines at the bottom of each file

asottile: (my suggestion here was to maybe `git mv links.rst _links.rst` and update the includes -- with the thought that maybe rtfd wouldn't serve an html page that's underscore prefixed (entirely a guess, I have no evidence / experience to back this up)

PR #858, opened 2018-06-23
Make py2 and py3 aliases also resolve via `py` on windows

### LINKED ISSUE #856: windows: py2 / py3 aliases do not resolve to `py -2` / `py -3`
When trying to write cross-platform `tox.ini` files it's important to use the `pyXY` aliases when configuring environments, as these work on Windows and Unix (using `pythonX.Y` is Unix-only, as the versioned Python executables don't exist on Windows).

But for environments that want to run just with "python 3" and which don't care about the precise version (a common example is running linters, doc builds or similar) there is currently no cross-platform solution.

I propose adding `py2` and `py3` aliases, similar to the existing `py27`/`py36` ones, which run the corresponding versions of Python (`python2` or `python3` on Unix, and `py -2` or `py -3` on Windows).

Supersedes #854 and #855

asottile:
Resolves #856 

CC @pfmoore

### DIFF
--- a/changelog/856.feature.rst
+++ b/changelog/856.feature.rst
@@ -1,0 +1,1 @@
+Make ``py2`` and ``py3`` aliases also resolve via ``py`` on windows
--- a/src/tox/interpreters.py
+++ b/src/tox/interpreters.py
@@ -143,11 +143,12 @@ else:
             return p
         actual = None
         # Is this a standard PythonX.Y name?
-        m = re.match(r"python(\d)\.(\d)", name)
+        m = re.match(r"python(\d)(?:\.(\d))?", name)
+        groups = [g for g in m.groups() if g] if m else []
         if m:
             # The standard names are in predictable places.
-            actual = r"c:\python{}{}\python.exe".format(*m.groups())
-        if not actual:
+            actual = r"c:\python{}\python.exe".format("".join(groups))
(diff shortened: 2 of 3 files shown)


gaborbernat:
Take a look at the ci as it broke on UNIX.

asottile:
heh, this was _passing_ for the wrong reason before (the assertion succeeded because the exit code of `check_call` was never verified) -- I guess I can add a similar `-c ''` check for unix as well

asottile:
@gaborbernat should be reviewable now!

gaborbernat:
@asottile ``py2`` and ``py3`` work as aliases for the ``basepython`` property, not?

asottile:
This enables the following things:

- `tox -e py2` + `tox -e py3` to work on windows (they already work on posix)
- `basepython=python2` / `basepython=python3` to work on windows (they already work on posix, and `basepython=pythonX.X` already works on windows)

This does not enable anything (imo) nonstandard like `basepython=py2`

pfmoore:
Thanks for this! Sorry I didn't have time to test it - you guys are to fast for me :smile:

gaborbernat:
We plan to do a release with ``3.1`` next week together with #852. If you find any issues with that feel free to come back and fill a bug.

PR #860, opened 2018-06-26
Replace the PEP-386 parser for the more general PEP-440

gaborbernat:
This means we drop the vendor and not really tested ``_verlib.py`` (PEP-386) in favour of a more general and battle-tested PEP-440 implementation of the packaging library (new install requires dependency).

Slight refactor of the functions affected to make them more readable.


setup.py
```diff
@@ -50,7 +50,13 @@ def main():
         },
         python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
         setup_requires=["setuptools_scm"],
-        install_requires=["py>=1.4.17", "pluggy>=0.3.0,<1.0", "six", "virtualenv>=1.11.2"],
+        install_requires=[
+            "py >= 1.4.17, <2",
+            "pluggy >= 0.3.0, <1",
+            "six >= 1.0.0, <2",
+            "virtualenv >= 1.11.2",
+            "packaging  >= 17.1",
```
asottile: let's sort these since we're rewriting them in this PR

src/tox/venv.py
```diff
@@ -184,21 +184,21 @@ def _getliveconfig(self):
...
-    def _getresolvedeps(self):
-        deps = []
-        for dep in self.envconfig.deps:
-            if dep.indexserver is None:
-                res = self.session._resolve_pkg(dep.name)
-                if res != dep.name:
-                    dep = dep.__class__(res)
-            deps.append(dep)
-        return deps
+    def _get_resolved_dependencies(self):
```
asottile: I realize it's not a public interface, but this breaks [at least this plugin](https://github.com/tox-dev/tox-pip-extensions/blob/3b62f3a331478329e80a16a61610fd72d0d4b47d/tox_pip_extensions.py#L111)

gaborbernat: Hmm, I was careful to not change anything that's public. If there's a need we should expose it as public, rather than encourage people using a private interface. :thinking:

asottile: let's not needlessly break this as part of this PR

gaborbernat: I find it hard to argue against breaking a private function. I'm inclined to break it just to force people not relying on it.

asottile: I find it hard to advocate for:
- renaming variables and functions unrelated to PR
- needlessly breaking compatibility
- invalidating `git blame`

if you were removing the function or changing its behaviour then fine, break compatibility -- but when there's no replacement, no migration pathway, and in a `tox-dev` plugin I can't see why you'd intentionally advocate for breaking something (that seems borderline malicious!)

asottile:
I'd also comb through the official plugins and make PRs if you're going to change it -- breaking backward compatibility should not be taken lightly :)

src/tox/venv.py
gaborbernat: @asottile 

Both the functions and variables are related to PR: I've refactored the code handling the dependency parsing: As part of the dependency parsing the version parsing is used and as such the version parser is invoked via the ``_getresolvedeps`` path.

gaborbernat:
Both the functions and variables are related to PR: I've refactored the code handling the dependency parsing: As part of the dependency parsing the version parsing is used and as such the version parser is invoked via the ``_getresolvedeps`` path.

Private functions are that private. They don't promise any compatibility. So there was no compatibility guarantee there in the first place. 

Git blame has little value overall (especially given no one in the git blame are maintaining the library currently), compared to maintaining the library. Recently I've tried adding type information to the library, and the version library what we don't test was a major pain point to showcase an example. I value an easy to read code over maintainability any day. ``_getresolvedeps`` break at least snake case usage as far as readability goes. 

I'm changing the behaviour ever so slightest, as note PEP-440 is a bit more permissive than PEP-338. Private functions are **private**, they don't need a migration path. That being said let's meet in the middle:

I've kept the private function as it was but made it deprecated. We'll remove later on. As migration path, you can use the public ``get_resolved_dependencies`` function in the future.

tests/test_z_cmdline.py
```diff
@@ -45,41 +44,6 @@ def wait(self):
     report.expect("logpopen")
 
 
-def test__resolve_pkg(tmpdir, mocksession):
```
gaborbernat: This file at the moment is more or less a general collection of everything else. We'll slowly phase it out. For now I've just moved out tests related to the version parsing, that have nothing to do with the command line.

obestwalter:
Nice one :+1: 

@gaborbernat, what's your opinion of the tests for this? Are we covering enough with the existing tests to be reasonably sure that this does not introduce any surprises? I did not have the time to have a closer look or even to check the branch out and play with it to look for potential problems, so I am merely curious at the moment.

gaborbernat:
@obestwalter I think we're covering enough. That being said this *might* break some people; who were relying on the less permissive PEP-386 logic. E.g. ``1.4.3dev`` now is accepted as ``1.4.3dev0``; while beforehand it failed. However, according to the latest PyPi rules, the aforementioned should be accepted as ``1.4.3dev0``, so I would consider this a bug fix.

PR #864, opened 2018-07-01
Prepare for first release by Gabor

obestwalter:
As @gaborbernat is driving the release this time, I had a look at what grew over the last year and all in all it's not too bad I guess. But those release projects I started to use for each release should be burned.

Projects are a major PITA to keep up to date and I don't think they provide much value. As long as we properly document the changes in the CHANGELOG, we're golden. So I removed the links to the old projects in the CHANGELOG and we can forget about them.

The only advantage these projects had, was that after wading through all the closed issues and assigning them to the right swimlanes in the project it was easy to see how many issues were closed and how many PRs were merged. I am sure there are other ways to find that out and all in all it is not really that important to know this anyway.

I also updated the release howto with some info that was still only living on my hard disk (special releases and release mail templates).

Looking forward to 3.1.0rc1 then - if I can help in any way, let me know :)


HOWTORELEASE.rst
```diff
@@ -80,6 +86,62 @@ Check that tests pass for relevant combinations with::
...
+
+    testing-in-python@lists.idyll.org, tox-dev@python.org
+
+
+feature release::
+
+    The tox team is proud to announce the <VERSION> release!
+
+    tox aims to automate and standardize testing in Python. It is part of a
+    larger vision of easing the packaging, testing and release process of
```
asottile: I would have oxford comma 🙃

PR #865, opened 2018-07-02
[wip] Add environment variables to detect running from within tox (#847)

### LINKED ISSUE #847: Set TOX_ENVDIR environment variable
I think it would be nice if tox would set some environment variable(s), which would help to detect being run in tox.

I suggest setting `TOX_ENVDIR` to what the `{toxenvdir}` substitution provides, since this would allow tools like pytest-testmon to place their data files there.

Ref: https://github.com/tarpas/pytest-testmon/pull/100

gaborbernat:
- ``TOX_WORK_DIR`` for session runs (only this set during packaging)
- ``TOX_ENV_NAME`` and ``TOX_ENV_WORK_DIR`` during the execution of a given virtual environment
- add a section inside documentation that will focus on the high-level information about tox (for now contains the existence of this os environment variable injection).

Resolves #847.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -1,0 +1,13 @@
+Overview
+========
+This section contains system level information about tox.
+
+Environment variables
+---------------------
+tox will inject the following environment variables that you can use to test that your command is running within tox:
+
+.. versionadded:: 3.1
+
+- ``TOX_WORK_DIR`` env var is set to the tox work directory
+- ``TOX_ENV_NAME`` is set to the current running tox environment name
+- ``TOX_ENV_DIR`` is set to the current tox environments working dir.
--- a/src/tox/config.py
+++ b/src/tox/config.py
@@ -611,7 +611,16 @@ def tox_addoption(parser):
         # Flatten the list to deal with space-separated values.
         value = list(itertools.chain.from_iterable([x.split(" ") for x in value]))
 
-        passenv = {"PATH", "PIP_INDEX_URL", "LANG", "LANGUAGE", "LD_LIBRARY_PATH"}
+        passenv = {
+            "PATH",
+            "PIP_INDEX_URL",
+            "LANG",
+            "LANGUAGE",
+            "LD_LIBRARY_PATH",
+            "TOX_WORK_DIR",
+            "TOX_ENV_NAME",
+            "TOX_ENV_WORK_DIR",
+        }
 
         # read in global passenv settings
         p = os.environ.get("TOX_TESTENV_PASSENV", None)
--- a/changelog/847.feature.rst
+++ b/changelog/847.feature.rst
@@ -1,0 +1,1 @@
+Set environment variables to indicate we are running inside tox: ``TOX_WORK_DIR`` env var is set to the tox work dir, ``TOX_ENV_NAME`` is set to the current running tox environment name, ``TOX_ENV_DIR`` is set to the current tox environments working dir. By :user:`gaborbernat`.
--- a/doc/_templates/localtoc.html
+++ b/doc/_templates/localtoc.html
@@ -1,37 +1,50 @@
-
 {%- if pagename != "search" %}
-<div id="searchbox" style="display: none">
-    <form class="search" action="{{ pathto('search') }}" method="get">
-      <input type="text" name="q" size="18" />
-      <input type="submit" value="{{ _('Search') }}" />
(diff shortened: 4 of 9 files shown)


doc/overview.rst
```diff
@@ -0,0 +1,13 @@
...
+
+Environment variables
+---------------------
+tox will inject the following environment variables that you can use to test that your command is running within tox:
+
+.. versionadded:: 3.1
+
+- ``TOX_WORK_DIR`` env var is set to the tox work directory
+- ``TOX_ENV_NAME`` is set to the current running tox environment name
+- ``TOX_ENV_DIR`` is set to the current tox environments working dir.
```
asottile: should we make these consistent with the names that get substituted in the configuraiton? https://tox.readthedocs.io/en/latest/config.html?highlight=substitution#substitutions

particularly `envname` but `ENV_NAME` seems off to me

should we export all of these substitution parameters as well?

do we even need this feature? someone can `setenv = TOX_ENV_NAME={envname}` already today

src/tox/config.py
```diff
@@ -611,7 +611,16 @@ def passenv(testenv_config, value):
...
-        passenv = {"PATH", "PIP_INDEX_URL", "LANG", "LANGUAGE", "LD_LIBRARY_PATH"}
+        passenv = {
+            "PATH",
+            "PIP_INDEX_URL",
+            "LANG",
+            "LANGUAGE",
+            "LD_LIBRARY_PATH",
+            "TOX_WORK_DIR",
+            "TOX_ENV_NAME",
+            "TOX_ENV_WORK_DIR",
```
asottile: hmmm, seems odd that they'd get whitelisted into `passenv` as well -- I guess that's an order-of-operations issue probably? (depending on when the variables get set)

if we're going forward with this, can we make tox augment `setenv` and then we don't need to worry about `passenv`?

rpkilby:
Would it also make sense to add `TOX=true`, so users can detect tox without depending on a more specific/functional variable? This would be analogous to how Travis sets `TRAVIS=true`.

doc/overview.rst
gaborbernat: Although one could do it partially with the ``setenv`` as discussed in the related #847 it adds value to have a general env var name. This way the tools can use to detect if we're inside tox and users no longer need to keep copying around the same set env for their project. 

Coming back to the ``setenv`` I don't think that cuts it. Mostly because setenv is only applied to the current command. I think we want a higher level to be able to be detected by both the package builder and the installer. These are also tools which we invoke; we're not talking here just about commands. In light of this can you have another look?

src/tox/config.py
gaborbernat: Because I wanted these environment variables to be usable across the tool lifecycle, not just the commands section, I change the values in the ``os.environ``. As such in order for the commands to inherit it we need them to be in the whitelist.

gaborbernat:
@rpkilby I could see some people wanting the work dir too besides just are we inside tox (e.g. ``pytest-testmon``). I would like to keep the injected environment variable count low, hence given the ``TOX_WORK_DIR`` already serves all the answers of a potential ``TOX`` env var, I don't think we want to add that.

doc/overview.rst
asottile: afaik setenv applies to all life cycles (I looked into this for tox-virtualenv-no-download)

gaborbernat: Hmmm, will double check.

asottile: Here's that analysis : https://github.com/asottile/tox-virtualenv-no-download/issues/4#issuecomment-362123210

obestwalter:
I definitely like the idea of doing this - I actually could swear that I opened an issue proposing this just after I joined the project but can't find it anymore ... 

Hi @gaborbernat I would really like to get my head around this PR first, so please wait for my feedback on this one. I would also propose not to add this to 3.1 yet. I think we have enough changes already to ship 3.1. We have some massive refactorings and there might be surprises, that are easier to catch, if we don't ship to many changes at once.

gaborbernat:
Ok let's move it to 3.2. As such I'll delay working on this for later.

obestwalter:
Was there a problem with the changes? I did not get around to check this out yet.

gaborbernat:
I did not get around to finish this. Will re-open later on; but given it's part of 3.2 can wait for now to fix the latest release.

PR #866, opened 2018-07-02
Adopt Python 3.7

gaborbernat:
- add Python 3.8 dev tests as failure allowed
- Python 3.7 tests are no longer allowed to fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,42 +1,68 @@
+group: travis_latest
 sudo: false
 language: python
 cache: pip
-python:
-  - 2.7
-  - 3.4
-  - 3.5
-  - 3.6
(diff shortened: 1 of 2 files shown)


asottile:
this is likely blocked on
- pyyaml making a release
- appveyor having python3.7 installed

gaborbernat:
I'll move ``fix-lint`` to Python 3.6 (we'll update it to 3.7 after https://github.com/yaml/pyyaml/issues/193). And for now, not run Windows test for Python 3.7 (we'll add it in a new PR once https://github.com/appveyor/ci/issues/2475 is released). 

With these three we can go ahead. Also defined stages to not run tests if the linting fails.

Issue #870, opened 2018-07-04
tox should warn if a command starts with -


psav:
If you have a "commands" section that starts with a dash, then it will always pass. This is dangerous especially if you have autodeploy on. Perhaps tox should warn users if a command starts with a dash.

gaborbernat:
Hello, can you detail a such example?

obestwalter:
Hi @psav, this is intentional and documented. Adding a dash is a way of saying that I am not interested in the return value of a specific command.

> Similar to make recipe lines, any command with a leading - will ignore the exit code.

https://tox.readthedocs.io/en/latest/config.html#confval-commands=ARGVLIST

obestwalter:
Although documented it was well hidden. Mentioning it in its own paragraph and adding a small example might improve visibility of this feature.

PR #874, opened 2018-07-05
Formatting, clarify phrasing

obestwalter:
Pulling out this functionality in its own paragraph, spelling out the name of the character (dash) and adding an example makes this feature more visible.

Might help with #870


gaborbernat:
You'll want to fix the linting 😁

asottile:
might also be useful to add something like I just added for `pytest`: https://github.com/pytest-dev/pytest/pull/3660

obestwalter:
Did through the web thingy ...

PR #875, opened 2018-07-06
release 3.1.0rc1

gaborbernat:
:+1:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,54 @@ Changes that already arrived in master, but are not released yet can be found as
 
 .. towncrier release notes start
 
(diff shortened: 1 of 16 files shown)


obestwalter:
woot!

gaborbernat:
The CI failed, will regroup.

obestwalter:
Looking good. One thing we loose by this automation, which is not dramatic though is merging all changes for the final release in a unifed version entry, removing the rc entries. Releasing like this wilh always have most changes in rc1 and only small fixes will be in the other rcs. But I guess, this "new" way is the "correct" way anyway. Merging and removing the rc entries for the final release is technically "lying" anyway.

PR #877, opened 2018-07-06
release 3.1.0a1

gaborbernat:
:+1: take two

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -8,11 +8,20 @@ coverage:
   precision: 2
   round: down
   range: "70...100"
-
   status:
-    project: yes
-    patch: yes
-    changes: no
+    project:
+      default:
+        threshold: null
+        if_no_uploads: success
+        if_not_found: success
+        if_ci_failed: success
+    patch:
+      default:
+        enabled: yes
+        threshold: 30
+        if_no_uploads: success
+        if_not_found: success
+        if_ci_failed: success
 
 parsers:
   gcov:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,10 @@
+if: tag IS present OR type = pull_request OR (branch = master AND type = push)   # we only CI the master, tags and PRs
+
+stages:
+- test
+- name: deploy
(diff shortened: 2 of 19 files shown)


.codecov.yml
```diff
@@ -8,11 +8,20 @@ coverage:
   precision: 2
   round: down
   range: "70...100"
-
   status:
-    project: yes
-    patch: yes
-    changes: no
+    project:
```
gaborbernat: Ok, let's report code coverage, but don't fail on it whatever the value is. The maintainer will make a judgment if the current value is ok or not (this takes care of flaky test reports, or cases when we don't want the CI to run fully - e.g. documentation only changes - AppVeyor).

.travis.yml
```diff
@@ -1,3 +1,10 @@
+if: tag IS present OR type = pull_request OR (branch = master AND type = push)   # we only CI the master, tags and PRs
```
gaborbernat: Build tags, pull requests and pushes on the master branch. Ignore all other.

.travis.yml
```diff
@@ -1,3 +1,10 @@
...
+
+stages:
+- test
+- name: deploy
+  if: repo = tox-dev/tox AND branch = master AND tag IS present
```
gaborbernat: Only deploy when tagging a commit on the master branch.

.travis.yml
```diff
@@ -64,6 +64,7 @@ before_install:
 install:
 - pip install -U six
 - pip install --pre -U tox
+- tox --notest
```
gaborbernat: move tox env setup to the setup phase

.travis.yml
nicoddemus: Interesting!

About tagging commit on master: I think it would be better to tag the commit at the HEAD on the release PR. This way you guarantee that you are publishing exactly what was tested on the PR.

nicoddemus: https://github.com/pytest-dev/pytest/pull/3666 😉 Thanks!

PR #878, opened 2018-07-08
Add an example of path variable substitution

williamjamir:
I know that is a minor detail, but I think that an example of how to update environment variable with the use of variable substitution is very handy here.


gaborbernat:
Thanks!

PR #880, opened 2018-07-08
release 3.1.0

gaborbernat:
:+1: 

This already went out, so per se it succeded. We'll just want to merge it in.


obestwalter:
congrats on your first relase :)

I noticed that the [rendering](https://pypi.org/project/tox/) on Pypi is broken though. This can be prevented by running a check over the generated changelog. I had this in the old semi automated script. This might not have survived the port from my old semi manual scripts but could protect us from that.

Maybe this can be fixed with a post release?

PR #883, opened 2018-07-09
Fix PyPi description does not render #879

### LINKED ISSUE #879: 3.1.0 release documentation remains raw
see https://pypi.org/project/tox/3.1.0/
screenshot-pypi org-2018 07 08-21-58-34

gaborbernat:
Added a test to check for this via tox env `package-description`. Using collective.checkdocs and `python setup.py checkdocs`. Note we could use restview too, however that requires a compilation step so decided against it. Users can use on an ad-hoc level the command `restview --pypi-strict --long-description` to visualize locally PyPi description.

Resolves #879.


obestwalter:
What is the reason for not using `readme-renderer` anymore?

From the old `tasks/pra.sh`

    pip install -U readme-renderer
    python setup.py check -r -s

This is the exact same tool that pypa uses for rendering in warehouse so I would say we should use this also to check.

gaborbernat:
Hmmm missed that that's what upstream uses will amend.

PR #884, opened 2018-07-09
RST code is two backticks

gaborbernat:
By using @asottile script from ``pytest``.


gaborbernat:
@obestwalter in case you wonder why the coverage is fluctuating, I've disabled running tests on AppVeyor that do not touch source files. Sadly Travis does not support similar by default. Hence documentation fixes have a lower coverage... Will need to see how we can bypass running tests for deltas that don't change the Python files.

obestwalter:
I don't really worry about that knowing it's a bogus result from the changes, but I would really prefer if we would get rid of the artificial preciseness re-introduced recently. 

What I am wondering about though is why whe introduction of the src layout seems to have resulted in a drop in coverage of about 3% - https://codecov.io/gh/tox-dev/tox/branch/master/commits?from=2018-06-10%2000%3A00%3A00&to=2018-06-19%2000%3A00%3A00.

gaborbernat:
What do you consider artificial preciseness?

obestwalter:
the two digits behind the comma :) - it's technically "correct", but from a laymans psychological standpoint I prefer to keep it clear even from the display and the expectations that coverage is in interesting metric that gives you information about your test coverage but not a precise measuring tool or even something that we should use as a target.

gaborbernat:
I tend to disagree, I think the two digits behind the comma offers some reward when adding coverage for previously uncovered lines. Granted there's no right number for coverage, but generally, a higher number is prefered over a lower number.

obestwalter:
I haven't seen the reward aspect yet and I do not have a super strong opinion about the whole thing anyway, which is why I approved your revert, although I would prefer to keep it more "fuzzy". 

Any idea though why coverage went down 3 points with the switch to src layout?

anthrotype:
> What I am wondering about though is why whe introduction of the src layout seems to have resulted in a drop in coverage of about 3%

probably because with non-src layout, by simply importing the top-level module you would execute and thus "cover" the lines in the global module scope.

obestwalter:
I wouldn't have thought that shouldn't make a difference. Code is either executed/imported as part of a test or not independent of the layout (is my naive view at least). Snyway - I am interested now and will have a look at what exactly changed.

obestwalter:
Well, this is the first time that I am rying to make sense of some more detailed coverage info from codecov and I am not really able to make any sense quickly. 

According to codecov in the change where the coverage supposedly dropped 16 "new" files where added, but they aren't new at all and their specific coverage is exactly the same as before:
https://codecov.io/gh/tox-dev/tox/commit/4e7d9091f1f7ffd8042936fe25cd2f66e196f57c

At least superficially this rather looks like we hit some kind of codecov bug, as the coverage should not have changed at all.

gaborbernat:
Yeah, will need to have a better look to understand.

PR #885, opened 2018-07-09
Use git commit history to have a more accurate contributors list

gaborbernat:
Comparing against the list we can see quite a few people omitted. Our Github bot should check automatically that new users add themselves (or even better ideally automatically inject it if it detects not added).

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,62 +1,130 @@
-Alexander Schepanovski
-Alex Grönholm
 Alexander Loechel
+Alexander Schepanovski
 Alexandre Conrad
+Alex Gaynor
+Alex Grönholm
 Allan Feldman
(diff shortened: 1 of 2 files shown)


obestwalter:
I am not sure, if we should do this. Being in the list of contributors is a voluntary opt-in, I'd say. There were even cases, where someone provided a fix, and I suggested that they add themselves to the list of contributors and they politely refused. I myself usually don't add myself to the contributors in other projects either, when I provided a trivial fix. There is not only code contributors either. If someone does mainly org stuff and what you would call open source gardening, but never contributes a line of code, they still should be on the contributors list, if they want to. So I don't see that strict connection between code contrib and contib list anyway.

gaborbernat:
Fair enough :+1:

PR #887, opened 2018-07-09
CI optimization

gaborbernat:
1. Run all tests on release commit.
2. Do not run pytest if the source files do not change.
3. Remove Makefile for Sphinx.


.travis.yml
```diff
@@ -61,17 +60,25 @@ matrix:
...
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then source .travis-osx; fi
+- export BUILD="setup.py pyproject.toml MANIFEST.in setup.cfg tox.ini .travis.yml appveyor.yml"
+- export SRC=$(git diff --name-only $TRAVIS_COMMIT_RANGE -- '*.py' $BUILD | grep -v conf.py | grep -v "tasks/.*.py")
+- [[ "$TOXENV"!="py" || (-n "$SRC" && "$TOXENV"="py")]] && export RUN=1
 
 install:
-- pip install -U six
-- pip install --pre -U tox
-- tox --notest
+- [ -n "$RUN" ] && pip install --pre -U tox
```
obestwalter: `[ -n "$RUN" ]` - never saw this before what does this do? Can you point me to some documentation? Would be interested to learn more about these options.

doc/Makefile
```diff
@@ -1,130 +0,0 @@
-# Makefile for Sphinx documentation
```
obestwalter: Good call - nobody used this anymore for a long time, I guess.

.travis.yml
gaborbernat: http://wiki.bash-hackers.org/commands/classictest#string_tests

obestwalter: thx

PR #892, opened 2018-07-09
WIP - Add VSTS integration

gaborbernat:
Try out using VSTS.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -32,3 +32,4 @@ phases:
     inputs:
       testResultsFiles: '.tox/test-results.*.xml'
     condition: always()
+  - script: echo hello world


zooba:
@gaborbernat I'm writing you an email now, but you'll need to point VSTS at your fork to get it to use this file. It won't use a definition that's new in a PR (though I believe it will use a newer version in a PR if it was already there in the target branch)

gaborbernat:
I just put the base into the master; and in this PR I just update it. It still does not work like that? :+1:

zooba:
It looks like you may also need to reauthorize VSTS. The errors on the builds look like an auth issue...

zooba:
For extra joy, you should add `--junitxml=test-results.xml` to your pytest command line and then add this snippet after running the tests:

```yaml
- task: PublishTestResults@2
  displayName: Publish Test Results **/test-results.xml
  inputs:
    testResultsFiles: '**/test-results.xml'
  condition: always()
```

This will upload test results to the TEST tab.

gaborbernat:
Yeah gathered, for now I'm more worried about the PR hook working though :thinking: did the re-enable authentication still nothing.

zooba:
Well, we got a successful build at least :)

You might need to add this to the YAML file (copied from the CPython repo, so adjust branches/paths as necessary - and I'm fairly sure we're still waiting for the paths parameter to be supported for GitHub, so it may not be working yet):

```yaml
trigger:
  branches:
    include:
    - master
    - 3.7
    - 3.6
  paths:
    exclude:
    - Doc/*
    - Tools/*
```

zooba:
And it may also need to be merged before it gets picked up (since the main repo is going to provide the trigger settings).

gaborbernat:
Ok, done :+1: still not triggering :thinking: the PR webhook returns with ``{"count":0,"value":[]}``

PR #893, opened 2018-07-09
Explicitly specify AppVeyor image to VS 2017

gaborbernat:
:+1:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,5 @@
+image: "Visual Studio 2015"
+
 build: false  # Not a C# project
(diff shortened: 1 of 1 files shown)


gaborbernat:
It's a test to see build time difference compared to VS 2015.

nicoddemus:
I'm curious, what was the conclusion here?

gaborbernat:
Still a work in progress :thinking:

nicoddemus:
Weird, I wouldn't expect any change actually. Oh well. 😁

Issue #478, opened 2017-03-12
Naming and describing phases of a tox test run


obestwalter:
(See https://github.com/tox-dev/tox/issues/421#issuecomment-285889183 and the following comments for the initial discussion)

Summary: a tox run goes through distinct phases. This is implicitly explained in the documentation and becomes obvious when working a while with tox (especially when running `tox -vv`. Nevertheless it might help make this a bit more explicit in e.g. the documentation. It also might help in further development to have some "official" terminology for those phases. Might also help with naming of new hooks, etc.

If you look into the code there is the [action class](https://github.com/tox-dev/tox/blob/2.6.0/tox/session.py#L71), which has reporting hooks to log start and finish of those actions. Those are phases in one respect, but don't map entirely satisfactorily to what the user might think of as phases. The actions that are reported are: packaging, getenv, finishvenv, developpkg, installpkg, envreport, runtests. They are also marked by the context in which they run, which is **GLOB** for global context and **(some envname)** for environment context (which might be one of the default environments or whatever environment the user has created).

The test run is executed in [subcommand_test](https://github.com/tox-dev/tox/blob/2.6.0/tox/session.py#L530).

Those map vaguely onto what a tox user might think about how to organize things (and maybe where to hook into to change the behaviour), but not quite. e.g. finding the interpreter executable is not mirrored as action. I have not looked into the code in enough detail why that is.

Looking at a tox run from the outside, the phases of a test run (meaning: show help, show config or show envs not taken into account here) could be described as:

* **GLOB**  initialization
* **GLOB**  package build (optional)
* **ENV** interpreter discovery
* **ENV** environment creation
* **ENV** dependency installation (optional)
* **ENV** package under test installation (optional)
* **ENV** command execution (optional) (not atm (#454) but with next release again).
* **GLOB**  report results

gaborbernat:
shouldn't there be another reporting phase at the end, again with GLOB scope?

gaborbernat:
I've created a diagram of the stages, and the flow of the application:

![Tox phases](http://i.imgur.com/RvMPaM2.png)

Let me know what you think :+1:

obestwalter:
Great! Which tool did you use for that? Would it get too cuttered, if it would be mentioned which hook are called in which phase?

gaborbernat:
used Lucid chart diagram, I can give it a try and see how it looks; you are referring to the plugin hooks, yes?

obestwalter:
[...]

PR #897, opened 2018-07-10
Add a system overview section on the landing page #867

### LINKED ISSUE #867: general overview of tox - documentation
Hi, 
I would like to understand what is the basic flow of tox (which is an amazing project by the way !):
- I understand there is a run of setup.py in both environments (2.7, 3.6).
- But then, regarding the tests ran with pytest (or other test suite): are they running against package in source code, or against package installed in site-packages (by setup.py install) ? 

It's important in my case because my setup.py downloads a binary that I need and puts it in the site-packages directory, which is not originally in my source code package (because I need it to be cross-platform).

Thanks for help !

gaborbernat:
I find we lack a quick two-minute high-level overview of what tox is. This should help new people quickly understand easier what tox actually does and expose common gotchas.

See output at https://tox-gaborbernat.readthedocs.io/en/general/#system-overview.

Resolves #867.


obestwalter:
Oh good, this would also address where we started discussing these things: https://github.com/tox-dev/tox/issues/478.

What should be covered though in the explanation is, that the visualized default flow is highly configurable (e.g. you could not build a package and you could also switch off installing the package (and use tox in a `setup.py` less project) or you could install a devlopment version. 

I have no idea yet how and where this should be visualized/documented, but I think it should definitely be part of a high level overview to make clear that tox can do much more than the standard flow.

gaborbernat:
I know, and if you read the explanation for the phases you can see I pointed out clearly where things change behaviour.

obestwalter:
I have to admit that I did not read very carefully on the first go, I will have a closer look tonight and might have some ideas for improvments.

obestwalter:
Hi @gaborbernat, I will be able to dedicate next friday to tox properly and try to catch up then.

obestwalter:
BTW - did you watch the video? What do you think about us trying to apply this to the tox documentation?

gaborbernat:
I did not watch the video but was there in the audience. It makes some good points. As far as our project goes I think we just need to have a different page for the plugins.

obestwalter:
I think I will take the friday to have a proper look at our existing docs and go from there.

doc/index.rst
```diff
@@ -55,6 +55,72 @@ you'll note that it runs much faster because it keeps track of virtualenv detail
...
+
+3. **environment** - for each tox environment (e.g. ``py27``, ``py36``) do:
+
+    1. **environment creation**: create a fresh environment, by default Python virtual environments
+    are used via the virtualenv_ project. tox will automatically
+    try to discover a valid Python interpreter version by using the environment name (e.g. ``py27``
+    means Python 2.7 and the ``basepython`` configuration value) and the current operating system
+    ``PATH`` value. This step is done at first run only (or when :confval:`basepython` value
+    changes) and re-used at subsequent runs, unless the user requests explicitly recreating this
+    by passing in the ``-r`` flag.
```
obestwalter: A bit more handwavy but maybe sufficient here?

    This is created at first run only to be re-used at subsequent runs. If certain aspects of the project change, a re-creation of the environment is automatically triggered. To force the recreation (e.g. when changes weren't properly picked up), tox can be invoked with ``-r``/``--recreate``.

doc/index.rst
```diff
@@ -55,6 +55,72 @@ you'll note that it runs much faster because it keeps track of virtualenv detail
...
+
+    2. **install** (optional): install the environment dependencies specified inside the
+    :confval:`deps` configuration section, and then the earlier packaged source distribution.
+    By default ``pip`` is used to install packages, however one can customise this via
+    :confval:`install_command`. Note ``pip`` will not update project dependencies (specified either
+    in the ``install_requires`` or the ``extras`` section of the ``setup.py``) if any version already
+    exists in the virtual environment; therefore we recommend to recreate your environments entirely
```
obestwalter: "entirely" is not necessary here (and maybe even potentially confusing?), as there is no way to recreate the environment partially.

doc/index.rst
```diff
@@ -55,6 +55,72 @@ you'll note that it runs much faster because it keeps track of virtualenv detail
...
+    whenever your project dependencies change.
+
+    3. **commands**: run one by one the specified commands. Whenever the exit code of any of them
```
obestwalter: How about:

    run the specified commands in the specified order.

doc/index.rst
```diff
@@ -55,6 +55,72 @@ you'll note that it runs much faster because it keeps track of virtualenv detail
...
+    is not zero stop, and mark the environment failed. Note, starting a command with a single dash
+    character means ignore exit code.
```
obestwalter: Or:

   To ignore the exit code of a command, a dash can be prefixed to the command.

doc/index.rst
```diff
@@ -55,6 +55,72 @@ you'll note that it runs much faster because it keeps track of virtualenv detail
...
+6. **report** print out a report of outcomes for each tox environment:
+
+   .. code:: bash
+
+      ____________________ summary ____________________
+      py27: commands succeeded
+      ERROR:   py36: commands failed
+
+    Only if all environments finished with success will the exit code of tox be success. In this
+    case you'll also see the message ``congratulations :)``.
```
obestwalter: or:

    Only if all environments ran successfully tox will return `0` (success). In this case you'll also see the message ``congratulations :)``.

doc/index.rst
```diff
@@ -55,6 +55,72 @@ you'll note that it runs much faster because it keeps track of virtualenv detail
...
+
+tox will take care of environment isolation for you: it will strip away all operating system
+environment variables not specified via :confval:`passenv`. Furthermore, it will also alter the
+``PATH`` variable so that your commands resolve first and foremost within the current active
+tox environment. It's possible to fallback to executables outside of this, however these need to
+be explicitly enabled via the :confval:`whitelist_external` configuration variable.
```
obestwalter: This is not quite accurate: tox uses whatever is available, but will warn when it was not expllicitly whitelisted.

How about:

    In general all executables in the path are available in ``commands``, but tox will emit a warning if it was not explicitly allowed via :confval:`whitelist_external`.

doc/index.rst
```diff
@@ -55,6 +55,72 @@ you'll note that it runs much faster because it keeps track of virtualenv detail
...
+
+      python setup.py sdist
+
+   Note that for this operation the same Python environment will be used as the one tox is
+   installed into (therefore you need to make sure that it contains your build dependencies).
+   Skip this step for application projects that don't have a ``setup.py``.
+
+3. **environment** - for each tox environment (e.g. ``py27``, ``py36``) do:
+
+    1. **environment creation**: create a fresh environment, by default Python virtual environments
```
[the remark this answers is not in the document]
gaborbernat: I prefer having it as phase names, to help new users grasp tox run concepts. Hence I still think **environment creation** is better.

doc/index.rst
gaborbernat: I'll use most of it other than the `` (e.g. when changes weren't properly picked up)``. Some changes are not tracked, but I don't think that warrants it being stated not properly.

doc/index.rst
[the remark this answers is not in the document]
obestwalter: Sorry if that wasn't clear - I wasn't referring to the phase name, but to the explanation after that. Basically to shorten the first sentence and add the replacement info via plugin there as well (but I am not sure, if that is not overloading that part).

PR #901, opened 2018-07-10
Add a test to ensure all rst ini blocks parse correctly

### LINKED ISSUE #900: ensure tox.ini examples in our documentation can be parsed
Would be nice if we could ensure that our examples in our documentation parse correctly. 

@asottile @nicoddemus how hard would it be to generate for pytest extra tests by parsing the ``rst`` files inside the documentation, looking for ``.. codeblock:: ini`` and wrapping it into a generated test (e.g. ``parse_config("ini")``). I assume this requires adding an additional test discovery mechanism and an additional test runner for it. Does anyone have examples of similar? Docstrings are sort of similar to this.

asottile:
Resolves #900

### DIFF
--- a/tests/test_docs.py
+++ b/tests/test_docs.py
@@ -1,0 +1,53 @@
+import os.path
+import re
+import textwrap
+
+import pytest
+
+from tox.config import parseconfig
+
+INI_BLOCK_RE = re.compile(
+    r"(?P<before>"
+    r"^(?P<indent> *)\.\. (code-block|sourcecode):: ini\n"
+    r"((?P=indent) +:.*\n)*"
+    r"\n*"
+    r")"
+    r"(?P<code>(^((?P=indent) +.*)?\n)+)",
+    re.MULTILINE,
+)
+
+
+RST_FILES = []
+TOX_ROOT = os.path.join(os.path.dirname(os.path.dirname(__file__)))
+for root, _, filenames in os.walk(os.path.join(TOX_ROOT, "doc")):
+    for f in filenames:
+        if f.endswith(".rst"):
+            RST_FILES.append(os.path.join(root, f))
+
+
+def test_some_files_exist():
+    assert RST_FILES
+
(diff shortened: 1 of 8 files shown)


tests/test_docs.py
```diff
@@ -0,0 +1,32 @@
...
+    r"((?P=indent) +:.*\n)*"
+    r"\n*"
+    r")"
+    r"(?P<code>(^((?P=indent) +.*)?\n)+)",
+    re.MULTILINE,
+)
+
+
+rst_files = []
+for root, _, filenames in os.walk("doc"):
```
gaborbernat: does pytest guarantee the current working directory is the root; or should we do some kind of relative from this file logic here?

asottile: it doesn't if some other test runs `chdir` then this points at the wrong file locations -- I can `abspath` + join with `__file__` to make this better

gaborbernat: Yeah I think that would make this more robust; otherwise looks good :+1:

gaborbernat: Also in the CI I see:

```
SKIP [1] tests/test_docs.py:27: got empty parameter set ['filename'], function test_all_rst_ini_blocks_parse at /home/travis/build/tox-dev/tox/tests/test_docs.py:26
```

asottile: heh, probably due to the `changedir`

obestwalter:
Nice one :)

nicoddemus:
👍 simple and to the point. 😁

asottile:
I suspect the tests are failing (and unreliably) due to random test ordering -- I'll see if I can't improve that...

asottile:
hah wow, this wasn't actually testing what I expected -- turns out `parseconfig` will just "try its best" when fed garbage.

asottile:
want to re-review now that I've rewritten a bunch of `.rst`?

PR #902, opened 2018-07-10
VSTS demo

gaborbernat:
:+1:

### DIFF
--- a/.travis-osx
+++ b/.travis-osx
@@ -1,6 +1,0 @@
-# Perform the manual steps on osx to install python3 and activate an environment
-brew update
-brew upgrade python
-rm /usr/local/bin/python
-ln -s python3 /usr/local/bin/python
-ln -fs pip3 /usr/local/bin/pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,25 +17,13 @@ env:
 
 jobs:
   include:
-  - python: 2.7    # 2.7.14 pip  9.0.1
   - python: 3.4    # 3.4.6  pip  9.0.1
     dist: trusty
-  - python: 3.5    # 3.5.5  pip  9.0.1
-  - python: 3.6    # 3.6.3  pip  9.0.1
-  - python: 3.7    # 3.7.0  pip 10.0.1
   - python: 3.8-dev
   - python: pypy   # 2.7.13 pip  9.0.1
     dist: trusty
   - python: pypy3  # 3.5.3  pip  9.0.1
     dist: trusty
-  - os: osx
-    language: generic
-  - python: 3.7
-    env: TOXENV=docs
-  - python: 3.7
-    env: TOXENV=fix-lint
-  - python: 3.7
-    env: TOXENV=package-description
 
   - stage: deploy
     python: '3.7'
@@ -59,7 +47,6 @@ matrix:
   - python: 3.8-dev
   - python: pypy
   - python: pypy3
-  - os:  osx
 
   # Sadly Travis at the moment does not provide as job filter option git changeset, however we emulate it via bash
   # we have two type of tasks to run here
@@ -69,7 +56,6 @@ matrix:
 
 before_install:
 - python -c 'import sys; print(sys.version)'
-- if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then source .travis-osx; fi
 - export BUILD="setup.py pyproject.toml MANIFEST.in setup.cfg tox.ini .travis.yml appveyor.yml"
 - export SRC=$(git diff --name-only $TRAVIS_COMMIT_RANGE -- '*.py' $BUILD | grep -v conf.py | grep -v "tasks/.*.py")
 - '[[ "$TOXENV"!="py" || (-n "$SRC" && "$TOXENV"="py") ]] && export RUN="YES"'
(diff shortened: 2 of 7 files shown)


obestwalter:
Thanks @gaborbernat is there a link you could share to see what is being run and how this looks like? I don't know vsts yet at all and would be interested to compare it to what Travis and Appveyor does.

obestwalter:
Thanks. Will check it out in friday!

gaborbernat:
I'll not be available this weekend (Friday included, but do share your thoughts).

obestwalter:
I spent a litle time with it now but did not have much time as I has wished, so my impressions are still quite superficial:

* I like the UI - it seems more accessible than Travis and Appveyor.
* Pity that it can't cope with `README.rst`
* There is a tempting button to set up continuous delivery, but releasing to PyPI is not supported out of the box. Might be low hanging fruit though to implement

But in general I have no problem with tox being an early adopter and experiencing a few bumps for a while. As Github is now part of Microsoft there is a good chance that the level of integration between GH and CI might significantly improve. 
As long as we can always do manual releases, if necessary I am not really bothered.

Thank you @gaborbernat for exploring this and @zooba for being our "inside man" in this :)

zooba:
> There is a tempting button to set up continuous delivery

It's coming :) I'm on the discussions ([here](https://github.com/Microsoft/vsts-tasks/tree/master/templates/ci) is where the templates live).

> releasing to PyPI...

It works well enough for me. As usual, it's a credentials challenge. There is a [PyPI Publisher](https://github.com/Microsoft/vsts-tasks/tree/master/Tasks/PyPIPublisherV0) task, and it uses a service connection that you would configure while logged in to the site and reference from the YAML (I don't actually know how to spell it in YAML though...), but it also tries to rebuild. One of these days I'll just dive in and revamp it, but there's another guy who has it on his list and it's actually his job, so I'm hoping he'll get to it first ;)

The other way of doing it is to use a secret variable (defined in the UI for the build definition, not in the YAML) and then pass it as an option in the YAML command (e.g. `python -m twine upload -u $(PyPIUsername) -p $(PyPIPassword) ...`). Even if the build definition is public, it _should_ hide the value even in the logs - the one trick is that you can't set environment variables with secret values, you have to pass them as a parameter.

One other way is to use a secure file. Under "Build and release" there's a section called "Library" where you can upload a secure file. This is kept encrypted (I believe in Azure KeyVault, but don't quote me on that) and there's a [Download Secure File](https://docs.microsoft.com/en-us/vsts/pipelines/tasks/utility/download-secure-file?view=vsts) task that will get it at the start of your build and automatically delete it at the end. You can put your `pypirc` file in here and then reference it as part of the build.

(Also look at the "Publish build artifacts" task, which can upload the built files and attach them to the build results page for you to download later - see the little "build" and "test" items [here](https://python.visualstudio.com/cpython/_build/results?buildId=12545) from the Python 3.7.0 build).

I'd love to hear about your experiences if you try any of these approaches. We're currently working on building out specific advice/tutorials/etc. for Python developers, so knowing what works best for you will help.

obestwalter:
Hi @zooba, thanks for the details about the PyPI release. I like the option to use a secure file to make sure the release credentials are reasonably secure.

What I was referring to by "supported out of the box", was something else though. When I was clicking around, I landed on: https://toxdev.visualstudio.com/tox/_releaseDefinition?definitionId=0&path=%5C&_a=action-create-definition where you have many "classic" deployment options (like "Deploy a Python app to Azure App Service ..."), but not e.g. "Release a Python package to PyPI". It would be great for project maintainers if they could "deploy" their build artifacts to well established package repositories (or things that provide the same interface, e.g. a PyPI-like repo provided by devpi or artifactory) just like deploying some app to some cloud.

gaborbernat:
I'm not sure I follow you here @obestwalter. You already have this option by adding a task that does this. We even have one added, see https://github.com/tox-dev/tox/blob/master/.vsts-ci.yml#L198. For now it's triggered whenever you tag a commit. I'm not entirely ready for release every commit model. The project needs to mature a bit I believe.

obestwalter:
Take all this with a grain of salt please. I simply don't know exactly yet how this all plays together. I was giving some feedback to @zooba from the "stupid fresh user" perspective. When I am clicking around in the UI then I might naively expect a deployment step that models the idea of what a deployment might mean for a package maintainer. Because in my view, what deployment means has a lot to do with what your project is doing.

At least for me, deployment can mean different things depending on what you are doing. For us maintainers in e.g. the tox project a "deployment" of a tox build means "upload relevant files to PyPI". Yes I know "publish" or "release" are better words here, but in terms of process the upload to PyPI is the same kind of step as a deploy to some cloud platform for an app in a different project. 

To give an example where this works for me: at work we use Atlassian Bamboo and there we have deployment projects which are connected to build plans. Some of them are strictly concerned with uploading the build and tested package to our in house package repository, some of them deploy an app to AWS, some distribute whatever to in-house servers via SaltStack, etc. Process-wise they all have as the deploy step, the thing that means the updated project is available where it is needed. Same difference whether that means being downloadable as a package from some repo or the code running on AWS. To me this feels right.

obestwalter:
> I'm not entirely ready for release every commit model. The project needs to mature a bit I believe. 

Sure. That's not what I suggested anyway. I think what we have now thanks to you automating the release step is just right.

Issue #909, opened 2018-07-13
'develop-inst-nodeps' activity always runs


stephenfin:
When a project is using `[tox] skipsdist = True` and `[testenv] usedevelop = True`, tox will attempt to determine when it needs to reinstall the package (the `develop-inst-nodeps` activity) and when it does not (the `develop-inst-noop` target). It does this by comparing the last modified timestamps of the `setup.py` or `setup.cfg` files with that of the `.egg-info` directory.

https://github.com/tox-dev/tox/blob/b1e526db570bdb28f4d9a6d51ef1a77fe84b8c43/src/tox/venv.py#L253-L257

Unfortunately, this is a flawed check as, even after running the `develop-inst-nodeps` activity (which executes a variant of `pip install -e $setupdir --no-deps`) the timestamp of the `.egg-info` directory is not updated. This is a particularly irritating issue for larger projects when the install step can take many seconds or even minutes.

The ideal fix would be in setuptools and a patch to this effect has been [submitted](https://github.com/pypa/setuptools/pull/1427). However, it will take time for this to filter down to tox and, IMO, we should fix this ourselves in the interim.

obestwalter:
Hi @stephenfin, I agree. This is a long standing issue and it would be great if someong would find the time to fix this properly. There is whole project collectiong problems in that area, if you are interested: https://github.com/tox-dev/tox/projects/9

gaborbernat:
Fixed via https://github.com/tox-dev/tox/pull/910

PR #910, opened 2018-07-13
Touch '.egg-info' directory after install (#909)

stephenfin:
When a project is using `skipsdist = True` and `usedevelop = True`, tox will attempt to determine when it needs to reinstall the package (the `develop-inst-nodeps` activity) and when it does not (the `develop-inst-noop` target). It does this by comparing the last modified timestamps of the 'setup.py' or 'setup.cfg' files with that of the `.egg-info` directory.

Unfortunately, this is a flawed check as, even after running the `develop-inst-nodeps` activity (which executes a variant of `pip install -e $setupdir --no-deps`) the timestamp of the `.egg-info` directory is not updated. This is a particularly irritating issue for larger projects when the install step can take many seconds or even minutes.

The ideal fix would be in setuptools. However, while we wait on this, we can fix it ourselves by simply `touch`ing the `.egg-info` directory after a reinstall. We do this.

### DIFF
--- a/changelog/909.bugfix.rst
+++ b/changelog/909.bugfix.rst
@@ -1,0 +1,5 @@
+A caching issue that caused the ``develop-inst-nodeps`` action, which
+reinstalls the package under test, to always run has been resolved. The
+``develop-inst-noop`` action, which, as the name suggests, is a no-op, will now
+run unless there are changes to ``setup.py`` or ``setup.cfg`` files that have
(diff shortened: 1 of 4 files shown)


stephenfin:
> @stephenfin in order to consider this can you please link us the upstream issue tracking the mentioned underlying issue? thanks!

Yup, you can find it here https://github.com/pypa/setuptools/pull/1427

PR #933, opened 2018-08-03
release 3.1.3

gaborbernat:
first with vsts :+1:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,38 @@ with advance notice in the **Deprecations** section of releases.
 
 .. towncrier release notes start
 
+3.1.3 (2018-08-03)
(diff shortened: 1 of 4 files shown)


gaborbernat:
There's a slight duplication here. Seems both the tag and the pr ci runs even though they are the same. Moreover, the tag commit is not added as ci check. Plus needed some fixing on the vsts side to trigger build on tags.

gaborbernat:
The release should happen on https://toxdev.visualstudio.com/tox/_build/results?buildId=250&view=logs.

gaborbernat:
Success.

nicoddemus:
Congrats on the first with VSTS! I'm curious, is there a badge for VSTS?

PR #935, opened 2018-08-09
Prefer -m to invoke modules instead of direct executable

gaborbernat:
This helps in some cases to bypass the shebang limitation.

### DIFF
--- a/changelog/xxx.feature.rst
+++ b/changelog/xxx.feature.rst
@@ -1,0 +1,2 @@
+Switch pip invocations to use the module ``-m pip`` instead of direct invocation. This could help
+avoid some of the shebang limitations.  - by :user:`gaborbernat`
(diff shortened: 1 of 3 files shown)


rpkilby:
This PR has a curious side effect. `python -m` inserts the current directory into `sys.path`, meaning that `.egg-info` metadata located in the current directory will now be picked up during the test run ([info](https://github.com/pypa/pip/issues/4621#issuecomment-315123477)). e.g., with a `foo.egg-info`, you get the following results:

```sh
(.venv)$ pip list
Package    Version
---------- -------
pip        18.0   
setuptools 39.0.1 

(.venv)$ python -m pip list
Package    Version
---------- -------
foo        0.0.1
pip        18.0   
setuptools 39.0.1 
```

I don't know if there are any practical effects here. It theoretically affects package installation, but it would be unusual for someone to place a dependency's egg-info in their `tox.ini` directory.

PR #943, opened 2018-08-11
Add Docker tools

amancevice:
- Add Dockerfile with Python 2.7+3.6 interpreters and tox installed
- Add docker-compose configuration to run tests with '-e fix-lint,py27,py36'

Run tests with `docker-compose run --rm tox`.

Consider publishing official Docker image that can be pulled with `docker pull tox`. This would allow users to run tox tests without ever needing to install tox locally. Users can mount their projects to the `/tox` directory inside the image and run tests in a containerized environment (as is done in `docker-compose.yml`)

## Contribution checklist:

- [ ] added/updated test(s)
      in message body
- [ ] added news fragment in [changelog folder](https://github.com/tox-dev/tox/tree/master/changelog)

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,6 +1,7 @@
 Alexander Schepanovski
 Alex Grönholm
 Alexander Loechel
+Alexander Mancevice
 Alexandre Conrad
 Allan Feldman
 Andrii Soldatenko
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM python:3.6
+ARG TOX_VERSION=3.2.1
+ENV TOX_VERSION=${TOX_VERSION}
+RUN apt-get install -y python-dev && \
+    pip3 install tox==${TOX_VERSION}
+VOLUME /tox
+WORKDIR /tox
+ENTRYPOINT ["tox"]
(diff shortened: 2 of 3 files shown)


amancevice:
Using 3.7 is fine -- I was going by the contribution guidelines, which ask to run `tox -e fix-lint,py27,py36`.

As for who's going to test/maintain it, I could remove the pegged dependency from the Dockerfile to be simply `pip install tox`. This isn't _great_ practice in Docker, but official/automated builds on Dockerhub can be set up to produce images on every commit so you wouldn't technically need to do anything to produce new image tags; Dockerhub will just build/host them for you. That would make it pretty hands-free.

As for heavy/light I'd argue that's a bit of a matter of opinion. This isn't meant to be an _additional_ requirement for users, but another option for installation (or "anti-installation"). Some devs prefer to work entirely in containers (many of my coworkers, for example) for a variety of reasons, one of which is not having to worry about polluting their host environments with potentially incompatible installations across projects.

At the end of the day, it's totally up to you. I can always post this image to my personal Dockerhub if it's not welcome here.

gaborbernat:
I'm sort of agreeing we might want to do this; however in a fashion where we can enforce some CI rules. Until we setup docker hub I would be hesitant to accept anything. Furthermore, I would say we don't want to pass in anything under ``tox -e``, we should just let the default envlist kick in.

gaborbernat:
I'll close this now, in favor of https://github.com/tox-dev/tox/issues/1035. I feel we should first discuss how we want to provide and maintain this.

PR #949, opened 2018-08-26
Use 3.7 for version spec instead of 3.7.0-b2

asottile:


asottile:
yikes, looks like a pytest breaking change :S

PR #861, opened 2018-06-29
[wip] Pep 518

gaborbernat:
#850 PEP-518 support: provide a tox configuration flag ``build`` which can be either ``sdist`` or ``wheel``. For ``sdist`` (default) we build the package as before by using ``python setup.py sdist``. However, when ``wheel`` is enabled now we'll use ``pip wheel`` to build it, and we'll also install wheels in these case into the environments. Note: ``pip`` 10 supports specifying project dependencies (such as ``setuptools-scm``, or a given ``setuptools`` version) via ``pyproject.toml``. Once ``pip`` supports building ``sdist`` to we'll migrate over the ``sdist`` build too. ``wheel`` support also enforces to have ``pip 10+`` and a ``pyproject.toml``. 

#851 While running tox invokes various commands (such as building the package, pip installing dependencies and so on), these were printed in case they failed as Python arrays. Changed the representation to a shell command, allowing the users to quickly replicate/debug the failure on their own.

PR #952, opened 2018-08-29
Expose the packaging as a plugin hook

### LINKED ISSUE #951: the initial sdist phase should be plugin-able
This would allow people to customize the build phase while we get PEP-517/518 out of the door via plugins.

gaborbernat:
Resolves #951. This will allow having PEP-517/518 for the end users.

yay


src/tox/session.py
```diff
@@ -564,47 +517,14 @@ def installpkg(self, venv, path):
...
-
     def subcommand_test(self):
         if self.config.skipsdist:
             self.report.info("skipping sdist step")
-            path = None
         else:
-            path = self.get_installpkg_path()
-            if not path:
-                return 2
+            for venv in self.venvlist:
```
gaborbernat: We now allow separate packages per virtual environment. The use case for this extension is to allow e.g. binary packages (might need different compilation) and/or wheels (different wheel per venv).

src/tox/package.py
```diff
@@ -0,0 +1,85 @@
...
+        path = config.option.installpkg
+        if not path:
+            path = config.sdistsrc
+        path = session._resolve_package(path)
+        report.info("using package {!r}, skipping 'sdist' activity ".format(str(path)))
+    else:
+        try:
+            path = make_sdist(report, config, session)
+        except tox.exception.InvocationError:
+            v = sys.exc_info()[1]
```
asottile: (I know this is existing, but I couldn't help notice it) -- shouldn' tthis be `as v:` and not retrieving `sys.exc_info()`?

src/tox/session.py
```diff
@@ -564,47 +517,14 @@ def installpkg(self, venv, path):
...
+                venv.package = self.hook.tox_package(session=self, venv=venv)
```
asottile: Ah ok, this will cause sdist to be called N times (previously 1) -- should we make this use the virtualenv's python as well in this pass?

src/tox/package.py
gaborbernat: Indeed it should :+1:

src/tox/session.py
gaborbernat: That would be a breaking change. This is mostly just reorg to have a clear separation of packaging and plugin-ablity.  So not just yet, maybe with tox 4. For now, the plan is to just keep using the host python to build the venv. 

I'll have a follow-up PR that will create a new virtualenv for building the package, parse the pyproject.toml, install those dependencies. That will effectively implement PEP-517/518 as an opt-in :+1:  As external plugin people can implement wheel support, etc. I'll retract my https://github.com/tox-dev/tox/pull/861 as such.

PR #954, opened 2018-08-30
PEP-517 source distribution support

### LINKED ISSUE #573: Add support for PEP-517

gaborbernat:
create a ``.package`` virtual environment to perform build operations inside

@asottile @nicoddemus @obestwalter feel free to take a look at it and try to find holes in it :+1: base implementation there, just need to add documentation and some more tests :100:

FYI @pfmoore (we probably will switch over pip once pip will have pep-517 support and ``build`` command - that being said I find this level a support a bit more generic e.g. python version selection). With this tox should finally allow users to specify their packaging dependencies, and use other than setuptools.

on it

Resolves #573 and #820.


pfmoore:
Nice! I'll try to find some time to take a proper look at this. (FYI, the initial implementation of PEP 517 in pip won't include a `pip build` command, but `pip wheel` will use PEP 517, I don't know if that's sufficient for you).

asottile:
I'll take a look at this later today -- excited for this!

src/tox/config.py
```diff
@@ -1005,6 +1005,13 @@ def __init__(self, config, inipath):  # noqa
         )
 
         config.skipsdist = reader.getbool("skipsdist", all_develop)
+        config.isolated_build = reader.getbool("isolated_build", False)
+        if config.isolated_build is True:
+            name = ".package"
+            if name not in config.envconfigs:
```
gaborbernat: the build env name should be configurable

src/tox/package.py
```diff
@@ -83,3 +99,113 @@ def make_sdist(report, config, session):
...
+    if not previous_config or not previous_config.matches(live_config, True):
+        session.finishvenv(package_venv)
+
+    build_requires = get_build_requires(build_info, package_venv, session)
+    for requirement in build_requires:
+        pkg_requirement = pkg_resources.Requirement(requirement)
+        if pkg_requirement.key not in toml_require:
+            package_venv.envconfig.deps.append(DepConfig(requirement, None))
+
+    if not session.setupenv(package_venv):
```
gaborbernat: here we are recreating the env maybe instead just install new dependencies and finalize again the env

src/tox/package.py
```diff
@@ -83,3 +99,113 @@ def make_sdist(report, config, session):
...
+
+    live_config = package_venv._getliveconfig()
+    previous_config = CreationConfig.readconfig(package_venv.path_config)
+    if not previous_config or not previous_config.matches(live_config, True):
+        session.finishvenv(package_venv)
+
+    build_requires = get_build_requires(build_info, package_venv, session)
+    for requirement in build_requires:
+        pkg_requirement = pkg_resources.Requirement(requirement)
+        if pkg_requirement.key not in toml_require:
```
gaborbernat: note here we might potentially have a issue that this requirement might differ from the first, maybe check for it and raise if that's the case

src/tox/package.py
```diff
@@ -83,3 +99,113 @@ def make_sdist(report, config, session):
...
+        script = textwrap.dedent(
+            """
+            import sys
+            import {}
+            basename = {}.build_{}("{}", {{ "--global-option": ["--formats=gztar"]}})
+            print(basename)""".format(
+                build_info.backend_module,
+                build_info.backend_object,
+                "sdist",
+                str(config.distdir).replace("\\", "\\\\"),
```
gaborbernat: instead of the replace mark the output build directory as a raw string

tests/integration/test_package_int.py
```diff
@@ -0,0 +1,78 @@
...
+    assert result.ret == 0, result.out
+
+    result2 = cmd("--sdistonly")
+    assert result2.ret == 0, result.out
+    assert ".package recreate" not in result2.out
+
+
+@pytest.mark.network
+@need_git
+@pytest.mark.skipif(sys.version_info < (3, 0), reason="flit is Python 3 only")
```
gaborbernat: in CI instead of skip should raise instead

tests/lib/__init__.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
```
gaborbernat: @asottile @nicodeamus stolen this from pip any better way of doing this?

tox.ini
```diff
@@ -125,8 +125,8 @@ multi_line_output = 3
 include_trailing_comma = True
 force_grid_wrap = 0
 line_length = 99
-known_first_party = tox
-known_third_party = apiclient,git,httplib2,oauth2client,packaging,pkg_resources,pluggy,py,pytest,setuptools,six
+known_first_party = tox,tests
+known_third_party = apiclient,git,httplib2,oauth2client,packaging,pkg_resources,pluggy,py,pytest,setuptools,six,tests,toml
```
gaborbernat: @asottile for some reason tests shows up as third party even though it is marked as first party

src/tox/package.py
```diff
@@ -83,3 +99,113 @@ def make_sdist(report, config, session):
...
+            )
+        )
+        config.distdir.ensure_dir()
+        result = action.popen([package_venv.envconfig.envpython, "-c", script], returnout=True)
+        return config.distdir.join(result.split("\n")[-2])
```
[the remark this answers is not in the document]
gaborbernat: returnout implies universal newlines which means \n only

gaborbernat:
@asottile 

> needs documentation: isolated_build setting, .package environment name
> maybe an example which shows how to configure the executable for the .package environment (and/or other useful things to override there?)

Added some: the config flags plus an example page related to packaging.

> does the .package environment leak into tox --listenvs (or whatever the option is called) -- (either way) should it?
It probably does, we need to filter it out. Will do and add test for it.

> is .package a good enough name to avoid conflicts (probably)? if not maybe .tox-package?
Given that now it's configurable it's good enough I think especially with the ``.`` start to make it private/hidden so to say.

> Will wheel support want to make N .package environments -- does this enable that or make it more difficult?

Wheel support will have to do the ``n`` separate package environments. Going from there maybe we should instead make the isolated_build a choice so people can swap out to other alternatives/options; such as the wheel instead of ``sdist``. I would say the way we have packaging plugabble should be fairly easy to add wheel support later on (no way around it - will have to create up to ``n`` separate build envs though - ``n`` is the number of unique binary type - e.g. 2 if only py2 and py3, but with abi and other parameters can be a lot more).

gaborbernat:
This now is mostly complete. Will merge in on Monday unless we find something significant until then. Will re-read https://www.python.org/dev/peps/pep-0517/ before that to make sure we did not miss anything.

obestwalter:
Better later than never: thank you @gaborbernat for your great work on this :fireworks: :fireworks: :fireworks:

PR #960, opened 2018-09-04
VSTS build improvements

zooba:
Improves the VSTS build configuration in a few minor ways:
* skips codecov upload when CODECOV_TOKEN is not set (makes it easier for other people to run the build in their own account)
* fixes test run title to include OS and Python version
* ensures test results are uploaded even if test run fails (non-zero exit code)
* publish HTML code coverage report

The HTML coverage report is specified in tox.ini, so it'll be created for all builds. If you'd prefer, I can switch it to use `{posargs}` and only enable the report when using it in VSTS. Or I can just leave it out - unfortunately, the embedding UI doesn't look so pretty (images/CSS break), but at least this makes it available for download if the simple % number isn't sufficient and codecov is not available.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -85,8 +85,9 @@ jobs:
     inputs:
       testResultsFiles: '.tox/test-results.*.xml'
       mergeTestResults: true
-      testRunTitle: '$(toxenv)'
+      testRunTitle: '$(agent.os) py$(python.version)'
       platform: linux
+    condition: succeededOrFailed()
 
   - script: 'python -m tox -e coverage'
     displayName: generate coverage.xml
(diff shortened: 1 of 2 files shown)


zooba:
And a minor side not that I have not validated the macOS changes fully yet because there's a major Azure outage that's still affecting that build queue. Given the rest are fine, I assume it'll be okay, but can recheck before merging if you think it's necessary.

zooba:
I've pinged the team about the errors in the VSTS build. Guessing it's because the change is in a fork, and so the authorization isn't flowing through correctly. Will let you know what they say - my own builds (not in a public project, I'm afraid) worked fine.

zooba:
Okay, it seems to be happy about the code coverage publish now :) Must just be a fork thing.

Thanks for merging!

gaborbernat:
thanks for the patch

PR #964, opened 2018-09-11
tests: newconfig: do not load setuptools entrypoints 

### LINKED ISSUE #963: Tests are failing after installing pdbpp in the test tox env
I am used to install [pdbpp](https://github.com/antocuni/pdb) when debugging issues, but after `.tox/py37/bin/pip install ~v/pdbpp` a lot of tests are faling, e.g.

```
________________________________________________________ TestConfigTestEnv.test_take_dependencies_from_other_testenv[envlist1-deps1] _________________________________________________________

self = <test_config.TestConfigTestEnv object at 0x7fca74b06c50>, newconfig = <function create_new_config_file.<locals>.create_new_config_file_ at 0x7fca74acbea0>, envlist = ['py27', 'py34']
deps = ('pytest', 'py{27,34}: pytest-cov')

    @pytest.mark.parametrize(
        "envlist, deps",
        [
            (["py27"], ("pytest", "pytest-cov")),
            (["py27", "py34"], ("pytest", "py{27,34}: pytest-cov")),
        ],
    )
    def test_take_dependencies_from_other_testenv(self, newconfig, envlist, deps):
        inisource = """
                [tox]
                envlist = {envlist}
                [testenv]
                deps={deps}
                [testenv:py27]
                deps=
                    {{[testenv]deps}}
                    fun
                    frob{{env:ENV_VAR:>1.0,<2.0}}
            """.format(
            envlist=",".join(envlist), deps="\n" + "\n".join([" " * 17 + d for d in deps])
        )
        conf = newconfig([], inisource).envconfigs["py27"]
        packages = [dep.name for dep in conf.deps]
>       assert packages == list(deps) + ["fun", "frob>1.0,<2.0"]
E       AssertionError: assert ['pytest', 'p...2.0', 'pdbpp'] == ['pytest', 'py...rob>1.0,<2.0']
E         Left contains more items, first extra item: 'pdbpp'
E         Use -v to get the full diff

conf       = <tox.config.TestenvConfig object at 0x7fca74a18748>
deps       = ('pytest', 'py{27,34}: pytest-cov')
envlist    = ['py27', 'py34']
inisource  = '\n            [tox]\n            envlist = py27,py34\n            [testenv]\n            deps=\n                 pyte...    deps=\n                {[testenv]deps}\n                fun\n                frob{env:ENV_VAR:>1.0,<2.0}\n        '
newconfig  = <function create_new_config_file.<locals>.create_new_config_file_ at 0x7fca74acbea0>
packages   = ['pytest', 'py{27,34}: pytest-cov', 'fun', 'frob>1.0,<2.0', 'pdbpp']
self       = <test_config.TestConfigTestEnv object at 0x7fca74b06c50>

unit/test_config.py:1443: AssertionError
```

Not sure how to address this best.
Since other deps installed there do not appear to cause this issue it might be related to pdbpp's use of `.pth`?

blueyed:
Fixes #963.

### DIFF
--- a/changelog/964.misc.rst
+++ b/changelog/964.misc.rst
@@ -1,0 +1,1 @@
+do not load entrypoint plugins during tests by default with ``newconfig`` fixture - by :user:`blueyed` (`#964 <https://github.com/tox-dev/tox/pull/964>`_)
--- a/src/tox/_pytestplugin.py
+++ b/src/tox/_pytestplugin.py
@@ -49,7 +49,7 @@ def work_in_clean_dir(tmpdir):
 
 @pytest.fixture(name="newconfig")
 def create_new_config_file(tmpdir):
-    def create_new_config_file_(args, source=None, plugins=()):
+    def create_new_config_file_(args, source=None, plugins=(), load_entrypoints=False):
         if source is None:
             source = args
             args = []
@@ -57,7 +57,7 @@ def create_new_config_file(tmpdir):
         p = tmpdir.join("tox.ini")
         p.write(s)
         with tmpdir.as_cwd():
-            return parseconfig(args, plugins=plugins)
+            return parseconfig(args, plugins=plugins, load_entrypoints=load_entrypoints)
 
     return create_new_config_file_
 
--- a/changelog/examples.rst
+++ b/changelog/examples.rst
@@ -15,4 +15,4 @@ file ``571.bugfix.rst``::
     installation type if the package is installed and ``skip_install`` determines if it should be
     installed at all) - by :user:`ferdonline`
 
-.. see tox/pyproject.toml for all available categories
+.. see pyproject.toml for all available categories
--- a/src/tox/config.py
+++ b/src/tox/config.py
@@ -36,8 +36,8 @@ default_factors = tox.PYTHON.DEFAULT_FACTORS
 """DEPRECATED MOVE - please update to new location."""
 
 
-def get_plugin_manager(plugins=()):
-    # initialize plugin manager
+def get_plugin_manager(plugins=(), load_entrypoints=True):
+    """Initialize plugin manager"""
     import tox.venv
 
     pm = pluggy.PluginManager("tox")
@@ -49,7 +49,8 @@ def get_plugin_manager(plugins=()):
     from tox import package
 
     pm.register(package)
-    pm.load_setuptools_entrypoints("tox")
+    if load_entrypoints:
+        pm.load_setuptools_entrypoints("tox")
     for plugin in plugins:
         pm.register(plugin)
     pm.check_pending()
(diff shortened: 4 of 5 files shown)


blueyed:
JFI: "tox ci" appears to not being able to upload coverage: https://toxdev.visualstudio.com/tox/_build/results?buildId=380&view=logs

src/tox/_pytestplugin.py
```diff
@@ -47,17 +47,17 @@ def work_in_clean_dir(tmpdir):
         yield
 
 
-@pytest.fixture(name="newconfig")
-def create_new_config_file(tmpdir):
-    def create_new_config_file_(args, source=None, plugins=()):
+@pytest.fixture
+def newconfig(tmpdir):
+    def create_new_config_file_(args, source=None, plugins=(), load_entrypoints=False):
```
blueyed: Not sure if it should be added here as kwarg.  We could just use `False` below always, and add it if needed later?

gaborbernat: please revert this line, use the name to avoid non pep 8 function names

blueyed: `newconfig`?
Would `new_config` be better then?
(I think it is important for developers to goto the function easily, and ctags does not understand the redirection trhough `name` here)

gaborbernat: pycharm understands and it's up to the other tools to improve their smartness, not up to us to write badly formatted code. the name needs to be newconfig for backwards compatibility.

blueyed: Why is `newconfig` not PEP8 btw?

So you are saying "goto newconfig" in pycharm goes to this fixture?
You agree that it is not really trivial (but I can imagine pycharm has implemented it), since you a) need to be aware of pytest specialities and then b) parse any function decorators for it.

I'll revert it.

blueyed: I've also not exposed load_entrypoints in the fixture's kwargs here now.

> backwards compatibility

Isn't this just used with tox itself?

gaborbernat: it's a pytest plugin shipped with tox, e.g. plugins use it

blueyed: Then we should also default to the old behavior of loading entrypoints, no?

blueyed: Can you also my other questions, please?

blueyed: Also btw:

> it's up to the other tools to improve their smartness

https://github.com/blueyed/tox/blame/a2e4b804c150ac6303e0293f5e4517b2b4580c08/tests/conftest.py#L4

('# noqa` should be enough for all tools there)

gaborbernat: yes we probably should

blueyed:
> Then we should also default to the old behavior of loading entrypoints, no?

IIRC this will require a lot of changes to the code then, and might not be really feasible then.

gaborbernat:
@blueyed so why is this change needed can you elaborate?

blueyed:
I think it is good for isolated tests in the first place, like https://github.com/tox-dev/tox/issues/963 showed (where I am install pdbpp automatically with tox envs, and that causes test failures).

gaborbernat:
can you resolve the conflict?

blueyed:
Done.

changelog/964.misc.rst
```diff
@@ -0,0 +1 @@
+do not load entrypoint plugins during tests by default with ``newconfig`` fixture - by :user:`blueyed` (`#964 <https://github.com/tox-dev/tox/pull/964>`_)
```
gaborbernat: please remove the issue reference, this is automatically added. hence why the city broke

blueyed:
I am closing this.
For the pdbpp tox plugin I am looking at "config.isolated_build" now, which appears to make more sense: https://github.com/antocuni/pdb/pull/98/files#diff-732bb213d07fa5858e8f898f21f197f6R14.

PR #966, opened 2018-09-11
pytest: use -ra by default

blueyed:
Let's see failing tests in the summary, ok?  ("f")

[ci skip]
[build skip]

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -115,7 +115,7 @@ source = src/tox
          .tox/pypy*/site-packages/tox
 
 [pytest]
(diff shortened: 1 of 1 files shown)


blueyed:
How can you skip "tox ci"?

gaborbernat:
you cannot 😎

blueyed:
Too bad (wasteful!).
Especially since it appears to have problems with coverage (only from PRs?): https://github.com/tox-dev/tox/pull/964#issuecomment-420400251.

PR #992, opened 2018-09-17
tox injects env vars to indicate running within tox

### LINKED ISSUE #847: Set TOX_ENVDIR environment variable

gaborbernat:
Resolves #847.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -805,7 +805,6 @@ the following:
       ``mysql-py36``,
     - but not ``py2``, ``py36-sql`` or ``py36-mysql-dev``.
 
-
 Advanced settings
 -----------------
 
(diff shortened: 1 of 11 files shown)


doc/config.rst
```diff
@@ -829,6 +828,19 @@ user can bypass the system's interpreter parser by defining the
...
 
+Injected environment variables
+==============================
+tox will inject the following environment variables that you can use to test that your command is running within tox:
+
+.. versionadded:: 3.4
+
+- ``TOX_WORK_DIR`` env var is set to the tox work directory
+- ``TOX_ENV_NAME`` is set to the current running tox environment name
+- ``TOX_ENV_DIR`` is set to the current tox environments working dir.
```
asottile: I'm still not convinced this is a good idea.  One can already do this with `setenv` today.

It also creates a difference between how tests would run outside / inside tox, and potentially an environment-specific behaviour (which blanking the environment + `passenv` was intended to isolate).

More importantly, this change has the potential to break tests which _expect_ a blank environment.

gaborbernat: True one could do with ``setenv``, however that would be an inconsistent and laborious process. Note we still shield from all user envs, we just inject these three special tox prefixed envs. 

I'm not sure I've ever seen tests expecting blank environments. Note we already pass in the PYTHONHASHSEED (although that's swallowed by Python I assume).

asottile: change seems fine though, I'm just not convinced it's a good idea :laughing:

gaborbernat: ```It also creates a difference between how tests would run outside / inside tox`` - isolating the tox env already creates such discrepancy. I'm slightly in favor of this, and given that @blueyed requested it from user side, and @obestwalter was also in favor let's trial it.

Issue #994, opened 2018-09-17
vsts fails to discover the current python 


gaborbernat:
see https://toxdev.visualstudio.com/tox/_build/results?buildId=454&view=logs

```bash
018-09-17T10:58:19.1541324Z __________________ test_tox_env_var_flags_inserted_isolated ___________________
2018-09-17T10:58:19.1541601Z [gw0] win32 -- Python 3.5.4 d:\a\1\s\.tox\py\scripts\python.exe
2018-09-17T10:58:19.1541733Z 
2018-09-17T10:58:19.1541970Z popen_env_test = <function popen_env_test.<locals>.func at 0x000001913AA75158>
2018-09-17T10:58:19.1543263Z 
2018-09-17T10:58:19.1543416Z     @pytest.mark.network
2018-09-17T10:58:19.1543668Z     def test_tox_env_var_flags_inserted_isolated(popen_env_test):
2018-09-17T10:58:19.1543898Z         cur_py = "py{}".format("".join(str(i) for i in sys.version_info[0:2]))
2018-09-17T10:58:19.1544629Z         res = popen_env_test(cur_py, True)
2018-09-17T10:58:19.1544962Z >       assert_popen_env(res)
2018-09-17T10:58:19.1545114Z 
2018-09-17T10:58:19.1545317Z cur_py     = 'py35'
2018-09-17T10:58:19.1545521Z popen_env_test = <function popen_env_test.<locals>.func at 0x000001913AA75158>
2018-09-17T10:58:19.1545984Z res        = <session.test_session.popen_env_test.<locals>.func.<locals>.IsolatedResult object at 0x000001913C055F28>
2018-09-17T10:58:19.1546142Z 
2018-09-17T10:58:19.1546313Z D:\a\1\s\tests\unit\session\test_session.py:299: 
2018-09-17T10:58:19.1546584Z _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
2018-09-17T10:58:19.1547282Z 
2018-09-17T10:58:19.1547552Z res = <session.test_session.popen_env_test.<locals>.func.<locals>.IsolatedResult object at 0x000001913C055F28>
2018-09-17T10:58:19.1547718Z 
2018-09-17T10:58:19.1547869Z     def assert_popen_env(res):
[...]

PR #995, opened 2018-09-17
fallback to current interpreter if matches at discovery

gaborbernat:
#994

### DIFF
--- a/src/tox/interpreters.py
+++ b/src/tox/interpreters.py
@@ -130,6 +130,8 @@ if not tox.INFO.IS_WIN:
 
     @tox.hookimpl
     def tox_get_python_executable(envconfig):
+        if envconfig.basepython == "python{}.{}".format(*sys.version_info[0:2]):
+            return sys.executable
         return py.path.local.sysfind(envconfig.basepython)
 
 
@@ -141,7 +143,7 @@ else:
         p = py.path.local.sysfind(name)
         if p:
             return p
-        actual = None
+
         # Is this a standard PythonX.Y name?
         m = re.match(r"python(\d)(?:\.(\d))?", name)
         groups = [g for g in m.groups() if g] if m else []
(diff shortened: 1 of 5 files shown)


gaborbernat:
@asottile here it seems tox is not able to find pyton3.5 when running under python3.5 :open_mouth:

asottile:
wow that seems pretty special -- that's windows if I'm reading correctly -- is the `py` launcher configured correctly? what happens with `py -3.5`

gaborbernat:
```bash
2018-09-17T20:14:15.5777113Z py -3.5 -c "import sys; print(sys.executable)"
2018-09-17T20:14:15.6419862Z ##[command]"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\15c4da66-e82e-4a75-ab7d-6d7f0803f389.cmd""
2018-09-17T20:14:16.2417178Z Requested Python version (3.5) not installed
2018-09-17T20:14:16.4035016Z ##[error]Cmd.exe exited with code '103'.
```
fails with exit code 103 :thinking: definitely an Azure bug here.

asottile:
whew!

DavidStaheli:
Strange!  We'll look into this right now.  Thanks for the heads-up.

gaborbernat:
@asottile the deliverable on our end I believe is in such cases before giving up we should check if the tox host executable matches up with the requested version, if so we should use it.

src/tox/interpreters.py
```diff
@@ -130,6 +130,8 @@ def __str__(self):
 
     @tox.hookimpl
     def tox_get_python_executable(envconfig):
+        if envconfig.basepython == "python{}.{}".format(*sys.version_info[0:2]):
```
asottile: `[0:2]` is just `[:2]`

PR #996, opened 2018-09-17
Do not publish test results to vsts

gaborbernat:
Takes a long time and we seldom check it.


blueyed:
Let's remove it from the tox command then also?

gaborbernat:
Where at?

blueyed:
https://github.com/tox-dev/tox/pull/997

Issue #851, opened 2018-06-21
commands executed by tox should be correctly quoted


gaborbernat:
Following the success of https://github.com/tox-dev/tox/issues/845, we should make tox ran command displays from Python arrays to commands too. This should allow users to be able to easily debug without tox if one of those fail, by simply copy-pasting into the shell.

PR #1000, opened 2018-09-18
Add pre and post commands

### LINKED ISSUE #167: Have pre and post test run in config
New summary:

Plugins can already run pre and post test commands via https://tox.readthedocs.io/en/latest/plugins.html#tox.hookspecs.tox_runtest_post and https://tox.readthedocs.io/en/latest/plugins.html#tox.hookspecs.tox_runtest_pre. We should allow users to add this inside their config file too, so they don't have to create tox plugins for one time behaviours (e.g. setup/cleanup operations of dependencies).

### LINKED ISSUE #1003: PEP-517 packaging fails with sdist already exists on version change
This has two bugs:
- upstream at https://github.com/pypa/setuptools/pull/1481,
- we need to ensure the sdist folder we request the build to happen exists but is empty.

### LINKED ISSUE #1004: motivate the reason for recreating the virtual environment
Often times I've look puzzled why a virtualenv is recreated. Whenever we decide to re-create an existing virtual environment we should provide motivation as a level two verbosity.

gaborbernat:
Resolves #167.
Resolves #1003.
Resolves #1004.
Fixup for #987 and #851.

PS. Once we resolve and merge this we'll release ``3.4.0``.

### DIFF
--- a/src/tox/_pytestplugin.py
+++ b/src/tox/_pytestplugin.py
@@ -69,7 +69,7 @@ def cmd(request, capfd, monkeypatch):
     request.addfinalizer(py.path.local().chdir)
 
     def run(*argv):
-        key = str(b"PYTHONPATH")
+        key = b"PYTHONPATH" if six.PY2 else "PYTHONPATH"
         python_paths = (i for i in (str(os.getcwd()), os.getenv(key)) if i)
         monkeypatch.setenv(key, os.pathsep.join(python_paths))
 
@@ -403,33 +403,94 @@ def mock_venv(monkeypatch):
     Note: because we inherit, to keep things sane you must call the py environment and only that;
     and cannot install any packages. """
 
+    # first ensure we have a clean python path
+    monkeypatch.delenv("PYTHONPATH", raising=False)
+
+    # object to collect some data during the execution
+    class Result(object):
+        def __init__(self):
+            self.popens = []
+            self.cwd = None
+            self.session = None
+
+    res = Result()
+
+    # convince tox that the current running virtual environment is already the env we would create
     class ProxyCurrentPython:
         @classmethod
         def readconfig(cls, path):
-            assert path.dirname.endswith("{}py".format(os.sep))
-            return CreationConfig(
-                md5=getdigest(sys.executable),
-                python=sys.executable,
-                version=tox.__version__,
-                sitepackages=False,
-                usedevelop=False,
-                deps=[],
-                alwayscopy=False,
-            )
+            if path.dirname.endswith("{}py".format(os.sep)):
+                return CreationConfig(
+                    md5=getdigest(sys.executable),
+                    python=sys.executable,
+                    version=tox.__version__,
+                    sitepackages=False,
+                    usedevelop=False,
+                    deps=[],
+                    alwayscopy=False,
+                )
+            elif path.dirname.endswith("{}.package".format(os.sep)):
+                return CreationConfig(
+                    md5=getdigest(sys.executable),
+                    python=sys.executable,
+                    version=tox.__version__,
+                    sitepackages=False,
+                    usedevelop=False,
+                    deps=[(getdigest(""), "setuptools >= 35.0.2"), (getdigest(""), "wheel")],
+                    alwayscopy=False,
+                )
+            assert False  # pragma: no cover
 
     monkeypatch.setattr(CreationConfig, "readconfig", ProxyCurrentPython.readconfig)
 
+    # provide as Python the current python executable
     def venv_lookup(venv, name):
         assert name == "python"
+        venv.envconfig.envdir = py.path.local(sys.executable).join("..", "..")
         return sys.executable
 
     monkeypatch.setattr(VirtualEnv, "_venv_lookup", venv_lookup)
 
+    # don't allow overriding the tox config data for the host Python
+    def finish_venv(self):
+        return
+
+    monkeypatch.setattr(VirtualEnv, "finish", finish_venv)
+
+    # we lie that it's an environment with no packages in it
     @tox.hookimpl
     def tox_runenvreport(venv, action):
         return []
 
     monkeypatch.setattr(venv, "tox_runenvreport", tox_runenvreport)
+
+    # intercept the build session to save it and we intercept the popen invocations
+    prev_build = tox.session.build_session
+
+    def build_session(config):
+        res.session = prev_build(config)
+        res._popen = res.session.popen
+        monkeypatch.setattr(res.session, "popen", popen)
+        return res.session
+
+    monkeypatch.setattr(tox.session, "build_session", build_session)
+
+    # collect all popen calls
+    def popen(cmd, **kwargs):
+        # we don't want to perform installation of new packages, just replace with an always ok cmd
+        if "pip" in cmd and "install" in cmd:
+            cmd = ["python", "-c", "print({!r})".format(cmd)]
+        activity_id = res.session._actions[-1].id
+        activity_name = res.session._actions[-1].activity
+        try:
+            ret = res._popen(cmd, **kwargs)
+        except tox.exception.InvocationError as exception:  # pragma: no cover
+            ret = exception  # pragma: no cover
+        finally:
+            res.popens.append((activity_id, activity_name, kwargs.get("env"), ret, cmd))
+        return ret
+
+    return res
 
 
 @pytest.fixture(scope="session")
--- a/src/tox/_pytestplugin.py
+++ b/src/tox/_pytestplugin.py
@@ -69,8 +69,8 @@ def cmd(request, capfd, monkeypatch):
     request.addfinalizer(py.path.local().chdir)
 
     def run(*argv):
-        key = b"PYTHONPATH" if six.PY2 else "PYTHONPATH"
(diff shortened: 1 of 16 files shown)


src/tox/venv.py
```diff
@@ -532,6 +542,31 @@ def tox_runtest(venv, redirect):
     return True  # Return non-None to indicate plugin has completed
 
 
+@tox.hookimpl
+def tox_runtest_pre(venv):
```
asottile: I think these should be separate hooks for maximum compatibility with previous plugins implementing these (otherwise this is an added responsibility that every existing plugin that implements `tox_runtest_pre` now needs to also do!)

gaborbernat: The only added responsibility we're implying here is that the ``venv.status`` should be still zero after ``tox_runtest_pre``. I don't think anyone modified that until now :thinking: So what other responsibilities am I missing? (would prefer not to add new hooks, to pollute our hook namespace even more).

asottile: ah ok, these aren't `firstresult=True` (they maybe should be? hard to say) nevermind this then

gaborbernat: only the commands are first result :+1: the setup and teardown (pre and post are not). And makes sense as while we don't want users to change the command evaluation; we do want to allow them to setup/teardown their plugin.

asottile:
jfc codecov SETTLE DOWN :laughing:

gaborbernat:
@asottile While trying to write tests for this and making them pass discovered two feature request that helped me understand what's happening
- reasoning for venv recreation,
- only update-ing ``.tox.config1`` when it changes, not always;

And two user affecting bugs:
- PEP-517 packaging
- and command quoting being incorrect.

And two internal bugs (thing happened to work by chance, but not by design):
- python md5 being zero when there were dependencies because of variable re-usage;
- and pragma no cover being ignored.

As such this PR became a sort of longer one. However due to the iterative nature of these would be troublesome to split it up. As such I would appreciate to review and merge it as one.

src/tox/_pytestplugin.py
```diff
@@ -69,7 +69,7 @@ def cmd(request, capfd, monkeypatch):
     request.addfinalizer(py.path.local().chdir)
 
     def run(*argv):
-        key = str(b"PYTHONPATH")
+        key = b"PYTHONPATH" if six.PY2 else "PYTHONPATH"
```
asottile: "native" string can just be `str("PYTHONPATH")` -- though as written is just as explicit and correct

src/tox/_pytestplugin.py
```diff
@@ -69,7 +69,7 @@ def cmd(request, capfd, monkeypatch):
...
         python_paths = (i for i in (str(os.getcwd()), os.getenv(key)) if i)
```
asottile: hmm, I can't think of a situation where `getcwd()` returns non-native string -- I suspect the `str(...)` call here is superfluous

src/tox/_pytestplugin.py
```diff
@@ -403,34 +403,95 @@ def mock_venv(monkeypatch):
...
+    def popen(cmd, **kwargs):
+        # we don't want to perform installation of new packages, just replace with an always ok cmd
+        if "pip" in cmd and "install" in cmd:
+            cmd = ["python", "-c", "print({!r})".format(cmd)]
+        activity_id = res.session._actions[-1].id
+        activity_name = res.session._actions[-1].activity
+        try:
+            ret = res._popen(cmd, **kwargs)
+        except tox.exception.InvocationError as exception:  # pragma: no cover
+            ret = exception  # pragma: no cover
```
asottile: p3: this no cover comment shouldn't be necessary since the `except` block itself is no-covered

src/tox/package.py
```diff
@@ -183,7 +183,12 @@ def perform_isolated_build(build_info, package_venv, session, config):
                 build_info.backend_module, build_info.backend_object, "sdist", str(config.distdir)
             )
         )
+
+        # need to start with an empty (but existing) source distribution folder
+        if config.distdir.exists():
+            config.distdir.remove(rec=1, ignore_errors=True)
```
asottile: potential race here (exists check finishes, other process creates), though I suspect it doesn't matter -- though could this just be `config.distdir.remove(rec=1, ignore_errors=True)`?

src/tox/session.py
```diff
@@ -290,13 +290,12 @@ def verbosity(self):
...
-    def logpopen(self, popen, env):
+    def logpopen(self, popen, cmd_args_shell):
         """ log information about the action.popen() created process. """
-        cmd = " ".join(map(str, popen.args))
         if popen.outpath:
-            self.verbosity1("  {}$ {} >{}".format(popen.cwd, cmd, popen.outpath))
+            self.verbosity1("  {}$ {} >{}".format(popen.cwd, cmd_args_shell, popen.outpath))
         else:
-            self.verbosity1("  {}$ {} ".format(popen.cwd, cmd))
+            self.verbosity1("  {}$ {} ".format(popen.cwd, cmd_args_shell))
```
asottile: oh nice, this is the shlex fix in the output?

src/tox/venv.py
```diff
@@ -35,13 +35,13 @@ def writeconfig(self, path):
             lines.append("{} {}".format(*dep))
         path.ensure()
         path.write("\n".join(lines))
+        return lines
```
asottile: (perhaps I'm not seeing it) is this return value used?

src/tox/venv.py
```diff
@@ -50,25 +50,32 @@ def readconfig(cls, path):
...
-            and self.sitepackages == other.sitepackages
-            and self.usedevelop == other.usedevelop
-            and self.alwayscopy == other.alwayscopy
-            and (
-                all(d in self.deps for d in other.deps)
-                if deps_matches_subset is True
-                else self.deps == other.deps
-            )
-        )
+    def matches(self, other, deps_matches_subset=False, provide_reason=False):
```
asottile: this function (if annotated) would be `Union[Tuple[bool, str], bool]` -- does it make sense to instead have two functions: one without the reason and one with the reason?

src/tox/venv.py
```diff
@@ -226,20 +239,32 @@ def matching_platform(self):
...
     def _needs_reinstall(self, setupdir, action):
         setup_py = setupdir.join("setup.py")
         setup_cfg = setupdir.join("setup.cfg")
         args = [self.envconfig.envpython, str(setup_py), "--name"]
         env = self._get_os_environ()
         output = action.popen(args, cwd=setupdir, redirect=False, returnout=True, env=env)
-        name = output.strip()
-        args = [self.envconfig.envpython, "-c", "import sys; print(sys.path)"]
+        name = next(
+            (i for i in output.split("\n") if i and not i.startswith("pydev debugger:")), ""
```
asottile: is this left over from a debugging session?

tox.ini
```diff
@@ -107,7 +107,6 @@ branch = true
 [coverage:report]
 skip_covered = True
 show_missing = True
-exclude_lines = if __name__ == ["']__main__["']:
```
asottile: ah yeah here's what I usually put here:

https://github.com/asottile/all-repos/blob/756c89aa5513235a6f152ddefad311a38470974f/.coveragerc#L15-L28

src/tox/package.py
gaborbernat: that was my first solution, until the test started failing; seems rec=1 is only active if the folder exists because of ``https://github.com/pytest-dev/py/blob/master/py/_path/local.py#L204``. Long story short you can only delete if already exists...

src/tox/session.py
gaborbernat: yeah should be 👍

src/tox/venv.py
gaborbernat: indeed not needed 👍

src/tox/venv.py
gaborbernat: I would like to avoid duplication of the logic though.

src/tox/venv.py
gaborbernat: More like a small fixup 👍 Ideally people should be able to run the test suite in debug mode too with pydev (both eclipse and Pycharm). For that to work this is needed.

src/tox/venv.py
asottile: ah yeah you can just do:

```python
def whatever_with_reason(...):
    return True, 'reason'

def whatever_without_reason(...):
    ret, _ = whatever_with_reason(**kwargs)
    return ret
```

or something like that :man_shrugging:

gaborbernat: I see, my idea was that only makes the code more type correct; which is not really needed until we implement type annotations (probably once we drop Python 2 and 3.4).

asottile: ah yeah I guess I didn't put my justifcation, my bad!

It would also be nice so we don't have a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) parameter -- especially until there's enforced keyword-only-arguments in python3 they're a pretty easy pit to fall into in python where you accidentally pass too many positional arguments and it splats over a defaulted keyword argument

gaborbernat:
Made PR feedback; not all should be ok.

PR #1014, opened 2018-09-22
tests: pytest: use testpaths

blueyed:
Removes then unnecessary norecursedirs.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -132,7 +132,7 @@ source = src/tox
 addopts = -ra --showlocals
 rsyncdirs = tests tox
 looponfailroots = tox tests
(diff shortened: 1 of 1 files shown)


nicoddemus:
They are not quite the same thing, because `norecursedirs` will ignore directories even if you pass a target in the command-line (say `pytest .`), while `testpaths` is the default target when no targets are given.

Having said that I think the change is fine. 😁

blueyed:
Good point.
We could keep `norecursedir` then, but I am also OK with it as-is, too.

blueyed:
The default for `norecursedirs` includes `.*` though, so removing it is fine.

PR #1017, opened 2018-09-25
add --nodeps command line option

janhn:
Kindly asking for a review of suggested changes to implement #410.

Tests have been run as suggested in the contributing guide, with and without the change to verify that no failures are being introduced.

```
=========================== short test summary info ============================
SKIP [1] tests/integration/test_package_int.py:38: flit is Python 3 only
SKIP [1] tests/unit/test_interpreters.py:30: non posix test
XFAIL unit/test_config.py::TestConfigTestEnv::()::test_regression_test_issue_706[envlist0]
  reproduce bug 706
XFAIL unit/test_config.py::TestCmdInvocation::()::test_colon_symbol_in_directory_name
  Upstream bug. See #203

============== 452 passed, 2 skipped, 2 xfailed in 100.44 seconds ==============
```

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -162,7 +162,11 @@ Here is a simple overview, with tox-specific bits:
 
    Or to only run tests in a particular test module on Python 3.6::
 
-    $ tox -e py36 -- testing/test_config.py
+    $ tox -e py36 -- tests/unit/test_config.py
+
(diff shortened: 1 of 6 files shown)


gaborbernat:
Hello @janhn, thanks for the contribution. That being said I don't think this is the right direction to go, so I'll close this down now. Can you pick up the discussion and your use case under #410.

Issue #1024, opened 2018-09-28
move to coverall


gaborbernat:
codecov behaves strangely (fail to publish, incorrect reports, etc)

gaborbernat:
Resolved via #1025, see https://codeclimate.com/github/tox-dev/tox/code?sort=test_coverage.

PR #1025, opened 2018-09-28
azure pipeline restructure and migration to codeclimate

gaborbernat:
:+1: Resolves https://github.com/tox-dev/tox/issues/1024.


obestwalter:
Looks good. Something still missing?

gaborbernat:
Failing Windows tests on Python 2.7 (seems someone somewhere sets a unicode PYTHONPATH environ, and that what breaks the world; trying to identify it) and coverall notifications.

src/tox/_pytestplugin.py
```diff
@@ -404,7 +404,7 @@ def mock_venv(monkeypatch):
     and cannot install any packages. """
 
     # first ensure we have a clean python path
-    monkeypatch.delenv("PYTHONPATH", raising=False)
+    monkeypatch.delenv(str("PYTHONPATH"), raising=False)
```
gaborbernat: @nicoddemus @RonnyPfannschmidt @asottile slight bug here inside pytest monkeypatch. The key passed in here is used to re-insert the value at teardown, instead of the key that was there. Because during the ``get`` unicode keys degrade to bytes keys this can mean that by invoking this operation we transform bytes key to unicode inside ``os.environ`` when the key passed into delenv is unicode. 

In such cases, python 2 Windows will refuse to run subprocesses leaving the user to wonder who and where and how it set non bytes value into ``os.environ``.

gaborbernat:
@obestwalter seems merge-ing the coverages by different jobs is not yet supported out of box on non-Travis CI...

gaborbernat:
Hit production? Seems not supported yet, and no plans to add support.

gaborbernat:
Turns out coveralls does not support merge reports from Azure DevOps. Will need to do somehow a manual merge and upload that.

src/tox/_pytestplugin.py
nicoddemus: Thanks for the ping: https://github.com/pytest-dev/pytest/pull/4056

obestwalter:
Just because they are not talking about it yet, does not mean it could be supported at some point not too far in the future. But o.k. let's try coveralls, ignore my approval in the other PR then :)

gaborbernat:
Eh, feels like we got hit by:
> Yes publishing artifacts from fork builds is currently blocked as we did not have a way to given a sufficiently scoped token to the agent. This is something we are working on fixing but it is going to be a few weeks.

https://github.com/Linuxbrew/brew/issues/746#issuecomment-422955300

anthrotype:
Just curious, why the switch from codecov to coveralls?

anthrotype:
just saw your comment https://github.com/tox-dev/tox/pull/1028#issuecomment-425679584

> Seems codecov does not support Azure DevOps

Pity

gaborbernat:
@anthrotype furthermore codecov turned out to be really flaky for us in the last few months (many unexplainable results) and we got no support when reaching out to them. ``coveralls` also does not implicitly supports it but at least their API is somehow easier to integrate at some level.

gaborbernat:
Closing in favour of https://github.com/tox-dev/tox/pull/1031; I've kept this option only to validate that PR-ing from a fork still works correctly (although no global coverage report). Once Azure Pipelines offers artifact uploads for forks (ETA around 3 weeks), we'll enable coverage generation on all branches. Note in case Azure Pipelines would not deliver this we can migrate to storing the coverage artifacts on Amazon S3 (for a very small cost probably).

PR #1028, opened 2018-09-29
try to improve codecov

gaborbernat:
If we can't use coveralls; we might need to try to fix it. Here it just retries if fails.

### DIFF
--- a/azure-run-tox-env.yml
+++ b/azure-run-tox-env.yml
@@ -32,5 +32,14 @@ jobs:
     - script: 'python -m tox -e coverage'
       displayName: generate coverage report
 
-    - script: 'python -m tox -e codecov -- -t $(CODECOV_TOKEN) --required -n "$(agent.os)-$(python.version)" --build "$(Build.DefinitionName)" --env OS=$(agent.os) pyt+hon=$(python.version)'
+    - script: |
+        for i in 1 2 3; do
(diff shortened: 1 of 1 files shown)


gaborbernat:
Seems codecov does not support Azure DevOps. oh well codecov it is then.

obestwalter:
if-at-first-you-dont-succeed-try-try-and-try-again

obestwalter:
oh ... what? I thought coveralls does not play nicely yet with Azute DevOps. So both are not really there yet? Classic early adopter problems I guess.

azure-run-tox-env.yml
```diff
@@ -32,5 +32,14 @@ jobs:
     - script: 'python -m tox -e coverage'
       displayName: generate coverage report
 
-    - script: 'python -m tox -e codecov -- -t $(CODECOV_TOKEN) --required -n "$(agent.os)-$(python.version)" --build "$(Build.DefinitionName)" --env OS=$(agent.os) pyt+hon=$(python.version)'
+    - script: |
+        for i in 1 2 3; do
```
thedrow: You can use `travis_retry` instead of the shell script

gaborbernat: Note there's no plan to use travis also. This is Azure DevOps, not Travis CI.

PR #1031, opened 2018-10-02
CI - better coverage report, pypy on Azure, add codeclimate

gaborbernat:
This now is done. It's a major rewrite and was significantly more troublesome than expected. After all the effort put in, failed to convince either codecov or coveralls to function correctly. So where two does not work, let's use the third option: codeclimate (https://codeclimate.com/github/tox-dev/tox/builds). 

This one finally gives up on trying to be too smart (e.g. forcing users to merge reports locally during the build), but work correctly what it does offer:
- great documentation,
- Github App,
- great chrome extension to annotate PRs inline,
- Github comment support.

Does not have branch coverage for now but I don't think we used that much often that feature. With this, we also moved off Travis completely as now ``pypy`` tests are also done on Azure Pipelines.

PS. NOTE: we no longer really on a service to report/display code coverage. The CI itself collects all coverage reports, merges them and then displays them as a final step. Codeclimate is just an additional UI to this data, and the coverage gets uploaded to that service too; however, it's also available as part of the CI build log (both full coverage and diff coverage).


gaborbernat:
@obestwalter @asottile  Looks promising. Now it just remains to merge files across machines. Ability to upload artifacts from PRs is predicted for the end of the month, so in the meantime, we'll only get coverage reports if we PR branches of this repo; but not forks. Before merge we just need to configure coveralls to report coverage results as a comment.

gaborbernat:
This now is done. It's a major rewrite and was significantly more troublesome than expected. After all the effort put in, failed to convince either codecov or coveralls to function correctly. So where two does not work, let's use the third option: codeclimate (https://codeclimate.com/github/tox-dev/tox/builds). 

This one finally gives up on trying to be too smart (e.g. forcing users to merge reports locally during the build), but work correctly what it does offer:
- great documentation,
- Github App,
- great chrome extension to annotate PRs inline,
- Github comment support.

Does not have branch coverage for now but I don't think we used that much often that feature. With this, we also moved off Travis completely as now ``pypy`` tests are also done on Azure Pipelines.

obestwalter:
hi @gaborbernat sorry I missed that (like many other things at the end of last year). I am reading through what happened in reverse chronological order at the moment and am completely stumped by the amazing work that you have put in :bowing_man: :ok_hand:

PR #1032, opened 2018-10-03
Defend against parallel package builds via a file lock

### LINKED ISSUE #1026: Intermittent failures when using --parallel--safe-build with parallel builds
When using `--parallel--safe-build` option in Jenkins as described in the [documentation](https://tox.readthedocs.io/en/latest/example/jenkins.html#running-tox-environments-in-parallel) sometimes one of the builds fails when tox is generating the sdist.

It seems that the sdist is being generated in the source directory which is shared across the builds.

```
running sdist
running egg_info
writing <redacted>.egg-info/PKG-INFO
writing dependency_links to <redacted>.egg-info/dependency_links.txt
writing requirements to <redacted>.egg-info/requires.txt
writing top-level names to <redacted>.egg-info/top_level.txt
reading manifest file '<redacted>.egg-info/SOURCES.txt'
writing manifest file '<redacted>.egg-info/SOURCES.txt'
running check
warning: check: missing required meta-data: url

copying files to <redacted>-0.3.0...
copying setup.py -> <redacted>-0.3.0
copying <redacted>.egg-info/PKG-INFO -> <redacted>-0.3.0/<redacted>.egg-info
error: could not create '<redacted>-0.3.0/<redacted>.egg-info/PKG-INFO': No such file or directory
```

OS: RHEL 6.8 amd64
`pip list` output:
```
Package        Version
-------------- -------
<redacted>  0.3.0
pip            18.0
pluggy         0.7.1
py             1.6.0
setuptools     39.2.0
six            1.11.0
toml           0.9.6
tox            3.4.0
virtualenv     16.0.0
```

gaborbernat:
Resolves #1026.

### DIFF
--- a/src/tox/config.py
+++ b/src/tox/config.py
@@ -1024,6 +1024,7 @@ class ParseIni(object):
 
         reader.addsubstitutions(distdir=config.distdir)
         config.distshare = reader.getpath("distshare", dist_share_default)
+        config.temp_dir = reader.getpath("temp_dir", "{toxworkdir}/.tmp")
         reader.addsubstitutions(distshare=config.distshare)
         config.sdistsrc = reader.getpath("sdistsrc", None)
         config.setupdir = reader.getpath("setupdir", "{toxinidir}")
--- a/src/tox/config.py
+++ b/src/tox/config.py
@@ -8,7 +8,6 @@ import re
 import shlex
 import string
 import sys
-import uuid
 import warnings
 from collections import OrderedDict
 from fnmatch import fnmatchcase
@@ -419,7 +418,8 @@ def tox_addoption(parser):
         "--parallel--safe-build",
         action="store_true",
         dest="parallel_safe_build",
-        help="ensure two tox builds can run in parallel",
+        help="(deprecated) ensure two tox builds can run in parallel "
+        "(uses a lock file in the tox workdir with .lock extension)",
     )
     parser.add_argument(
         "--installpkg",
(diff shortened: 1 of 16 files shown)


src/tox/package.py
```diff
@@ -27,6 +28,22 @@ def get_package(session):
...
+        session.config.toxworkdir.join("{}.lock".format(session.config.isolated_build_env))
+    )
+    lock = FileLock(lock_file)
+    try:
+        try:
+            lock.acquire(0.0001)
+        except Timeout:
+            report.verbosity0("lock file {} present, will block until released".format(lock_file))
+            lock.acquire()
+        return acquire_package(config, report, session)
```
gaborbernat: The problem now (caught by Python2.7 builds) is that downstream operations (such as pip install) on this file will fail if a subsequent run removes it to regenerate it (and the current session still wants to use it). We need to defend against this by probably making a temporary copy of this and feed that downstream. Another option would be to move it to that temp folder and create a symlink instead of copy (and when subsequent builds clean it follow up on symlinks and delete the target too); we would still need to feed back the temporary name of it. @asottile opinions?

asottile: maybe I'm missing it, but couldn't the removal / creation just be moved into the critical section protected by the lock?

pre-commit has to do a similar thing: https://github.com/pre-commit/pre-commit/blob/c0b1f2ff25b53efcfa03098f2b4cb90ded905063/pre_commit/store.py#L79-L90

asottile: should it make a dist for each environment then?

gaborbernat: For each session more likely. We need to keep around dist folder to not break peoples workflow in parallel.

src/tox/config.py
```diff
@@ -1018,20 +1018,17 @@ def __init__(self, config, ini_path, ini_data):  # noqa
...
-        unique_id = str(uuid.uuid4())
 
         reader.addsubstitutions(toxworkdir=config.toxworkdir)
         config.distdir = reader.getpath("distdir", "{toxworkdir}/dist")
-        self._make_thread_safe_path(config, "distdir", unique_id)
 
         reader.addsubstitutions(distdir=config.distdir)
         config.distshare = reader.getpath("distshare", dist_share_default)
-        self._make_thread_safe_path(config, "distshare", unique_id)
+        config.temp_dir = reader.getpath("temp_dir", "{toxworkdir}/.tmp")
```
webknjaz, line 1027: @gaborbernat if I'm reading this right, you forgot to add it to substitutions

gaborbernat: This is meant as an internal implementation detail folder, not meant for public usage.

webknjaz: Ah, the docs suggest otherwise...

gaborbernat: Well, I guess now it's supported.

webknjaz: Hyrum's Law, huh?

PR #1033, opened 2018-10-05
document cli

### LINKED ISSUE #829: Improve --notest documentation
- [x] a clear problem statement with an example

When running tox in CI there's often job step separation and failures in different kinds of stages are accounted differently. For example, in Travis CI its error/failure depending on whether the command is failing in install of script step.

- [x] suggested change with example

I'd like to see smth like `--skip-command` flag or smth like it for pre-populating env (installing deps), which I'd run in the `install` stage, and then I'd run normal tests, for which tox would reuse env created earlier.
This would help separate different logical steps of doing automatic testing in CI.

gaborbernat:
Resolves #829.

URL of RTD document: https://tox.readthedocs.io/en/829/


webknjaz:
How does it improve `--notest` doc?

PR #1034, opened 2018-10-05
Move PyPi checks to twine

### LINKED ISSUE #1023: move to twine check
see https://github.com/pypa/readme_renderer/issues/118

gaborbernat:
Resolves #1023.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,9 +70,6 @@ def main():
             "console_scripts": ["tox=tox:cmdline", "tox-quickstart=tox._quickstart:main"]
         },
         python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
-        setup_requires=[
-            "setuptools-scm"
-        ],  # needed for https://github.com/pypa/readme_renderer/issues/118
         install_requires=[
             "setuptools  >= 30.0.0",
             "pluggy >= 0.3.0, <1",
--- a/tox.ini
+++ b/tox.ini
@@ -40,10 +40,12 @@ commands = sphinx-build -d "{toxworkdir}/docs_doctree" doc "{toxworkdir}/docs_ou
 [testenv:package_description]
 description = check that the long description is valid
 basepython = python3.7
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -70,9 +70,6 @@ def main():
             "console_scripts": ["tox=tox:cmdline", "tox-quickstart=tox._quickstart:main"]
         },
         python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
-        setup_requires=[
-            "setuptools-scm"
```
anthrotype: If you remove setuptools-scm from setup_requires the setup.py won’t work with pip versions that don’t understand the pyproject.toml build requirements, no?

gaborbernat: Not in this case, as setuptools-scm now is a source distribution to sdist dependency. It does not get used when we build from a source distribution a wheel, or when we install that. If you look into the output the sdist command you'll see the generated PKG-INFO already contains the version info. Furthermore, in our code base at no point do we use setuptools-scm (instead we generate the version.py https://github.com/tox-dev/tox/blob/master/setup.py#L52 at sdist time, and refer to that). 

```bash
Metadata-Version: 2.1
Name: tox
Version: 3.4.1.dev21+gf2d6c71
Summary: virtualenv-based automation of test activities
Home-page: https://tox.readthedocs.org/
Author: holger krekel
Author-email: holger@merlinux.eu
License: https://opensource.org/licenses/MIT
```

anthrotype:
I see. But then pip installing from a local git checkout will depend on a relatively recent version of pip, that can bootstrap the build requirements.

gaborbernat:
Installing tox HEAD (aka a git checkout) with an old pip? Seems a niche enough to ignore it. The only officially supported method of installing tox is via pip and PyPi. And for that setuptools-scm is not needed.

anthrotype:
Also python setup.py install would probably break, but I presume easy_install is de facto deprecated these days.
I’m interested because I may do the same in other projects of mine (remove setuptools-scm from setup_requires).

PR #1051, opened 2018-10-10
Bugfix: Do the same transformation to egg-info dirs that pkg_resources does

hashbrowncipher:
I noticed that my hyphenated project name is not using the `develop-inst-noop` action, instead using the `develop-inst-nodeps` (cf. #910). This is because it uses a hyphenated project name, and tox was looking at the mtime of a nonexistent `myproject-name.egg-info` directory, when distutils/setuptools instead create a directory called `myproject_name.egg-info`. I don't know why they behave this way.

I've added a test for this behavior, and also made it so that initproj can take hyphenated strings for project names. Then, to fix the failing test, I added a call to `to_filename`, the function which `pkg_resources` uses for the same transformation.


gaborbernat:
Interesting that this did not generate a CI run, will need to investigate. Will follow up tomorrow. Thanks for the PR.

gaborbernat:
Thanks!

Issue #1061, opened 2018-10-17
Small typo in docs/example/basic.rst


metasyn:
`contrains` should be `constraints`. Just a single word typo :)

crazymerlyn:
This issue was fixed by #1062.

PR #1062, opened 2018-10-17
Fix small typo where the word contraints was spelled as contrains

metasyn:
Single word typo  - very small PR.


metasyn:
Fixes #1061

gaborbernat:
thanks for the pr

PR #1066, opened 2018-10-19
fix #1047 and better structure for packaging

### LINKED ISSUE #1047: show packaging output
We should show packaging output when three level verbosity is passed in. This would help often with tracking down packaging issues. Three levels, as we reserve the first two as tox levels, two plus is verbosity for elements called by tox.

gaborbernat:
👍


gaborbernat:
🤦‍♂️ when you work two hours trying to do this...and you succeed, but you named some folders ``build`` which is ignored by git and you lose all... 🤦‍♂️

asottile:
:yikes:

PR #1068, opened 2018-10-20
Don't capture stderr when doing `setup.py --name` (closes #1052)

### LINKED ISSUE #1052: Bug: _needs_reinstall captures `python setup.py --name` stderr in addition to stdout
I ran into this bug because my `setup.py --name` uses the warnings module, which prints to stderr. As a result, tox was unable to [infer the location](https://github.com/tox-dev/tox/blob/8944dc1491c3bd74db957e6cdff29a005abc19c4/src/tox/venv.py#L282) of my `egg_info` directory.

hashbrowncipher:
As suggested in #1052, this prints the warning and continues with the build.


hashbrowncipher:
To answer the question from #1051, it appears that a CI run is not generated when I force push to the branch.

PR #1069, opened 2018-10-21
Split up packaging for better separation of concerns

gaborbernat:
👍


.gitignore
```diff
@@ -1,34 +1,28 @@
...
-*.rej
-*~
 __pycache__
 
+
+# packaging folders
 src/tox/version.py
+build/
+dist/
+src/tox.egg-info
```
asottile: another tip so you don't lose your files again, you can prefix these with a `/` to just cover the root of the repository [for instance](https://github.com/asottile/pyupgrade/blob/1db939202cdafbd1c144e165157cd4fffc8520c9/.gitignore#L3-L6)

src/tox/package/__init__.py
```diff
@@ -0,0 +1,68 @@
+import py
```
asottile: personally I dislike code in `__init__.py` -- hard to find, easy to make circular imports among other things.  not blocking though

nicoddemus: > personally I dislike code in __init__.py -- hard to find, easy to make circular imports among other things. not blocking though

👍, for the same reasons hehehe

gaborbernat: I'm not sure in this case. Would lead to some common.py/base.py files instead. Or package.py inside package module?

.gitignore
gaborbernat: Hmm, I thought to put the / at end will have the same effect. Not?

gaborbernat: And yes this is only moving code around.

asottile: nope!

```console
$ tree .
.
├── build
│   └── ignore_me
└── t
    └── build
        └── dont_ignore_me

3 directories, 2 files
$ cat .gitignore 
build/
$ git status
On branch master

No commits yet

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.gitignore

nothing added to commit but untracked files present (use "git add" to track)
```

PR #1071, opened 2018-10-24
Fix `commands_pre`/`commands_post` documentation.

benoit-pierre:
Mention the minimum version required.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -9,6 +9,7 @@ Anthony Sottile
 Asmund Grammeltwedt
 Barry Warsaw
 Bartolome Sanchez Salado
+Benoit Pierre
 Bernat Gabor
 Bruno Oliveira
 Carl Meyer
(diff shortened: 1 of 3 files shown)


benoit-pierre:
Should be fixed.

gaborbernat:
Thanks!

cjerdonek:
Has there been any consideration of having tox issue a warning if unrecognized key values are encountered? That could help people to know they could need to upgrade. Currently, such key values (e.g. `commands_pre`) are silently skipped if the user doesn't have a new enough version.

gaborbernat:
No, and probably we'll never have it. This would be backwards incompatible change. Note people can at the moment use empty keywords together with substitution to re-use fields, e.g.:

```
[testenv]
doc_python = python3

[testenv:doc_build]
basepython = {[testenv]doc_python}

[testenv:doc_build2]
basepython = {[testenv]doc_python}

``` 

If you depend on a feature that's only available under a given version you should specify it via the minversion https://tox.readthedocs.io/en/latest/config.html#conf-minversion

PR #1080, opened 2018-11-04
improve our own devenv in tox.ini

obestwalter:
* fix devenv != recommendation in our own documentation
* remove forced Python3.7 base (not necessary)
* remove ancient comment that makes only sense in a limited context
* improve description (contains path to devenv now)

see: https://tox.readthedocs.io/en/latest/example/devenv.html

[//]: # (rtdbot-start)

URL of RTD document: https://tox.readthedocs.io/en/improve-devenv/ Documentation Status

[//]: # (rtdbot-end)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -165,14 +165,12 @@ deps = gitpython >= 2.1.10
        oauth2client >= 4.1.2
 commands = python {toxinidir}/tasks/notify.py
 
-[testenv:dev]
-description = generate a DEV environment
+[testenv:devenv]
+description = create '{toxworkdir}/devenv' with all dependencies for development
 extras = testing, docs
 deps = {[testenv]deps}
        {[testenv:release]deps}
        {[testenv:notify]deps}
-# required to make looponfail reload on every source code change
 usedevelop = True
-basepython = python3.7
 commands = python -m pip list --format=columns
-           python -c 'import sys; print(sys.executable)'
+           python -c 'import sys; print("devenv at: %s" % sys.exec_prefix)'
--- a/tox.ini
+++ b/tox.ini
@@ -165,12 +165,12 @@ deps = gitpython >= 2.1.10
        oauth2client >= 4.1.2
 commands = python {toxinidir}/tasks/notify.py
 
-[testenv:devenv]
-description = create '{toxworkdir}/devenv' with all dependencies for development
+[testenv:dev]
+description = dev environment with all deps at {envdir}
 extras = testing, docs
 deps = {[testenv]deps}
        {[testenv:release]deps}
        {[testenv:notify]deps}
 usedevelop = True
 commands = python -m pip list --format=columns
-           python -c 'import sys; print("devenv at: %s" % sys.exec_prefix)'
+           python -c "print('{envpython}')"
--- a/docs/example/devenv.rst
+++ b/docs/example/devenv.rst
@@ -38,7 +38,7 @@ In it we state:
   develop``, as opposed to building and installing its source distribution using
   ``setup.py install``.
 
-The development environment will reside in `toxworkdir` (default is `.tox`) just
+The development environment will reside in ``toxworkdir`` (default is ``.tox``) just
 like the other tox environments.
 
 We can configure a lot more, if we want to. For example, we can add the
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -165,14 +165,12 @@ deps = gitpython >= 2.1.10
        oauth2client >= 4.1.2
 commands = python {toxinidir}/tasks/notify.py
 
-[testenv:dev]
-description = generate a DEV environment
+[testenv:devenv]
```
gaborbernat: I would be more inclined to update the documentation and make it dev instead.

obestwalter: I prefer devenv.

Aareon: I personally believe the name `devenv` is a bit redundant and non-standard in this case. No other environments are suffixed with `env` and it seems to me to be a little pedantic. I'd vouch for using `dev` in this case, unless there is any other reason besides personal preference.

tox.ini
```diff
@@ -165,14 +165,12 @@ deps = gitpython >= 2.1.10
...
 extras = testing, docs
 deps = {[testenv]deps}
        {[testenv:release]deps}
        {[testenv:notify]deps}
-# required to make looponfail reload on every source code change
 usedevelop = True
-basepython = python3.7
 commands = python -m pip list --format=columns
-           python -c 'import sys; print(sys.executable)'
+           python -c 'import sys; print("devenv at: %s" % sys.exec_prefix)'
```
gaborbernat: We don't use the ``%`` operator but prefer format instead. That being said paired with the fact that I also find env redundant I'm going to close this. Also having the python executable directly I think it's better because that's what you need to provide to the IDE as the interpreter to use.

obestwalter:
I rather have this fixed properly.

tox.ini
obestwalter: It's personal preferenve against personal preference here, I guess :)

obestwalter:
To summarize: there is disagreement about the name ("devenv" vs "dev") and about the printed user hint.

About the name: I think the name "devenv" expressed what this is without needing extra context if used standalone and this is why I think it is the better name. "dev" is too nondescript and overused (just like the name "tox" itself, but we're stuck with that, I guess :laughing:).

About the path: I think that the folder of the virtualenv is the information that matters in general and just printing the path without extra text is just a bit too terse. 

@gaborbernat said:
> Also having the python executable directly I think it's better because that's what you need to provide to the IDE as the interpreter to use.

One could also argue that it would be better to print the complete path to the activation script as "that's what you need to provide to the command line to activate the environment" (which I don't do but why I think that simply pointing out where the virtualenv is, is "universally" interesting to all kinds of potential contributors, wether they use an IDE or vim or whatever).

obestwalter:
Sure. My hope is still that maybe a few more opinions from contributors would trickle in to tip the scales one way or another. It's neither an important nor an urgent change, so this might as well hang a bit longer and mature a bit. Maybe someone comes up with a third option that is better.

Thinking about it: looking from the perspective of what the prompt looks like when the environment is activated it might even be more helpful, if the name of the development environment contains actually important info rather than a generic name - so how about recommending in the docs something like `[testenv:<project>-dev]` which then would translate into a prompt like 

```console
(tox-dev) ~ $ pytest
```

As someone who switches around a lot between many projects I would like this, if that would turn into an established practice. It is also still easy to script this to activate on changing into the project directory (and to activate either `<project>-dev` or `dev[env]`).

gaborbernat:
I still hardly disagree that at any point users should be activating tox virtual environments. This is just bound to leave us in a place where they inadvertently modify it without going through tox. As such I would like to keep the dev environment name as short as possible and universal: hence ``dev`` my proposal. No need to make it add a namespace of the project.

gaborbernat:
The only change here the change of environment name on which it seems to not be any agreement or willingness from other people to churn in. As such I'm inclined to close this for now.

obestwalter:
I don't understand what the rush is about.

obestwalter:
The [linting failures](https://toxdev.visualstudio.com/tox/_build/results?buildId=1003) are about things that my changes didn't touch but just happened to be in the same file. 

I would expect that we lint everything once and then the whole code base complies to e.g. a newly added "rst-backticks" check rather than hitting me with it now who didn't touch this part of the code.

obestwalter:
Linting errors were on me after all. I fixed my setup now and installed the pre-commit hook :)

gaborbernat:
Looks great, thanks!

PR #1092, opened 2018-11-26
Don't warn when PYTHONPATH is present but empty.

Julian:
Some users (e.g. me :) have an always set [but empty] PYTHONPATH
(for reasons like supporting zsh's typeset -T). Since Python's own
behavior is the same in this case as having the env var unset,
tox shouldn't warn.

## Contribution checklist:

- [x] added/updated test(s)
      in message body
- [x] added news fragment in [changelog folder](https://github.com/tox-dev/tox/tree/master/docs/changelog)

### DIFF
--- a/src/tox/venv.py
+++ b/src/tox/venv.py
@@ -387,7 +387,7 @@ class VirtualEnv(object):
             os.environ.pop(key, None)
         if "PYTHONPATH" not in self.envconfig.passenv:
             # If PYTHONPATH not explicitly asked for, remove it.
-            if "PYTHONPATH" in os.environ:
+            if os.environ.get("PYTHONPATH"):
                 self.session.report.warning(
                     "Discarding $PYTHONPATH from environment, to override "
                     "specify PYTHONPATH in 'passenv' in your configuration."
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -36,6 +36,7 @@ Johannes Christ
 Jon Dufresne
 Josh Smeaton
 Josh Snyder
+Julian Berman
 Julian Krause
 Jurko Gospodnetić
 Krisztian Fekete
--- a/docs/changelog/1092.misc.rst
+++ b/docs/changelog/1092.misc.rst
@@ -1,0 +1,1 @@
+Don't warn when PYTHONPATH is present but empty - by :user:`Julian`.
--- a/tests/unit/test_venv.py
+++ b/tests/unit/test_venv.py
@@ -630,6 +630,26 @@ class TestVenvTest:
         assert os.environ["PIP_NO_DEPS"] == "0"
 
     def test_pythonpath_usage(self, newmocksession, monkeypatch):
+        # Empty PYTHONPATH doesn't warn
+        monkeypatch.setenv("PYTHONPATH", "")
+        mocksession = newmocksession(
+            [],
+            """
+            [testenv:python]
+            commands=abc
+        """,
+        )
+        venv = mocksession.getenv("python")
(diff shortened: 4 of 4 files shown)


src/tox/venv.py
```diff
@@ -387,7 +387,7 @@ def ensure_pip_os_environ_ok(self):
             os.environ.pop(key, None)
         if "PYTHONPATH" not in self.envconfig.passenv:
             # If PYTHONPATH not explicitly asked for, remove it.
-            if "PYTHONPATH" in os.environ:
+            if os.environ.get("PYTHONPATH"):
```
RonnyPfannschmidt: note that this is a behavior change allowing the empty variable to pass trough

Julian: Yeah I considered that (and popping it out) but couldn't think of a decent reason why it'd matter.

Happy to change if someone does though.

gaborbernat:
zsh typeset support? feels to me like you're trying to fix a zsh bug in tox.

Julian:
It's not a zsh bug, I was giving some background on how I found this issue
(because I happen to have the variable always set).

Judge the patch on its own merit.

gaborbernat:
Without context, I'm tempted to reject this. Setting the env var as empty is likely to confuse users so warning on it seems a best practice (even if python silently swallows this kind of odd usage).

Julian:
Sorry, what? Which users is this confusing? A user has set PYTHONPATH to be empty. This is valid, has documented behavior equivalent to not having it set at all. I gave a reason why someone might be doing this, but equally well a user could be writing `export PYTHONPATH=` somewhere earlier in a terminal session and forgotten about it.

All this does is silence a warning for the case that it isn't needed. The purpose of the warning is to let the user know that their environment variable is being ignored (which by the way tox isn't even consistent about, but fine). In this case, it's not ignored, they're getting the same behavior, so don't pollute their stdout with red messages.

Probably should have written all that elaboration earlier I guess, but if you still disagree it's really not worth pursuing such a small fix so feel free to close.

gaborbernat:
Can you please link to ``This is valid, has documented behavior equivalent to not having it set at all.``? 👍

src/tox/venv.py
blueyed: Well, at least it will make the env larger, and therefore I think it should not be added in the first place - it is nicer to have less in `env` etc.

asottile: let's err on the side of caution and remove it anyway.  can dedent line 395 and change it to `os.environ.pop('PYTHONPATH', None)`

blueyed:
It appears the behavior for an empty PYTHONPATH changed in Python 3.4 for the `python` interpreter, previously it behaved like PYTHONPATH=. according to https://github.com/blueyed/cpython/blame/a346266c87c8f4bd2343bea1bbacd357d2ae435f/Doc/whatsnew/3.4.rst#L2249-L2255.  Therefore it would make sense to have the warning with earlier versions there still I suppose?!

Julian:
Sorry, lucky me caught the flu and has been completely out of commission for the last week. Will have to catch up with whatever comments here.

gaborbernat:
@Julian do you plan to finish this?

Julian:
Seemingly unlikely at the minute unfortunately, too many other things. Will close and send a new PR if I do get a moment to get back to it.

gaborbernat:
Let's keep it open then, I might finish it.

Aareon:
Excellent! Thank you, @gaborbernat 😊

gaborbernat:
superseded with https://github.com/tox-dev/tox/pull/1189

PR #1093, opened 2018-11-27
Link to related projects (Invoke and Nox) from the documentation

### LINKED ISSUE #1088: Consider linking to Nox & other related projects
Hey foiks,
I'm the maintainer of [Nox](http://nox.thea.codes/en/stable/), a tool that is similar to Tox in purpose but different in spirit. One of the things we found to be really useful was to tell our users about [other projects](http://nox.thea.codes/en/stable/#other-useful-projects) that they might be interested in looking at if they're considering our using Nox.

I'm wondering if Tox would be interested in likewise linking to Nox (& possibly PyInvoke). I think our projects could benefit from more cross-linking. For example, it seems that some of the feature requests here are already possible with Nox's approach (and vice-versa). We could point our users to each other in those cases so that they can get unblocked quickly and we can better prioritize requests.

I'm more than happy to write a pull request to add this to the docs, I just wanted to reach out and make sure that this would be a welcome addition. If not, that's totally fine. :)

Thanks!

theacodes:
Closes #1088

### DIFF
--- a/docs/changelog/1088.doc.rst
+++ b/docs/changelog/1088.doc.rst
@@ -1,0 +1,1 @@
+Link to related projects (Invoke and Nox) from the documentation - by :user:`theacodes`.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -160,6 +160,17 @@ Current features
 
 * supports :ref:`using different / multiple PyPI index servers  <multiindex>`
 
+
+Related projects
+----------------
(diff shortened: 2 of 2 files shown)


theacodes:
Thank you so much, @gaborbernat!

gaborbernat:
thank you for the pr

PR #1102, opened 2018-12-16
parallel invocation of tox environments #439

### LINKED ISSUE #439: Allow parallel execution of envs (merge or replace detox?)
This is more of an idea that would simplify parallel execution of tests using tox. Why not merging detox into tox?

I do see several benefits from this:
* better visibility
* simplify project management, they already share the same issue tracker
* easier to deploy
* more trust: detox project has last update from 2015 on pypi making people believe it may be abandonware, also the document page contain outdated links to bitbucket.
* once done, pypi admins can replace detox project with an alias to tox (they already did this for other projects).

As part of the proposed consolidation we can even trigger the parallel option via `-P` command line option.

gaborbernat:
This solution is ``subprocess`` based. We're already heavily using subprocesses, so I consider this implementation safe. 

TBD: 
- for non-live parallel show some progress notification,
- correctly forward all status-es and their messages from parallel invocations,
- better tests.

Resolves #439.


michaelmanganiello-eb:
Hi @gaborbernat !

I'm currently working on adding support to `pytest-django` to rename Django databases to avoid name collisions (https://github.com/pytest-dev/pytest-django/pull/680).

For this scenario at least, it would be great to have a public-facing way to check if Tox is being running in parallel or not. The `_PARALLEL_TOX` environment variable seems to achieve this, but I think it shouldn't be part of a private API that could be easily changed.

gaborbernat:
Sure we can add some environment variable, but I believe shouldn't two parallel pytest invocations work always, even if not made through tox, aka parallel shell invocations?

adamantike:
I'm not sure we would like to rename DBs by default, even in projects that won't run in parallel or have other requirements/limitations for test DB generation. For two parallel `pytest` invocations to always work, we should append the associated PID or any random number to DB names, something that could be not very straightforward.

However, to support scenarios where `tox --parallel` is executed, and also scenarios where the orchestation is implemented manually (e.g. `for` loop running `tox -vve {tox_env}` asynchronously), my idea is to only use the value of the (current) `_PARALLEL_TOX` variable just as a boolean, and, when that's set, change the DB name based on the content of the `TOX_ENV_NAME` environment variable.

That way, even if developers aren't using `tox --parallel`, they could change their own execution method, to use `_PARALLEL_TOX=1 tox ...` instead.

gaborbernat:
I'm not sure I follow why always adding the PID is a bad thing🤔

obestwalter:
> ... it would be great to have a public-facing way to check if Tox is running in parallel or not. The _PARALLEL_TOX environment variable seems to achieve this, but I think it shouldn't be part of a private API that could be easily changed.

I also think that this could be useful as a public facing env var. 

About the name: I think it would be good to always have `TOX_` as the prefix for all tox related ones.

src/tox/config.py
```diff
@@ -413,6 +415,15 @@ def tox_addoption(parser):
         dest="sdistonly",
         help="only perform the sdist packaging activity.",
     )
+    parser.add_argument(
+        "--parallel", action="store_true", dest="parallel", help="run tox environments in parallel"
+    )
+    parser.add_argument(
+        "--parallel-live",
```
obestwalter: Wouldn't it be better to use the verbosity level here? Something like:

* 0 no output from parallel run
* 1 simple progress bar (basically just showing that the whole thing is not stuck)
* 2+ life logs with growing verbosity (error -> warning -> info -> debug)

src/tox/config.py
```diff
@@ -1132,6 +1144,7 @@ def make_envconfig(self, name, section, subs, config, replace=True):
 
     def _getenvdata(self, reader, config):
         candidates = (
+            os.environ.get("_PARALLEL_TOXENV"),
```
obestwalter: Like already mentioned: this can be public but should have `TOX_` prefix then.

gaborbernat:
Thinking this through I now actually want to implement some throttling similar to how pytest-xdist has. ``--parallel-count`` flag would be a choice of {all,auto,count}. ``all`` would be the default but users could switch to auto (CPU count), or explicitly specify a number if they wish.

src/tox/session.py
```diff
@@ -592,9 +600,49 @@ def subcommand_test(self):
...
+        tox_runs = {}
+        for venv in self.venvlist:
+            env = os.environ.copy()
+            env["_PARALLEL_TOXENV"] = venv.envconfig.envname
+            args_sub = list(args)
+            if hasattr(venv, "package"):
+                args_sub.insert(position, str(venv.package))
+                args_sub.insert(position, "--installpkg")
+            stdout, stderr = (None, None) if live_out else (subprocess.PIPE, subprocess.PIPE)
+            run = subprocess.Popen(args_sub, env=env, stdout=stdout, stderr=stderr)
```
obestwalter: If it's only one target anyway: how about simplifying to

```python
sink = None if live_out else subprocess.PIPE
run = subprocess.Popen(args_sub, env=env, stdout=sink, stderr=sink)
```

?

src/tox/session.py
```diff
@@ -592,9 +600,49 @@ def subcommand_test(self):
...
+            tox_runs[venv.name] = (venv, run)
+
+        to_finish = set(tox_runs.keys())
```
obestwalter: `to_finish` reads a bit strage to me here. How about `not_done` or `still_running`?

src/tox/session.py
```diff
@@ -592,9 +600,49 @@ def subcommand_test(self):
...
+        while to_finish:
+            for name in set(to_finish):
+                venv, run = tox_runs[name]
+                res = run.poll()
+                if res is not None:
+                    venv.status = "skipped tests" if self.config.option.notest else res
+                    if not live_out:
+                        out, err = run.communicate()
+                        venv.out = out
+                        venv.err = err
```
obestwalter: simplify to:

```python
venv.out, venv.err = run.communicate()
```

src/tox/config.py
gaborbernat: the user already has TOXENV if they want public, this is solely an internal implementation detail and with a special meaning 🤔

obestwalter:
Really nice and simple way to "merge" detox into tox, I think.

How about also creating an entry point `detox` that internally calls `tox --parallel`? This would also be another reason to not need an extra flag for log levels but using verbosity levels, like mentioned in one of the code comments.

src/tox/session.py
gaborbernat: maybe ``not_done`` as some might still be waiting to be queued with the queuing logic 👍

obestwalter:
I tried running tox parallel now locally with tox itself and have the following output:

```
tox --parallel

.package create: /home/ob/tmp/tox/.tox/.package
.package installdeps: setuptools >= 40.0.4, setuptools_scm >= 2.0.0, <4, wheel >= 0.29.0
___________ summary ____________
  py27: commands succeeded
  py34: commands succeeded
  py35: commands succeeded
  py36: commands succeeded
  py37: commands succeeded
ERROR:   pypy: 1
  pypy3: commands succeeded
ERROR:   coverage: 1
  fix_lint: commands succeeded
  docs: commands succeeded
  package_description: commands succeeded
```

I can't find anything in the logs about the source of the error either :( this might well be a problem with tox itself. I have to admit I hardly ever look into the logs, so I don't even know exactly what I should expect.


Would it be feasible to always capture the output and print it out for all failed envs?

gaborbernat:
This should already happen will need to check why not did in your case.

src/tox/config.py
obestwalter: I meant that all env vars to do with tox should use the prefix `TOX_` so in this case then `TOX_PARALLEL` or something like that.

obestwalter:
> This should already happen will need to check why not did in your case.

Can you post an example output what it looks like when it works?

src/tox/config.py
adamantike: I don't think this should only be considered an implementation detail, as external resources (databases, Redis access...) must be configured accordingly to avoid collisions when toxenvs run in parallel, and we are not just dealing with proper isolated unit tests.

gaborbernat: I'm still on the opinion that the only safe implementation for databases/Redis access is always parallel, rather than a parallel and a non-parallel way. I would expect these two to have the same effect and work equally well: 

- ''tox -e py27,py37 --parallel``,
- ''bash -c 'tox -e py27' &; bash -c 'tox -e py37' &``.

src/tox/config/parallel.py
```diff
@@ -0,0 +1,74 @@
+from __future__ import absolute_import, unicode_literals
+
+from argparse import ArgumentTypeError
+
+ENV_VAR_KEY = "TOX_PARALLEL_ENV"
+OFF_VALUE = 0
+DEFAULT_PARALLEL = OFF_VALUE
+
+
+def auto_detect_cpus():
```
asottile: some prior art that's worth comparing:

- [seems it was copied from here and has drifted](https://github.com/pytest-dev/pytest-xdist/blob/dd63d0b6674be0372e0e47b452c96912a53f0ad3/xdist/plugin.py#L7-L27)
- [less cases, still the travis workaround](https://github.com/pre-commit/pre-commit/blob/46ae88c1f06b8dc59f3e049c68b2dcea22bb9a54/pre_commit/languages/helpers.py#L59-L70)

asottile: notably: I can't see a case where `os.cpu_count` exists but `os.sched_getaffinity` does not 🤷‍♂️

obestwalter:
Just tried out `detox -vvv` on a slow connection and it got stuck. Keyboard interrupt after ~25 minutes yielded:

```
[...]
Writing tox-3.5.1.dev56+g27a5012/setup.cfg
Creating tar archive
removing 'tox-3.5.1.dev56+g27a5012' (and everything under it)
tox-3.5.1.dev56+g27a5012.tar.gz

.package finish: perform-isolated-build after 0.48 seconds
copying new sdistfile to '/home/ob/.tox/distshare/tox-3.5.1.dev56+g27a5012.tar.gz'
package .tmp/package/1/tox-3.5.1.dev56+g27a5012.tar.gz links to dist/tox-3.5.1.dev56+g27a5012.tar.gz (/home/ob/tmp/tox/.tox)
⠙ [10] py27 | py34 | py35 | py36 | py37 | pypy | pypy3 | fix_lint | docs | package_desc⠹ [10] py27 | py34 | py35 | py36 | py37 | pypy | pypy3 | fix_lint | docs | package_desc⠸ [10] py27 | py34 | py35 | py36 | py37 | pypy | pypy3 | fix_lint | docs | package_desc⠼ [10] py27 | py34 | py35 | py36 | py37 | pypy | pypy3 | fix_lint | docs | package_desc⠴ [10] py27 | py34 | py35 | py36 | py37 | pypy | pypy3 | fix_lint | docs | package_desc⠦ [10] py27 | py34 | py35 | py36 | py37 | pypy | pypy3 | fix_lint | docs | package_desc✔ OK pypy3 in 0.669 second
✔ OK fix_lint in 25.623 seconds
✔ OK package_description in 31.974 seconds
_______________________________________ summary _______________________________________
cleanup /home/ob/tmp/tox/.tox/.tmp/package/1/tox-3.5.1.dev56+g27a5012.tar.gz
  File "/home/ob/tmp/tox/.tox/dev/bin/detox", line 11, in <module>
    sys.exit(run())
  File "/home/ob/tmp/tox/.tox/dev/lib/python3.7/site-packages/detox/__init__.py", line 9, in run
    cmdline()
  File "/home/ob/tmp/tox/.tox/dev/lib/python3.7/site-packages/tox/session.py", line 48, in cmdline
    main(args)
  File "/home/ob/tmp/tox/.tox/dev/lib/python3.7/site-packages/tox/session.py", line 55, in main
    retcode = build_session(config).runcommand()
  File "/home/ob/tmp/tox/.tox/dev/lib/python3.7/site-packages/tox/session.py", line 470, in runcommand
    return self.subcommand_test()
  File "/home/ob/tmp/tox/.tox/dev/lib/python3.7/site-packages/tox/session.py", line 604, in subcommand_test
    retcode = self._summary()
  File "/home/ob/tmp/tox/.tox/dev/lib/python3.7/site-packages/tox/session.py", line 748, in _summary
    status = venv.status
AttributeError: 'VirtualEnv' object has no attribute 'status'
^CException ignored in: <module 'threading' from '/usr/lib64/python3.7/threading.py'>
  File "/usr/lib64/python3.7/threading.py", line 1273, in _shutdown
    t.join()
  File "/usr/lib64/python3.7/threading.py", line 1032, in join
    self._wait_for_tstate_lock()
  File "/usr/lib64/python3.7/threading.py", line 1048, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
detox -vvv  37.77s user 5.23s system 2% cpu 25:23.56 total
```

I will have a closer look tomorrow.

I am also unsure how we can phase out the existing detox. If someone installs detox with a tox version that implements parallel execution there should be some kind of warning/error.

gaborbernat:
@obestwalter I recommend trying the latest master, and if it hangs use ``detox -ovv``. Only two level of verbosity is required, the third goes to pip.

tests/unit/util/test_graph.py
```diff
@@ -0,0 +1,5 @@
+from tox.util.graph import stable_topological_sort
+
+
+def test_topological_order():
```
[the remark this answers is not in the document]
gaborbernat: Hehe for sure, the main reason this is not merged is that I need to add more tests, not just here in some other places too.

src/tox/util/spinner.py
```diff
@@ -0,0 +1,156 @@
...
+
+        self._stop_spinner = threading.Event()
+        self._spinner_thread = threading.Thread(target=self.render)
+        self._spinner_thread.setDaemon(True)
+        self._spinner_id = self._spinner_thread.name
+        self._spinner_thread.start()
+        return self
+
+    def succeed(self, key):
+        self.finalize(key, "✔ OK")
```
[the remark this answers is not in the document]
gaborbernat: I assume just fine as the CI passes 😄 I tested from within cmder too on Windows and displays just fine (similar to how black emojis doo). So for now I don't plan adding some fallback symbols, although we may need down the line.

src/tox/session.py
```diff
@@ -593,9 +618,103 @@ def subcommand_test(self):
...
+                        self.report.logline_if(Verbosity.QUIET, message)
+
+            threads = []
+            todo_keys = set(self.venv_order)
+            todo = OrderedDict(
+                (i, todo_keys & set(self.getvenv(i).envconfig.depends)) for i in self.venv_order
+            )
+            done = set()
+            while todo:
+                for name, depends in list(todo.items()):
```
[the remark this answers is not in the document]
gaborbernat: Would be the same though not? I mean once the queue gets an element finished I'm back to enqueuing new, I actually feel this implementation is cleaner.

tests/unit/util/test_graph.py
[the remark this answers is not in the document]
gaborbernat: I've been using for my projects locally in last week and seems to work fine otherwise (Mac/Linux).

src/tox/util/spinner.py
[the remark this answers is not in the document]
asottile: Heh, black's don't render for me on windows so I suspect these might suffer as well

src/tox/session.py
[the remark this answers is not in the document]
nicoddemus: No, you would just fill the queue with all the environments before launching the threads. Each thread would then consume one item of the queue and call `Queue.task_done()` on it. In the main thread you can wait for all items to be processed by calling `Queue.join()`.

src/tox/util/spinner.py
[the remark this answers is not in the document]
nicoddemus: On `Cmder` it will work fine (which I use myself too), but on `cmd` things might be different... let me test this real quick.

tests/unit/util/test_graph.py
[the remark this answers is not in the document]
nicoddemus: Ahh OK! 😁

src/tox/util/spinner.py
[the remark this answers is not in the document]
nicoddemus: image

😢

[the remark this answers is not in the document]
nicoddemus: On Cmder it looks fine 😁 

image

Can we make use colors for the symbols? It would improve the output quite a bit I believe

src/detox/__init__.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import, unicode_literals
+
+from tox import cmdline
+from tox.config import parallel
+
+
+def run():
+    parallel.DEFAULT_PARALLEL = "all"
```
nicoddemus: Perhaps the default should be `auto`? I just run `detox` on `pytest` and it almost froze my computer when it launched 17 environments at the same time.

src/tox/session.py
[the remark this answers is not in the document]
nicoddemus: Here's a nice example:

https://docs.python.org/3/library/queue.html#queue.Queue.join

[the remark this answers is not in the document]
gaborbernat: Thanks, I'll have a think about it.

src/tox/util/spinner.py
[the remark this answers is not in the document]
gaborbernat: Sure any colour suggestions? Is there a reliable way to detect if the terminal supports these unicode characters, so we can fallback to something ASCII?

src/detox/__init__.py
gaborbernat: 😅 oh my, yeah it maxes out CPU quite nicely. For tox itself all works ok, but probably because it's dominated by network and disk operations the test suite. Will ammend.

src/tox/util/spinner.py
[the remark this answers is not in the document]
nicoddemus: > Sure any colour suggestions?

Green for "success" and red for "failure". The environment names might also appear in blue for easier viewing.

> Is there a reliable way to detect if the terminal supports these unicode characters, so we can fallback to something ASCII?

On pytest we use `colorama` to handle the colors, I assumed `tox` was using the same thing.

src/detox/__init__.py
gaborbernat: On the other side I'm usually running mostly something as ``detox -e py27,py37,fix_lint, coverage``.

src/tox/config/parallel.py
gaborbernat: do we need the Travis? I'm not expecting people running parallel mode on Travis. They probably should prefer parallel Travis job instead, not ?

nicoddemus: Actually we added that support in `pytest-xdist` because people were running in parallel on Travis. 🤷‍♂️

gaborbernat: Yeah, there it makes sense to parallelize tests. However in case parallelizing tox targets seems more idiomatic to parallelize at CI level (e.g. separate jobs). This is how tox-travis behaves too.

src/tox/session.py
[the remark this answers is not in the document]
gaborbernat: But that would mean that the workers now need to check if an item can be processed or not, moving more complexity into the worker, as compared to how it's now.

src/tox/util/spinner.py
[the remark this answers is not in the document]
gaborbernat: Nope, we only use the ``py.io.TerminalWriter``.

src/tox/session.py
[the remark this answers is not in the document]
nicoddemus: Well no worries, just thought I would mention the queue/consumer pattern. 👍 

It's not like the current implementation is terribly complicated anyway. 😁

docs/changelog/439.feature.rst
```diff
@@ -0,0 +1 @@
+Parallel mode added (effectively ``detox`` now is merged into tox), for more details see :ref:`parallel_mode` - by :user:`gaborbernat`.
```
nicoddemus: 👍

src/detox/__init__.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import, unicode_literals
+
+from tox import cmdline
+from tox.config import parallel
+
+
+def run():
+    parallel.DEFAULT_PARALLEL = "auto"
```
nicoddemus: 👍

src/tox/session.py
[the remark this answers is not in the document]
gaborbernat: Thanks for the suggestion, actually tried switching to this, but while trying to implement it realized the extra complexity for checking if an entry can be done is not worth it :+1:

obestwalter:
I'd like to have another look today, please don't merge yet :)

gaborbernat:
I'd like to release today as I'll not be able to fix after tomorrow if things come up. The PR has been open for well over a month now, I'll wait a few more hours (4pm UTC) and merge/release after. thanks 😀

obestwalter:
I don't think there is a pressing reason to get this feature out with this release, so I'd appreciate it, if you could leave this open. 

I'd like to discuss the user interface of this a bit more and we also have to think about transition from detox and deprecation of detox. We will likely have to make some adjustments to detox and make another release there to avoid pain there, so IMHO this is something that needs a bit more preparation and is not ready to be part of a release if you want to make on today.

gaborbernat:
Blocked on https://github.com/tox-dev/detox/issues/40 for now.

obestwalter:
Thanks for your patience. I have time today and tomorrow to prepare a proper funeral and come up with a plan how to make sure that this will be as painless as possible for everybody involved :)

obestwalter:
Thinking about it again: adding detox as command for this new functionality was BadIdea(TM), so this should not go forward. I'll make new detox release with an uppper bound for tox ad a warning and that should be it. that new parallel functionality can have a fresh start then without pretending to be something else.

gaborbernat:
going meta here:
```
 env DIFF_AGAINST=upstream/master PIP_INDEX_URL=http://localhost:3141/bernat/dev ../tox/.tox/dev/bin/tox -e fix_lint,py27,py37,coverage -p all                           1239ms  Fri 11 Jan 12:56:22 2019
.package create: /Users/bernat/git/tox/.tox/.package
.package installdeps: setuptools >= 40.0.4, setuptools_scm >= 2.0.0, <4, wheel >= 0.29.0
✔ OK fix_lint in 12.201 seconds
✔ OK py37 in 3 minutes, 44.444 seconds
✔ OK py27 in 3 minutes, 45.882 seconds
✔ OK coverage in 12.08 seconds
coverage recreate: /Users/bernat/git/tox/.tox/coverage
coverage installdeps: coverage >= 4.4.1, < 5, diff_cover
coverage installed: coverage==4.5.2,diff-cover==1.0.6,inflect==2.1.0,Jinja2==2.10,jinja2-pluralize==0.3.0,MarkupSafe==1.1.0,Pygments==2.3.1,six==1.12.0
coverage run-test-pre: PYTHONHASHSEED='4156706002'
coverage runtests: commands[0] | coverage erase
coverage runtests: commands[1] | coverage combine
coverage runtests: commands[2] | coverage report -m
Name                                  Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------------------------
src/tox/_pytestplugin.py                277     10     70      7    95%   27, 29, 69, 88, 135, 139, 249-252, 26->27, 28->29, 68->69, 87->88, 138->139, 169->182, 396->393
src/tox/_quickstart.py                  122     18     57      7    85%   76, 83, 91, 99, 117-128, 132-134, 275-277, 75->76, 82->83, 90->91, 98->99, 116->117, 176->186, 178->177
src/tox/cli.py                            7      7      0      0     0%   1-11
src/tox/config/__init__.py              897      3    341     15    99%   971, 1132, 1245, 191->199, 341->342, 570->582, 853->855, 883->884, 967->971, 1024->1025, 1084->exit, 1131->1132, 1244->1245, 1475->1476, 1498->1508, 1540->1541, 1554->1555, 1571->1590
src/tox/config/parallel.py               35      4      6      0    90%   14-15, 26-27
src/tox/interpreters.py                 120     35     32      3    63%   55-57, 74, 85, 87, 90-91, 141-179, 84->85, 86->87, 130->141
src/tox/package/__init__.py              50      4     10      2    90%   21-22, 55-56, 20->21, 51->59
src/tox/package/builder/isolated.py      64      3     25      2    94%   36-40, 24->27, 35->36
src/tox/package/view.py                  26      3     10      2    86%   34-35, 37, 30->36, 36->37
src/tox/result.py                        47      1      4      1    96%   14, 13->14
src/tox/session.py                      632     15    230     12    97%   59, 205, 225, 235-236, 268-270, 345, 511-513, 518, 552-553, 665, 58->59, 203->205, 224->225, 249->251, 256->258, 267->268, 332->exit, 510->511, 517->518, 627->629, 642->645, 664->665
src/tox/util/graph.py                    48      0     40      1    99%   65->exit
src/tox/util/spinner.py                 115      5     40      8    92%   16-18, 117, 125, 129, 15->16, 36->exit, 74->78, 113->114, 119->exit, 124->125, 130->exit, 149->152
src/tox/venv.py                         382     10    152     11    95%   95, 149, 186, 469-473, 498-499, 523, 94->95, 143->146, 146->149, 185->186, 189->188, 190->189, 266->exit, 396->397, 468->469, 495->459, 522->523
---------------------------------------------------------------------------------
TOTAL                                  2952    118   1049     71    95%

9 files skipped due to complete coverage.
coverage runtests: commands[3] | coverage xml -o /Users/bernat/git/tox/.tox/coverage.xml
coverage runtests: commands[4] | coverage html -d /Users/bernat/git/tox/.tox/htmlcov
coverage runtests: commands[5] | diff-cover --compare-branch upstream/master /Users/bernat/git/tox/.tox/coverage.xml
-------------
Diff Coverage
Diff: upstream/master...HEAD, staged and unstaged changes
-------------
src/tox/config/__init__.py (100%)
src/tox/config/parallel.py (88.6%): Missing lines 14-15,26-27
src/tox/session.py (99.2%): Missing lines 665
src/tox/util/__init__.py (100%)
src/tox/util/graph.py (100%)
src/tox/util/spinner.py (95.7%): Missing lines 16,18,117,125,129
src/tox/venv.py (100%)
-------------
Total:   337 lines
Missing: 10 lines
Coverage: 97%
-------------
_________________
```

gaborbernat:
Releasing now via #1134.

thanks all

dunossauro:
I love it :heart:

PR #1111, opened 2018-12-29
Clarify explanations in examples and avoid unsupported eol comments

### LINKED ISSUE #1109: tox.ini: make clear that end of line comments are not supported
The `ini` format is not really standardized, so basically all bets are off, but if you take the ConfigParser docs as a guide, it could be expected that end of line comments using a semikolon like:

```ini
[tox]
skipsdist = True ; some witty comment about this
```

should be supported. 

> Configuration files may include comments, prefixed by specific characters (# and ;). Comments may appear on their own in an otherwise empty line, or may be entered in lines holding values or section names. In the latter case, they need to be preceded by a whitespace character to be recognized as a comment. (For backwards compatibility, only ; starts an inline comment, while # does not.)

In the `tox.ini` they aren't and never where and I do not think it is necessary or even sensible to fix that. We will adding the possibility to configure tox in `pyproject.toml` anyway.

Going through the docs to fix examples where end of line comments are used and a remark that they are not supported would be sufficient to reduce potential confusion about that, I think.

obestwalter:
fixes #1109

[//]: # (rtdbot-start)

URL of RTD document: https://tox.readthedocs.io/en/fix-1109-end-line-comments/ Documentation Status

[//]: # (rtdbot-end)

### DIFF
--- a/docs/example/basic.rst
+++ b/docs/example/basic.rst
@@ -363,11 +363,11 @@ use :ref:`generative-envlist` and :ref:`conditional settings <factors>` to expre
     deps =
         django15: Django>=1.5,<1.6
         django16: Django>=1.6,<1.7
-        ; use PyMySQL if factors py34 and mysql are present in env name
+        # use PyMySQL if factors py34 and mysql are present in env name
         py34-mysql: PyMySQL
-        ; use urllib3 if any of py36 or py27 are present in env name
+        # use urllib3 if any of py36 or py27 are present in env name
         py27,py36: urllib3
-        ; mocking sqlite in python 2.x
+        # mocking sqlite in python 2.x
         py{27,36}-sqlite: mock
 
 Prevent symbolic links in virtualenv


obestwalter:
What's this bots problem?

gaborbernat:
Ignore it for now.

docs/example/basic.rst
```diff
@@ -356,9 +363,12 @@ use :ref:`generative-envlist` and :ref:`conditional settings <factors>` to expre
     deps =
         django15: Django>=1.5,<1.6
         django16: Django>=1.6,<1.7
-        py34-mysql: PyMySQL     ; use if both py34 and mysql are in an env name
-        py27,py36: urllib3      ; use if any of py36 or py27 are in an env name
-        py{27,36}-sqlite: mock  ; mocking sqlite in python 2.x
+        ; use PyMySQL if factors py34 and mysql are present in env name
```
gaborbernat: let's use in our examples only one comment character, aka ``#`` to not confuse people new to the documentation, the hash is the python comment to so I think it's ok to prefer that

obestwalter: true. done.

Issue #819, opened 2018-05-14
Improve the README


obestwalter:
The README is a bit terse and not a good starting point yet. I think it should already explain the most important concepts and have some good examples for basic usages showacsing some important functionality. A lot can be filched from the existing docs and edited for the README, I think.

Some things to mention:

* tox-quickstart
* plugins
* environment substitution
* posargs
* tox as CI frontend and developer task automation tool (not just for testing/packaging), e.g. build docs, pre-commit, etc.
* testing projects that don't even have a setup.py with `skipsdist`
* testing projects that don't build a package with `skispddist` + `usedevelop`
* ...

It could also be written in Markdown now, as the new https://pypi.org supports it.

ssbarnea:
I would add a section about tox related projects, official and non-official ones, plugins or stanalone. This would be very useful for users, and also avoid cases where people would ask about features that are already provided by a 3rd tool or extension.

gaborbernat:
I think we approved and merged the PR.

PR #1114, opened 2019-01-02
Improve README

obestwalter:
restart with a fresh addressing suggestions from previous PR

re-opening. Based on botched up https://github.com/tox-dev/tox/pull/1079

see #819

preview: https://github.com/tox-dev/tox/tree/819-improve-readme


obestwalter:
Looks like Github "sanitizes" all size hints: https://github.com/github/markup/issues/295

That would be one argument for switching to markdown and simply using HTML for things like that. I'd really like the logo right aligned and smaller next to the First heading and with markdown that worked.

Alternative in rst would be to keep a scaled copy and use that here.

gaborbernat:
I'm fine with switching to markdown, as long as works with both pypi and github 👍

obestwalter:
size is sanitized but you can use the raw directive - which is obviously much safer :rofl:

gaborbernat:
Do you think raw will work on PyPi too?

obestwalter:
A quick search suggests that it doesn't: https://stackoverflow.com/a/49107899/2626627

Question is if PyPI supports pure HTML coming from a markdown document either - I'll try it out on the test instance.

obestwalter:
raw directive is not supported for rst.

obestwalter:
It does work with markdown though :champagne: 

scrot

gaborbernat:
The CI still fails... 🤔

obestwalter:
Are we not checking the README with readme-renderer anymore? Can't find it, but might be packaged differently now.

gaborbernat:
readme-renderer is now part of twine 👍 because of PEP-517.

obestwalter:
```
2019-01-02T14:28:36.5844635Z package_description runtests: commands[1] | twine check '/home/vsts/work/1/s/.tox/package_description/tmp/build/*'
2019-01-02T14:28:37.0299156Z /home/vsts/work/1/s/.tox/package_description/lib/python3.7/site-packages/readme_renderer/markdown.py:38: UserWarning: Markdown renderers are not available. Install 'readme_render[md]' to enable Markdown rendering.
2019-01-02T14:28:37.0299940Z   warnings.warn(_EXTRA_WARNING)
2019-01-02T14:28:37.0300451Z Checking distribution /home/vsts/work/1/s/.tox/package_description/tmp/build/tox-3.6.2.dev13+g4e56ae3-py2.py3-none-any.whl: Failed
2019-01-02T14:28:37.0300991Z The project's long_description has invalid markup which will not be rendered on PyPI. The following syntax errors were detected:
2019-01-02T14:28:37.0301214Z line 4: Warning: "raw" directive disabled.
```

It's happening, but I haven't figured out yet where. Haven't looked at how we do stuff now yet, so will check.readme-renderer need top be installed with `[md]` for it to work.

obestwalter:
> readme-renderer is now part of twine +1 because of PEP-517.

cool but maybe without markdown check? I tried locally and still needed it pip install the md stuff.

obestwalter:
yep, that's the reason. Will check if there is an issue about this already. I guess we are not the only ones who want to do this and run into it.

gaborbernat:
https://github.com/pypa/twine/blob/master/setup.py#L84 could be a twine bug 👍 they probably should forward the extras 👍

obestwalter:
I just read a bit about it and am a bit confused about the state of markdown in readme-renderer, as apparently it can never fail, so this could mean its never checked but only fails when md extras are not installed. I'll add a dep in our tox env to install it extra but will get in touch with PyPA about how to solve this generally.

PR #1119, opened 2019-01-03
Add short name to --skip-missing-interpreters as -s

evandrocoan:
Related to:
1. https://github.com/tox-dev/tox/issues/672 --result-json does not respect --skip-missing-interpreters-option


Aareon:
Looks awesome! Thanks @evandrocoan

PR #1133, opened 2019-01-10
Improve accuracy of _alwayscopy_not_supported()

jdufresne:
Previously, this function checked against a whitelist of OSes that were
known to be a problem. This did include Fedora, which also fails this
test. Rather than chasing endless edge cases, test for this feature
programmatically.


jdufresne:
Thanks for the review. I have rebased and made the suggested changes. All additional feedback is welcome. :slightly_smiling_face:

gaborbernat:
Thanks now looks great 👍

PR #1140, opened 2019-01-14
improve README

obestwalter:
* @hackebrot pointed out that for beginners that example is not simple, but it's not minimal either, so we rather explain what it does
* add a link to the examples in the documentation


gaborbernat:
@obestwalter lint still broken

obestwalter:
ugh - editing in the web editor is not that great after all.

hackebrot:
Awesome! 👍

PR #1145, opened 2019-01-20
refactor session

gaborbernat:
While looking into doing #998 I realized the concept of env logs and actions have been heavily overlooked, furthermore the reporting should not be tied to the session (as we should report before the session object is constructed). This PR tries to fix all this and split up the longer and longer getting ``session.py``.


obestwalter:
Wow, that's a big one :)

Would you like feedback already about the direction this goes in? I'll have some time tomorrow afternoon to have closer look, if it makes sense already.

src/tox/_pytestplugin.py
```diff
@@ -107,12 +108,18 @@ def __init__(self, capfd, args):
 
     def __enter__(self):
         self._start = time.time()
-        self._capfd.readouterr()
+        self._out_tell = sys.stdout.tell()
+        self._err_tell = sys.stderr.tell()
```
asottile: hmmm how does this work

```
>>> pos = sys.stdout.tell()
  File "<stdin>", line 1, in <module>
io.UnsupportedOperation: underlying stream is not seekable
```

src/tox/_pytestplugin.py
```diff
@@ -210,36 +210,47 @@ def wait(self):
...
-    return MockSession()
+    return MockSession(config)
 
 
 @pytest.fixture
 def newmocksession(mocksession, newconfig):
     def newmocksession_(args, source, plugins=()):
-        mocksession.config = newconfig(args, source, plugins=plugins)
+        config = newconfig(args, source, plugins=plugins)
+        mocksession.__init__(config)
```
asottile: interesting, I've never seen this before -- I assume it just re-initializes the object? is this well defined?

src/tox/_pytestplugin.py
```diff
@@ -484,10 +475,25 @@ def tox_runenvreport(venv, action):
...
+
+    def popen(cmd, **kwargs):
+        # we don't want to perform installation of new packages,
+        # just replace with an always ok cmd
+        if "pip" in cmd and "install" in cmd:
+            cmd = ["python", "-c", "print({!r})".format(cmd)]
+        ret = None
+        try:
+            ret = subprocess.Popen(cmd, **kwargs)
+        except tox.exception.InvocationError as exception:  # pragma: no cover
```
asottile: this block is no-covered but also seems impossible (?) `Popen` could only raise `OSError` here afaik

src/tox/action.py
```diff
@@ -0,0 +1,190 @@
...
+        try:
+            if self.generate_tox_log and not redirect:
+                if process.stderr is not None:
+                    # prevent deadlock
+                    raise ValueError("stderr must not be piped here")
+                # we read binary from the process and must write using a binary stream
+                buf = getattr(sys.stdout, "buffer", sys.stdout)
+                out = None
+                last_time = time.time()
+                while 1:
```
asottile: probably existing, but I usually see this as `while True:`

src/tox/config/__init__.py
```diff
@@ -364,22 +365,7 @@ def tox_addoption(parser):
...
-        "above level two verbosity flags are passed through to pip (with two less level)",
-    )
-    parser.add_argument(
-        "-q",
-        action="count",
-        dest="quiet_level",
-        default=0,
-        help="progressively silence reporting output.",
-    )
+    add_verbosity_commands(parser)
```
asottile: I'd call this `add_verbosity_options`

src/tox/config/__init__.py
```diff
@@ -987,18 +973,23 @@ def __init__(self, config, ini_path, ini_data):  # noqa
...
-            if config_min_version > tox_version:
-                raise tox.exception.MinVersionError(
-                    "tox version is {}, required is at least {}".format(
-                        tox.__version__, self.config.minversion
+            # As older versions of tox may have bugs or incompatibilities that
+            # prevent parsing of tox.ini this must be the first thing checked.
+            config.minversion = reader.getstring("minversion", None)
+            if config.minversion:
+                tox_version = pkg_resources.parse_version(tox.__version__)
+                config_min_version = pkg_resources.parse_version(self.config.minversion)
```
asottile: I'd love for us to get away from `pkg_resources` at some point, but probably not today 😆

src/tox/constants.py
```diff
@@ -75,3 +78,10 @@ class PIP:
     ]
     INSTALL_SHORT_OPTIONS_ARGUMENT = ["-{}".format(option) for option in SHORT_OPTIONS]
     INSTALL_LONG_OPTIONS_ARGUMENT = ["--{}".format(option) for option in LONG_OPTIONS]
+
+
+_HELP_DIR = os.path.join(os.path.dirname(_THIS_FILE), "helper")
+VERSION_QUERY_SCRIPT = os.path.join(_HELP_DIR, "get_version.py")
+SITE_PACKAGE_QUERY_SCRIPT = os.path.join(_HELP_DIR, "get_site_package_dir.py")
+BUILD_REQUIRE_SCRIPT = os.path.join(_HELP_DIR, "build_requires.py")
+BUILD_ISOLATED = os.path.join(_HELP_DIR, "build_isolated.py")
```
asottile: should this use `importlib.resources` / `importlib_resources`?  iirc assuming `__file__` works will break zipapps / eggs / wheel execution

src/tox/helper/build_isolated.py
```diff
@@ -0,0 +1,14 @@
+import sys
+
+backend_spec = sys.argv[1]
```
asottile: wrap each of these scripts in `if __name__ == '__main__':` just for sanity sake?

maybe also give them `dashed-names.py` to better indicate that they aren't importable modules?

Maybe also name this directory `scripts` instead of `helpers`

(all minor suggestions, feel free to take and/or leave any of them!)

src/tox/package/__init__.py
```diff
@@ -1,9 +1,11 @@
 import py
-from filelock import FileLock, Timeout
 
 import tox
-from .view import create_session_view
+from tox.util.lock import get as hold_lock
```
asottile: v minor, `from tox.util import lock` ... `with lock.get(...):` seems readable too -- nbd though :)

src/tox/package/local.py
```diff
@@ -0,0 +1,64 @@
...
+
+def get_version_from_filename(basename):
+    m = _REGEX_FILE_NAME_WITH_VERSION.match(basename)
+    if m is None:
+        return None
+    version = m.group(1)
+    try:
+
+        return pkg_resources.packaging.version.Version(version)
+    except pkg_resources.packaging.version.InvalidVersion:
```
asottile: iirc this is `AttributeError` on debian due to their unvendoring

src/tox/reporter.py
```diff
@@ -0,0 +1,143 @@
...
+        self.logline_if(Verbosity.INFO, of, msg, **opts)
+
+    def verbosity2(self, msg, **opts):
+        self.logline_if(Verbosity.DEBUG, "verbosity2", msg, **opts)
+
+    def quiet(self, msg):
+        self.logline_if(Verbosity.QUIET, "quiet", msg)
+
+
+_INSTANCE = Reporter()
```
asottile: ah I see, so the meant of this change is making this a global

src/tox/session/__init__.py
```diff
@@ -0,0 +1,250 @@
+"""
```
asottile: I really dislike code in `__init__.py` -- it makes it very easy to accidentally circular import, and quite annoying to find the right file in an editor 🤷‍♂️

src/tox/_pytestplugin.py
gaborbernat: it will not if you do against stdin, but it does against pytest replaced stdin, where this tests do run 👍

asottile: Only by default, `pytest -s` and 💥

src/tox/_pytestplugin.py
gaborbernat: It's just a function call AFAIK 👍

asottile: seems [well defined but ill-advised](https://stackoverflow.com/questions/13091221/reinitialize-an-object-with-self-init)

src/tox/_pytestplugin.py
gaborbernat: Any sane way to get the run capfd from a fixture? (yeah this was a workaround for that)

asottile: does the original way (in master) work?

gaborbernat: only if you passed on the capfd from the test function, but I would like to avoid having to pass through that every time, as it's only used by this underlying fixture to collect the runtime output 👍

obestwalter:
I only had time to skim through, but I like the general direction of it. I'll try to carve out some more time tomorrow to play with it.

obestwalter:
I'd say let's run with this after the comments are addressed and test failures are fixed. 

Stripping down the several `__init__.py` modules to what really hast to be there can be also done in a later PR. 

Main thing is we keep backwards compatibility regarding the main objects that people use from other tools / plugins.

src/tox/session/__init__.py
gaborbernat: I rarely run into circular import problems once you have the proper abstractions, but as pointed out we can do this later.

src/tox/constants.py
gaborbernat: definitely a worthy optimization, maybe can be done as followup with zipapp support later on?

src/tox/config/__init__.py
gaborbernat: indeed 👍 any better suggestions to detect what's installed and what's not?

src/tox/_pytestplugin.py
gaborbernat: because we want to reset the report, calculated session venvs when a new config is loaded 👍 can use the _reset path though.

gaborbernat:
This looks OK for now. I'm not 100% settled on the reporting thing, but we can take it for a test drive. Maybe instead we should just use the logging framework which we kinda mimic. But this can be a subsequent PR to migrate to that.

cjerdonek:
It looks like this PR causes pip's CI to fail. Here is the traceback:

```
  File "/home/travis/virtualenv/python3.6.7/bin/tox", line 11, in <module>
    sys.exit(cmdline())
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox/session/__init__.py", line 42, in cmdline
    main(args)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox/session/__init__.py", line 63, in main
    retcode = session.runcommand()
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox/session/__init__.py", line 187, in runcommand
    return self.subcommand_test()
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox/session/__init__.py", line 215, in subcommand_test
    run_sequential(self.config, self.venv_dict)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox/session/commands/run/sequential.py", line 9, in run_sequential
    if venv.setupenv():
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox/venv.py", line 582, in setupenv
    status = self.update(action=action)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox/venv.py", line 240, in update
    self.hook.tox_testenv_create(action=action, venv=self)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/hooks.py", line 284, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/manager.py", line 67, in _hookexec
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/manager.py", line 61, in <lambda>
    firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/callers.py", line 208, in _multicall
    return outcome.get_result()
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/callers.py", line 80, in get_result
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pluggy/callers.py", line 187, in _multicall
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tox_venv/hooks.py", line 99, in tox_testenv_create
    venv.session.make_emptydir(venv.path)
AttributeError: 'VirtualEnv' object has no attribute 'session'
```

Apparently this is because `tox-venv` relies on the `VirtualEnv` object's `session` attribute: https://github.com/tox-dev/tox-venv/blob/31b874292ed3da5fee701d2a048db4c8b1ad8cae/src/tox_venv/hooks.py#L99

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/tox-dev/tox-venv">tox-dev/tox-venv</a></strong></div><div>Use Python 3 venvs for Python 3 test environments. Contribute to tox-dev/tox-venv development by creating an account on GitHub.</div></blockquote>

cjerdonek:
Also, `tox-venv`'s `setup.py` says:

```
install_requires=['tox>=2.0'],
```

PR #1148, opened 2019-01-25
Update getcommandpath to use the PATH variable set through setenv=PATH=...

### LINKED ISSUE #1146: DOUBT: Regarding PATH environment variable
How to make a command of `tox` modify the `PATH` environment variable, so that the other commands can use the modified `PATH` environment variable. As of now, if a command changes the `PATH` variable, then the changed `PATH` variable is not being used for the following commands.

Roughly, it is something like the following example
```
[testenv]
setenv = 
  PATH={env:PATH}:{env:HOME/install/bin}
passenv =
  PATH
commands =
  bash install_tool.sh
  tool
```
The command `bash install_tool.sh` would install and store a `binary executable` in `$HOME/install/bin`. Hence I have set the PATH initially, but finally the command `tool` is not working. The error I am getting is `InvocationError: could not find executable 'tool'`.

Could someone help me with this?

Bhargavasomu:
Fixes #1146

### DIFF
--- a/src/tox/venv.py
+++ b/src/tox/venv.py
@@ -165,7 +165,11 @@ class VirtualEnv(object):
         return py.path.local.sysfind(name, paths=[self.envconfig.envbindir])
 
     def _normal_lookup(self, name):
-        return py.path.local.sysfind(name)
+        if self.envconfig.setenv["PATH"]:
+            search_path = self.envconfig.setenv["PATH"].split(":")
+        else:
+            search_path = os.environ["PATH"].split(":")
+        return py.path.local.sysfind(name, paths=search_path)
 
     def _check_external_allowed_and_warn(self, path):
         if not self.is_allowed_external(path):
@@ -514,7 +518,12 @@ class VirtualEnv(object):
         os.environ.pop("VIRTUALENV_PYTHON", None)
         env = self._get_os_environ(is_test_command=is_test_command)
         bin_dir = str(self.envconfig.envbindir)
-        env["PATH"] = os.pathsep.join([bin_dir, os.environ["PATH"]])
+        if "PATH" in self.envconfig.setenv:
+            env["PATH"] = os.pathsep.join([bin_dir, self.envconfig.setenv["PATH"]])
+        else:
+            env["PATH"] = os.pathsep.join([bin_dir, os.environ["PATH"]])
+        # Remove redundantly duplicated directories in PATH without changing order
+        env["PATH"] = ":".join(list(dict.fromkeys(env["PATH"].split(":"))))
         self.session.report.verbosity2("setting PATH={}".format(env["PATH"]))
 
         # get command
--- a/tests/unit/test_venv.py
+++ b/tests/unit/test_venv.py
@@ -80,6 +80,32 @@ def test_create(mocksession, newconfig):
     assert venv.path_config.check(exists=False)
 
 
+def test_setenv_executable_found(tmpdir, monkeypatch, mocksession, newconfig):
+    config = newconfig(
+        [],
+        """
+        [testenv]
+        passenv=
+          PATH
+        setenv=
+          PATH={env:PATH}:""" + str(tmpdir) + """/scrap/bin
+    """,
+    )
+    # Creating an executable file
+    basedir = str(tmpdir) + "/scrap/bin"
+    if not os.path.exists(basedir):
+        os.makedirs(basedir)
+    temp_executable_path = basedir + '/temp_executable'
+    open(temp_executable_path, 'a').close()
+    os.chmod(temp_executable_path, 0o755)
+    assert os.path.isfile(temp_executable_path)
+
+    envconfig = config.envconfigs["python"]
+    venv = VirtualEnv(envconfig, session=mocksession)
+    p = venv.getcommandpath("temp_executable")
+    assert str(p) == temp_executable_path
+
+
 def test_commandpath_venv_precedence(tmpdir, monkeypatch, mocksession, newconfig):
     config = newconfig(
         [],


Bhargavasomu:
Could you please help me with the errors that I am seeing.

src/tox/venv.py
```diff
@@ -515,6 +515,11 @@ def _pcall(
         env = self._get_os_environ(is_test_command=is_test_command)
         bin_dir = str(self.envconfig.envbindir)
         env["PATH"] = os.pathsep.join([bin_dir, os.environ["PATH"]])
+        if "PATH" in self.envconfig.setenv:
+            env["PATH"] = os.pathsep.join([env["PATH"], self.envconfig.setenv["PATH"]])
+            # Remove redundantly duplicated directories in PATH
+            env["PATH"] = ':'.join(list(set(env["PATH"].split(":"))))
```
obestwalter: I don't think we should do that. A set is unordered and `PATH` is ordered - if I put myself in the shoes of the user, I would be surprised if that order would be arbitrarily changed. So I would completely remove that line.

obestwalter:
https://toxdev.visualstudio.com/tox/_build/results?buildId=1152 - black needed to reformat your changes. Please run `tox -e fix_lint` locally.

obestwalter:
The other error seems unrelated to your changes on first look but does not occur on master, so this needs looking into more closely. Please run the tests locally and have a closer look.

src/tox/venv.py
```diff
@@ -515,6 +515,11 @@ def _pcall(
         env = self._get_os_environ(is_test_command=is_test_command)
         bin_dir = str(self.envconfig.envbindir)
         env["PATH"] = os.pathsep.join([bin_dir, os.environ["PATH"]])
+        if "PATH" in self.envconfig.setenv:
+            env["PATH"] = os.pathsep.join([env["PATH"], self.envconfig.setenv["PATH"]])
+            # Remove redundantly duplicated directories in PATH
```
gaborbernat: I like the idea, but you should make sure you don't re-order path values (use a set and a list, to keep the first appearance) - also this logic should run always not just when setenv is set

src/tox/venv.py
```diff
@@ -515,6 +515,11 @@ def _pcall(
         env = self._get_os_environ(is_test_command=is_test_command)
         bin_dir = str(self.envconfig.envbindir)
         env["PATH"] = os.pathsep.join([bin_dir, os.environ["PATH"]])
+        if "PATH" in self.envconfig.setenv:
+            env["PATH"] = os.pathsep.join([env["PATH"], self.envconfig.setenv["PATH"]])
```
gaborbernat: this needs more thought, why join here? this so to say overrides the setenv 🤔 we should only join if the setenv includes the previous, not?

src/tox/venv.py
```diff
@@ -515,6 +515,11 @@ def _pcall(
...
+            # Remove redundantly duplicated directories in PATH
+            env["PATH"] = ':'.join(list(set(env["PATH"].split(":"))))
+        os.environ["PATH"] = env["PATH"]
```
gaborbernat: we should never modify the global path, chaning the env should be all you need to propagate to the subprocess call

Bhargavasomu: @gaborbernat if we don't modify the global path, then while doing the `_normal_lookup`, it won't identify the executable. Attaching the snippet for reference
https://github.com/tox-dev/tox/blob/d6cd8d42fa06cc02d747ebe92b3179bc677f0591/src/tox/venv.py#L167-L168

gaborbernat: that's a bug for the normal lookup though, we should inject the environments generated paths https://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath.sysfind to fix it

src/tox/venv.py
Bhargavasomu: Right, will have to do something about that.

Bhargavasomu:
@gaborbernat @obestwalter how come `Tox CI` hasn't run? Also please review the recent changes, I have fixed this in kind of an elegant way to read and understand. Please let me know for further changes.

Bhargavasomu:
Also when I installed this version of `tox`, the bug which I was personally experiencing got fixed !!

Bhargavasomu:
On `python3.5` running the test `tests/unit/test_venv.py` is behaving randomly. Please apply this PR on a `python3.5` environment and see the test behaving randomly on multiple runs. Could someone please explain me why that is happening?

Bhargavasomu:
EDIT : Fixed all the local errors in `py3.5` and `py3.4` etc. But it is not visible, because `Tox CI` is not running. Can someone please look into this?

gaborbernat:
This PR still lacks tests, without it it's small chance of getting accepted. Please  add  unit test for each feature improvement one by one, taking care of edge cases too. Furthermore I would prefer to have a one place path mangling, now you do it in two places which makes the code hard to maintain.

Bhargavasomu:
> path mangling, now you do it in two places which makes the code hard to maintain

@gaborbernat I understand your view, but the following issue was what I faced when I tried to `PATH` mangling only once.

In the following snippet, the path order to be maintained is `envbindir + os.environ["PATH"]`
https://github.com/tox-dev/tox/blob/d6cd8d42fa06cc02d747ebe92b3179bc677f0591/src/tox/venv.py#L514-L518.
But when I try to pass that variable to `_normal_lookup`, even `envbindir` would be passed, which in practice/theory is not correct. Hence I put it this way. Please correct me if I am wrong.

I will add test cases, in a while. Could you please let me know why `Tox CI` is not running?

gaborbernat:
https://dev.azure.com/toxdev/tox/_build/results?buildId=1153

Bhargavasomu:
@gaborbernat that was an old build. I have made 2 pushes after that. (force pushes). But it is not updating.

src/tox/venv.py
```diff
@@ -514,7 +518,12 @@ def _pcall(
         os.environ.pop("VIRTUALENV_PYTHON", None)
         env = self._get_os_environ(is_test_command=is_test_command)
         bin_dir = str(self.envconfig.envbindir)
-        env["PATH"] = os.pathsep.join([bin_dir, os.environ["PATH"]])
+        if "PATH" in self.envconfig.setenv:
+            env["PATH"] = os.pathsep.join([bin_dir, self.envconfig.setenv["PATH"]])
+        else:
+            env["PATH"] = os.pathsep.join([bin_dir, os.environ["PATH"]])
+        # Remove redundantly duplicated directories in PATH without changing order
+        env["PATH"] = ":".join(list(dict.fromkeys(env["PATH"].split(":"))))
```
asottile: this is (for instance) incorrect on windows -- use `os.pathsep` instead of `':'` -- also it's unnecessary (and imo a surprising behaviour) to remove duplicates so this modification probably shouldn't happen at all

asottile: (above is true for `.split(":")` as well -- quite surprised this patch passes tests on windows?

Bhargavasomu: This is not passing on the windows test cases, and you gave me a solution. Thankyou.

Bhargavasomu:
@asottile I have added the new testcase. Please check it out.

gaborbernat:
I'll be honest I still don't entirely see the reason for this. What you describe sounds like you want to side step the tox as entry point which I find very dangerous. The more I think about it the surer I am that the only reasonable thing which we could allow is injecting elements into the path before the environment path, but after the tox environments bin folder.

gaborbernat:
On the subject of simplifying the path I think it's ok and will help decrease log size and debug. We should definitely keep only the first instance of path occurances. We could also remove not existing paths (I'm not entirely settled on this yet though). That being said this change is orthogonal to this PR so maybe let's keep it separate.

Bhargavasomu:
@asottile @gaborbernat I'm really confused from all the above comments. Could you please frame it in a better way as to do what next?

gaborbernat:
@asottile was alluding to the fact that's probably more robust to alter the PATH from within the tests (via some session fixture e.g.) rather than inside tox. This way your tests will work also for cases when users are not calling the test from within tox (e.g. they do from within an IDE).

Bhargavasomu:
Thankyou for the support.

PR #1155, opened 2019-02-01
-eALL should also include environment from envlist

goneri:
According to the `--help` output, `-eALL` should work against specified
environments (ALL selects all). However, the environments from the
envlist key were ignored if they were not associated with a section.

For instance, with:
```
[tox]
envlist = py{a,b}
```

No environment was selected, and the default `python` was used instead.
With this patch, `tox` with test both `a` and `b`.


gaborbernat:
seal of approval
Thanks!

PR #1156, opened 2019-02-03
Remove `distshare` / `sdistsrc` functionality

asottile:
_this is a breaking change_


gaborbernat:
I would say yes, both go. But we'll need to do this after our next breaking change, aka most likely when we drop python 2 in the summer. Ant reason why you want to drop it now?

asottile:
mostly I found a large `.tox` directory in my homedir and wondered what it was.  no urgency -- I'll remove the other part of this as well 👍

asottile:
I can reopen then I guess

PR #1170, opened 2019-02-22
reqs: bump pytest to "< 5"

blueyed:
It was pinned to "< 4" in 1641dc2 without any specific reason.


asottile:
or just unbound imo

asottile:
Thanks!

PR #1173, opened 2019-02-23
ci: download cc-reporter only once

blueyed:


blueyed:
Not shared between jobs apparently - IIRC this is rather big, so that's a pity.

PR #1171, opened 2019-02-22
toxworkdir: use realpath

blueyed:
Fixes https://github.com/tox-dev/tox/issues/1169.

PR #1174, opened 2019-02-23
tests: do not change cwd

blueyed:
Ref: https://github.com/tox-dev/tox/pull/1171#issuecomment-466641517

### DIFF
--- a/src/tox/_pytestplugin.py
+++ b/src/tox/_pytestplugin.py
@@ -49,6 +49,15 @@ def work_in_clean_dir(tmpdir):
         yield
 
 
+@pytest.fixture(autouse=True)
+def check_cwd_not_changed_by_test():
+    old = os.getcwd()
+    yield
+    new = os.getcwd()
+    if old != new:
+        pytest.fail("test changed cwd: {!r} => {!r}".format(old, new))
+
+
 @pytest.fixture(name="newconfig")
 def create_new_config_file(tmpdir):
     def create_new_config_file_(args, source=None, plugins=()):
--- a/src/tox/_pytestplugin.py
+++ b/src/tox/_pytestplugin.py
@@ -350,7 +350,8 @@ def initproj(tmpdir):
         base.chdir()
         return base
 
-    return initproj_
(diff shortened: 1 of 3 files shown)


blueyed:
The following could be added / was used to find these:
```diff
diff --git a/src/tox/_pytestplugin.py b/src/tox/_pytestplugin.py
index 52eba9a..c59dcdb 100644
--- a/src/tox/_pytestplugin.py
+++ b/src/tox/_pytestplugin.py
@@ -49,6 +49,15 @@ def work_in_clean_dir(tmpdir):
         yield


+@pytest.fixture(autouse=True)
+def check_cwd_not_changed_by_test():
+    old = os.getcwd()
+    yield
+    new = os.getcwd()
+    if old != new:
+        pytest.fail("test changed cwd: %r => %r" % (old, new))
+
+
 @pytest.fixture(name="newconfig")
 def create_new_config_file(tmpdir):
     def create_new_config_file_(args, source=None, plugins=()):
```

tests/unit/test_z_cmdline.py
```diff
@@ -564,16 +564,15 @@ def test_hello(pytestconfig):
     )
 
     # see that things work with a different CWD
-    monkeypatch.chdir(base.dirname)
```
blueyed: Note that `initproj()` changed it already.

src/tox/_pytestplugin.py
```diff
@@ -49,6 +49,15 @@ def work_in_clean_dir(tmpdir):
...
 
+@pytest.fixture(autouse=True)
+def check_cwd_not_changed_by_test():
+    old = os.getcwd()
+    yield
+    new = os.getcwd()
+    if old != new:
+        pytest.fail("test changed cwd: {!r} => {!r}".format(old, new))
+
+
```
blueyed: @gaborbernat 
Since this adds overhead to every test I've not meant to keep it - maybe comment it out / only use it on CI?

gaborbernat:
It's alright, the overhead for now is negligible 👍and is a sane check

PR #1185, opened 2019-03-11
implement tox environment provisioning

### LINKED ISSUE #998: auto provision requires environment

gaborbernat:
Resolves #998.

### DIFF
--- a/tests/integration/test_provision_int.py
+++ b/tests/integration/test_provision_int.py
@@ -1,0 +1,63 @@
+import signal
+import subprocess
+import sys
+import time
+
+import pytest
+from pathlib2 import Path
+
+
+def test_provision_missing(initproj, cmd):
+    initproj(
+        "pkg123-0.7",
+        filedefs={
+            "tox.ini": """
+                    [tox]
+                    skipsdist=True
+                    minversion = 3.7.0
+                    requires = setuptools == 40.6.3
+                    [testenv]
+                    commands=python -c "import sys; print(sys.executable); raise SystemExit(1)"
+                """
+        },
+    )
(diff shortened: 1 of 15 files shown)


tests/integration/test_provision_int.py
```diff
@@ -0,0 +1,60 @@
...
+def test_provision_missing(initproj, cmd):
+    initproj(
+        "pkg123-0.7",
+        filedefs={
+            "tox.ini": """
+                    [tox]
+                    skipsdist=True
+                    minversion = 3.7.0
+                    requires = setuptools == 40.6.3
+                         l
```
asottile: spurious character?

gaborbernat: yeah, thanks for spotting 👍

PR #1189, opened 2019-03-12
do not warn if python path is empty on python3.4+

gaborbernat:


docs/changelog/1092.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+In a posix shell, setting the PATH environment variable to an empty value is equivalent to not setting it at all;
+therefore we no longer if the user sets PYTHONPATH an emtpy string on python 3.4 or later - by :user:`gaborbernat`.
```
asottile: emtpy

PR #1193, opened 2019-03-16
virtual environment creation is optional now (use the host tox python)

### LINKED ISSUE #1181: New option to skip creation of virtual enviroment
With all of the recent enhancements, tox almost replaces the need for using external scripts or make. However, for all of the scripts that are executed, tox creates a new venv (unless I'm missing something). For many people, tox is already running inside of a virtual env. It would be cool if that venv was allowed to be used for some tasks.

**_ENHANCEMENT issue:_**

**- [x] a clear problem statement with an example**

tox is often run within a virtual environment already.
It would be great to have an option for some environments to be able to run within the existing venv that tox is being run in already, and not create a new one.

Examples for many projects that would  benefit from this option:

* Installing developer dependencies to work on a project (but install them in the outer venv)
   * create new venv, activate, install tox, run `tox -e dev`, and start developing.
* distribution related activities
  * install wheel, build sdist and wheel, install twine, push to pypi
* run lint tools on all appropriate files
* basically, all the stuff that many people put in Makefiles

**- [x] suggested change with example**

I suggest `skipvenv=true` as an option that you can place in an environment.

**- [x] if you have want help to do a PR yourself**

I would be fine with helping with this, but I don't know where to start.

gaborbernat:
Resolves #1181.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -218,6 +218,21 @@ Defaults for each setting in this section are looked up in the::
 
 Complete list of settings that you can put into ``testenv*`` sections:
 
+.. conf:: host_env ^ true|false ^ false
+
+    If ``true`` tox will not create a virtual environment to run your commands inside and instead use the hosts
+    tox environment (the python from where tox is running). Some options for such environments cannot be changed,
+    and setting these config values will be ignored:
+
+    - :conf:`skip_install` is always ``True``,
+    - :conf:`deps` is always empty (in essence no install operation allowed to not break the host tox),
+    - :conf:`basepython` will always be the full path of the host python interpreter.
+
+    If one does specify pip install via the :conf:`commands` we'll not stop it, however one should consider doing
+    so is a bad practice and unsupported (do it at your own peril). This mode is designed to be used when the host
+    tox already contains all the dependencies you want and one wants to run a few python commands within that
+    (mostly standard library only dependencies).
+
 .. conf:: basepython ^ NAME-OR-PATH
 
     Name or path to a Python interpreter which will be used for creating the virtual environment,
--- a/docs/changelog/1181.feature.rst
+++ b/docs/changelog/1181.feature.rst
@@ -1,0 +1,1 @@
+Creating a virtual environment is now optional.
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -23,6 +23,7 @@ from tox.constants import INFO
 from tox.interpreters import Interpreters, NoInterpreterInfo
 from tox.reporter import update_default_reporter
 
+from .env import TestenvConfig
 from .parallel import ENV_VAR_KEY as PARALLEL_ENV_VAR_KEY
 from .parallel import add_parallel_config, add_parallel_flags
 from .reporter import add_verbosity_commands
@@ -132,6 +133,8 @@ class DepOption:
 
     def postprocess(self, testenv_config, value):
         deps = []
+        if testenv_config.host_env is True:
+            return deps
         config = testenv_config.config
         for depline in value:
             m = re.match(r":(\w+):\s*(\S+)", depline)
@@ -357,6 +360,11 @@ class SetenvDict(object):
 
 @tox.hookimpl
 def tox_addoption(parser):
+    add_cli_options(parser)
+    add_test_env_attributes(parser)
+
+
+def add_cli_options(parser):
     parser.add_argument(
         "--version",
         action="store_true",
@@ -412,6 +420,7 @@ def tox_addoption(parser):
         dest="sdistonly",
         help="only perform the sdist packaging activity.",
     )
+
     add_parallel_flags(parser)
     parser.add_argument(
         "--parallel--safe-build",
@@ -420,6 +429,7 @@ def tox_addoption(parser):
         help="(deprecated) ensure two tox builds can run in parallel "
         "(uses a lock file in the tox workdir with .lock extension)",
     )
+
     parser.add_argument(
         "--installpkg",
         action="store",
@@ -465,7 +475,6 @@ def tox_addoption(parser):
         help="write a json file with detailed information "
         "about all commands and results involved.",
     )
-
     # We choose 1 to 4294967295 because it is the range of PYTHONHASHSEED.
     parser.add_argument(
         "--hashseed",
@@ -494,7 +503,6 @@ def tox_addoption(parser):
     parser.add_argument(
         "--alwayscopy", action="store_true", help="override alwayscopy setting to True in all envs"
     )
-
     cli_skip_missing_interpreter(parser)
     parser.add_argument(
         "--workdir",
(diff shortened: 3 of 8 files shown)


docs/config.rst
```diff
@@ -218,6 +218,21 @@ Defaults for each setting in this section are looked up in the::
...
+
+    If ``true`` tox will not create a virtual environment to run your commands inside and instead use the hosts
+    tox environment (the python from where tox is running). Some options for such environments cannot be changed,
+    and setting these config values will be ignored:
+
+    - :conf:`skip_install` is always ``True``,
+    - :conf:`deps` is always empty (in essence no install operation allowed to not break the host tox),
+    - :conf:`basepython` will always be the full path of the host python interpreter.
+
+    If one does specify pip install via the :conf:`commands` we'll not stop it, however one should consider doing
```
asottile: this really concerns me.  I spend most of my time with my personal virtualenv activated which is writable.  I don't want some random thing I'm developing on to muck with that.  This feature allows them to do that.

gaborbernat: hence why not allowing by default but giving the option if they are explicit about it (aka commands only). Oliver was in favour of this a while back, not sure if he changed his mind.

asottile: I think that stops being ok because people fork and develop on other people's software.  granted they can already do some crazy stuff with `tox.ini` today, but this feature makes it way easier to do so on accident and without warning.

gaborbernat: Yeah I definitely have such worries hence my comments on the issue and disallowing deps and enforcing skip install true... It's non trivial to do changes by intent or accident. Requiring explicit install in command felt a decent middle ground. 🤔

asottile: I went to check out this branch to play with this, and it seems even when using `host_env` that a tox environment gets created.

Even if the feature worked (?) this feature "optimizes" 3 seconds away on first run and then no optimization on second+ runs.  I don't think that's worth enabling an easy to abuse feature.

```ini
[testenv:old]
skip_install = true
deps =
commands =
    python -c 'print("hello world")'
```

```console
$ time tox -e old
old create: /tmp/x/x/.tox/old
old runtests: commands[0] | python -c print("hello world")
hello world
  old: commands succeeded
  congratulations :)

real	0m3.454s
user	0m3.119s
sys	0m0.227s

$ time tox -e old
old runtests: commands[0] | python -c print("hello world")
hello world
  old: commands succeeded
  congratulations :)

real	0m0.107s
user	0m0.095s
sys	0m0.012s
```

gaborbernat: it does work, see the test added with this pr,  your pr doesn't enable the host tox env....

asottile: lol, so I did `git checkout 1181` and instead of checking out your branch it checked out a sha with that prefix 🤦‍♂️ 

Here's more accurate timing differences:

```console
$ time tox -e new
new installed: filelock==3.0.10,pluggy==0.9.0,py==1.8.0,six==1.12.0,toml==0.10.0,tox==3.7.1.dev24+g455e4de,virtualenv==16.4.3
new run-test-pre: PYTHONHASHSEED='2687712007'
new runtests: commands[0] | python -c 'print("hello world")'
hello world
  new: commands succeeded
  congratulations :)

real	0m0.557s
user	0m0.516s
sys	0m0.040s
$ time tox -e old
old create: /tmp/x/x/.tox/old
old run-test-pre: PYTHONHASHSEED='2024296911'
old runtests: commands[0] | python -c 'print("hello world")'
hello world
  old: commands succeeded
  congratulations :)

real	0m3.868s
user	0m3.528s
sys	0m0.244s
$ time tox -e old
old run-test-pre: PYTHONHASHSEED='1190934059'
old runtests: commands[0] | python -c 'print("hello world")'
hello world
  old: commands succeeded
  congratulations :)

real	0m0.535s
user	0m0.503s
sys	0m0.032s
```

I wonder what has regressed 400ms of startup time since the old version I was running as well 🤔 

with the current startup time, the savings is even less pronounced

gaborbernat: it's much more pronounced on Windows to be fair, for the unit test on my machine its a 20 second difference...

asottile: that's worse yeah, but only for the first run

I think you can also avoid that by setting `envdir` to some other tox env right?

gaborbernat: well sometimes only at first run is not a good enough excuse,  it could be fast straight from out of box  😃🤔

gaborbernat:
TBH I'm also fairly worried about the ease of exploitation, therefore given that two current active maintainers (out of three) have concerns I'll reject this as won't do for now.

obestwalter:
Oh ... Didn't see that one at all. 

I am afraid, I am simply not able everything that's going on, so I am dropping balls left and right - especially in tox (which tbh is due to me knowing that tox is in good hands atm mainly thanks to you two :heart:). 

With this one I would have opted in to have a go at it, as in certain special use cases it might actually be helpful. I created an [experimental (and very rudimentary) plugin](https://pypi.org/project/tox-direct/) 

Things like that don't really need to be in core anyway (at least not right away until it is clear that there is a bigger need in the community for it). We have decent hook support to make most things possible with plugins now.

docs/config.rst
obestwalter: Hi, sorry that I didn't chip in here. It completely went under my Radar. It's too late, but I agree with your decision anyway. 

I think nowadays that having the need to do that is an "infrastructure smell", meaning that it shows that there might be a deeper problem with the way how the project is structured or there are some huge problems regarding the ease of making dependencies available.

Something like this would have no business in any OSS project I think and I don't know if any project that could profit for that. 

As this might tamper with an environment that should not be tampered with this is definitely "dangerous", so not advisable in this form. 

But the situation is usually different behind corporate firewalls there it might be a quick fix in some existing projects that have been built with different standards and assumption. Making the developer workflow more pleasant with hacks like this one, might help tidying them over to a cleaner, more reproducible build, test and deploy workflow. Then those hacks can be removed again.

That's why I think It's better not in core. Not many people ask for it (and I really think they shouldn't). But having this in a plugin with a caveat might be helpful in cases like described above.

I started a plugin yesterdy that just scratched an itch without knowing about this. I will have a look what you did here and see if I can turn that plugin into something useful that we could refer to for situations like this.

choldgraf:
For what it's worth, I came across this myself as I was looking for just this feature. I find tox to be much more intuitive to configure than `make`, but in some cases I just want to quickly use my local environment rather than a tox-specific virtualenv. This is usually because I want to play around with a locally-installed dependency or quickly update a dep and test it out, but don't want to go through the process of specifying that in a `setup.py` file and re-building the whole virtualenv.

So in these cases, I think it would be useful to be able to do all the things `tox` does, but without building a new environment and just using whatever the current `PATH` happens to point to. Just my 2 cents!

gaborbernat:
There's a plugin for that maintained by Fedora, but we don't plan to provide it part of core at the moment 🙏

choldgraf:
amazing! I assume you mean https://github.com/fedora-python/tox-current-env

that works just fine :-)

gaborbernat:
Yes 👍

Issue #1192, opened 2019-03-15
Different output between --parallel and -p


timdaman:
I have run into a odd bug with the --parallel flag. When I use `-p auto` it works as expected but when I use the full form of the flag, `--parallel auto` I get test failures. The `auto` is passed as an argument to my pytest command. I would expect both of these flags to return the same results.

Below I have output showing:
* contents of my `tox.ini`
* `tox --version` output
* `pip list` showing versions of all the packages installed
* `-p auto` working
* `--parallel auto` failing

My OS in Mac OS Mojave

```
$  cat tox.ini 
[tox]
envlist = py35,py36,py37
tox_pyenv_fallback = False
isolated_build = True
skip_missing_interpreters = True

[testenv]
deps = pipenv
commands =
    pipenv install --skip-lock
    pytest {posargs: -v --random-order-bucket module}
setenv =
    LC_ALL = {env:LC_ALL:en_US.UTF-8}
    LANG = {env:LC_ALL:en_US.UTF-8}

$ pipenv run pip list
Package             Version
------------------- -------
atomicwrites        1.3.0  
attrs               19.1.0 
coverage            4.5.3  
filelock            3.0.10 
more-itertools      6.0.0  
pip                 18.1   
pluggy              0.9.0  
py                  1.8.0  
pyfakefs            3.5.8  
pytest              4.3.1  
pytest-cov          2.6.1  
pytest-random-order 1.0.4  
setuptools          40.6.3 
six                 1.12.0 
toml                0.10.0 
tox                 3.7.0  
tox-pyenv           1.1.0  
virtualenv          16.4.3 
wheel               0.32.3 

$ pipenv run tox --version
3.7.0 imported from /Users/tim/.local/share/virtualenvs/check_docker-8LxgOXjy/lib/python3.7/site-packages/tox/__init__.py
registered plugins:
    tox-pyenv-1.1.0 at /Users/tim/.local/share/virtualenvs/check_docker-8LxgOXjy/lib/python3.7/site-packages/tox_pyenv.py

$ pipenv run tox -p auto
✔ OK py37 in 24.014 seconds
✔ OK py35 in 24.991 seconds
✔ OK py36 in 25.597 seconds
______________________________________________________________________________________________ summary ______________________________________________________________________________________________
  py35: commands succeeded
  py36: commands succeeded
  py37: commands succeeded
  congratulations :)

$ pipenv run tox --parallel auto
✖ FAIL py37 in 20.412 seconds
[...]

PR #1197, opened 2019-03-17
Parallel arg bug, fix argument removal code

timdaman:
This resolves issue #1192 where the argument to --parallel was sometimes being passed as a `posarg`

Fixes: #1192

PR #1198, opened 2019-03-17
Parallel arg bug, remove broken arg cleanup code

timdaman:
This resolves issue #1192 where the argument to --parallel was sometimes being passed as a `posarg`

Fixes: #1192

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -68,6 +68,7 @@ Selim Belhaouane
 Stephen Finucane
 Sridhar Ratnakumar
 Sviatoslav Sydorenko
+Tim Laurence
 Ville Skyttä
 Xander Johnson
 anatoly techtonik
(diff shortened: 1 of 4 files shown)


gaborbernat:
How does this relates to #1197? And also add tests, please 👍

timdaman:
If merging this do NOT merge #1197, close it instead.

timdaman:
> How does this relates to #1197? And also add tests, please 👍

There is no unit tests that make sense for this since the code is being removed so there is nothing to test. In my other PR where I repair the logic I do have something to test but I I need help figuring out how.

asottile:
> 
> 
> > How does this relates to #1197? And also add tests, please 👍
> 
> There is no unit tests that make sense for this since the code is being removed so there is nothing to test. In my other PR where I repair the logic I do have something to test but I I need help figuring out how.

In either case you can still demonstrate the behaviour in the original ticket is fixed (whether or not deleting code fixes that!)

timdaman:
Not adding tests per comment in #1192. This also fixes a small bug in the unit-test themselves.

PR #1200, opened 2019-03-19
bump pre-commit

gaborbernat:


gaborbernat:
just want to say that auto-update is wonderful 😄

PR #1202, opened 2019-03-22
Fix deadlock with --parallel and lots of output

### LINKED ISSUE #1183: detox worked, `tox -p auto` hangs with `trial -j` tests
I've been happily running my test suite in `detox` for the past year or so.  Today I upgraded to `tox` with the `--parallel` option, and suddenly my `unit-tests` environment hangs, despite `mypy`, `lint`, and `integration-tests` (which uses the same test runner and similar options) all passing and exiting nicely.

Here are some hopefully relevant details:

- The test runner in question is `trial -j 8`, so the subprocess has subprocesses of its own, which may be confounding things.

- `unit-tests` is the longest running environment.

- When the tests hang, I see a `--installpkg` process as well as all of my `trial` worker processes hanging.  As such, I tried `--parallel--safe-build` just in case, but it didn't change the behavior at all.

> If submitting a BUG please provide:
> 
> - [ ] Minimal reproducible example or detailed description, assign "bug"

Sorry to say that I can't produce a minimal reproducer; thus far I've only managed to produce this on a proprietary test suite.

> - [x] OS and `pip list` output

macOS 10.14.3

```
Package    Version
---------- -------
filelock   3.0.10
pip        18.1
pluggy     0.9.0
py         1.8.0
setuptools 40.5.0
six        1.12.0
toml       0.10.0
tox        3.7.0
virtualenv 16.4.3
wheel      0.32.2
```

asottile:
Resolves #1183

### DIFF
--- a/src/tox/session/commands/run/parallel.py
+++ b/src/tox/session/commands/run/parallel.py
@@ -1,12 +1,22 @@
 import os
 import subprocess
 import sys
+import tempfile
 from collections import OrderedDict
 from threading import Event, Semaphore, Thread
 
 from tox import reporter
 from tox.config.parallel import ENV_VAR_KEY as PARALLEL_ENV_VAR_KEY
 from tox.util.spinner import Spinner
+
+if sys.version_info >= (3, 7):
+    from contextlib import nullcontext
+else:
+    import contextlib
+
+    @contextlib.contextmanager
+    def nullcontext(enter_result=None):
+        yield enter_result
 
 
 def run_parallel(config, venv_dict):
@@ -23,10 +33,12 @@ def run_parallel(config, venv_dict):
         max_parallel = len(venv_dict)
     semaphore = Semaphore(max_parallel)
     finished = Event()
-    sink = None if live_out else subprocess.PIPE
+
+    ctx = nullcontext if live_out else tempfile.NamedTemporaryFile
+    stderr = None if live_out else subprocess.STDOUT
 
     show_progress = not live_out and reporter.verbosity() > reporter.Verbosity.QUIET
-    with Spinner(enabled=show_progress) as spinner:
+    with Spinner(enabled=show_progress) as spinner, ctx() as sink:
 
         def run_in_thread(tox_env, os_env):
             res = None
@@ -41,7 +53,7 @@ def run_parallel(config, venv_dict):
                     args_sub,
                     env=os_env,
                     stdout=sink,
-                    stderr=sink,
+                    stderr=stderr,
                     stdin=None,
                     universal_newlines=True,
                 )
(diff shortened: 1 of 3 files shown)


src/tox/session/commands/run/parallel.py
```diff
@@ -1,13 +1,23 @@
 import os
 import subprocess
 import sys
+import tempfile
```
gaborbernat: I believe this is already implemented partially within Action, we just need to make this an Action and perhaps have a flag allowing this👍

asottile: this is way way simpler than involving `Action`

gaborbernat: I know it's simpler but not using the Action is a bug, as generally, that's what we **should** use. That was added especially for such use cases, and somewhere we even have an issue to make this content available as a file afterwards (for CIs); so that's the right solution to solve both at the same time.

asottile: that seems orthogonal and existing :P

gaborbernat: not really orthogonal given in sense that the buffer overload is what is breaking here things, and would we write to file that would not exist 🤔

asottile: we're talking small fix for a showstopping bug vs ~full rewrite of `--parallel` (and `Action` to some extent to make it threadsafe).  I'd much rather fix a bug than rewrite especially when the rewrite is likely to change and break behaviour.

gaborbernat: I don't believe making the change requires a full rewrite, probably just switching over to action and passing some flags along 🤔

asottile: I started looking at it and it requires passing enough state into this thread to build an Action which is no small task

gaborbernat: isn't just tox_env.new_action ? https://github.com/tox-dev/tox/blob/master/src/tox/venv.py#L116 🤔

asottile: it's not that simple, no -- `tox -p` is implemented by running sub process which invoke `python -m tox`: https://github.com/tox-dev/tox/pull/1202/files#diff-7f1e93c77a2f9f1b1f2c01a8950bad0dR25

gaborbernat: I know, I did it, bur action also uses underneath subprocess via popen.

gaborbernat: Anyways we can go ahead just mind you we'll need it switch at some point. Thought if not too hard we could do it now. If you find it non trivial we can hold for now.

asottile: my initial try ended in a different deadlock due to the log files, I'd rather not

gaborbernat: let's come back to this later then 👍

PR #1209, opened 2019-03-28
doc: improve/revisit ignore_errors

blueyed:
It is not like GNU Make's ``--ignore-errors`` after all (which exits
with status 0), but like GNU Make's ``--keep-going``.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -292,14 +292,14 @@ Complete list of settings that you can put into ``testenv*`` sections:
 
     .. versionadded:: 2.0
 
+    If ``false``, a non-zero exit code from one command will abort execution of
+    commands for that environment.
     If ``true``, a non-zero exit code from one command will be ignored and
-    further commands will be executed (which was the default behavior in tox <
-    2.0).  If ``false``, then a non-zero exit code from one
-    command will abort execution of commands for that environment.
+    further commands will be executed.  The overall status will be
+    "commands failed", i.e. tox will exit non-zero in case any command failed.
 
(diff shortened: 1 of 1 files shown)


blueyed:
Might be worth having a test for this - no existing one failed with the following patch:
```
diff --git i/src/tox/venv.py w/src/tox/venv.py
index 18a9dd8..f6ac19e 100644
--- i/src/tox/venv.py
+++ w/src/tox/venv.py
@@ -483,7 +483,10 @@ def test(
                 env = self._get_os_environ(is_test_command=True)
                 # Display PYTHONHASHSEED to assist with reproducibility.
                 action.setactivity(name, "PYTHONHASHSEED={!r}".format(env.get("PYTHONHASHSEED")))
+            got_error = False
             for i, argv in enumerate(commands):
+                if got_error:
+                    assert 0
                 # have to make strings as _pcall changes argv[0] to a local()
                 # happens if the same environment is invoked twice
                 message = "commands[{}] | {}".format(
@@ -521,6 +524,7 @@ def test(
                     self.status = "commands failed"
                     if not ignore_errors:
                         break  # Don't process remaining commands
+                    got_error = True
                 except KeyboardInterrupt:
                     self.status = "keyboardinterrupt"
                     raise
```

blueyed:
For reference:
```Makefile
false:
	false

true:
	true

all: false true
```

```
$ make all -i; echo $?
false
make: [Makefile:2: false] Error 1 (ignored)
true
0

$ make all -k; echo $?
false
make: *** [Makefile:2: false] Error 1
true
make: Target 'all' not remade because of errors.
2
```

PR #1220, opened 2019-03-28
fix #1184 add nospinner

### LINKED ISSUE #1184: option to disable parallel spinner for CI envs
In some cases (CIs mostly) it makes sense to not pollute the output with the progress bar, let's allow somehow to disable it in such rare cases 👍 #1142 An environment variable will suffice.

themr0c:
Add a  `--parallel-nospinner` option, as requested in #1184.

Signed-off-by: Fabrice Flore-Thebault <themr0c@users.noreply.github.com>


gaborbernat:
The expectation is that this will be used within CIs. In that context I think makes more sense to add an environment variable, to not make the user cli more complicated.

src/tox/session/commands/run/parallel.py
```diff
@@ -25,7 +26,7 @@ def run_parallel(config, venv_dict):
     semaphore = Semaphore(max_parallel)
     finished = Event()
 
-    show_progress = not live_out and reporter.verbosity() > reporter.Verbosity.QUIET
+    show_progress = (not live_out or not nospinner) and reporter.verbosity() > reporter.Verbosity.QUIET
```
webknjaz: I think it'd cleaner this way:
```suggestion
    show_progress = not nospinner and not live_out and reporter.verbosity() > reporter.Verbosity.QUIET
```
plus it'd abort other checks immediatelly...

src/tox/session/commands/run/parallel.py
```diff
@@ -13,6 +13,7 @@
 def run_parallel(config, venv_dict):
     """here we'll just start parallel sub-processes"""
     live_out = config.option.parallel_live
+    nospinner = config.option.parallel_nospinner
```
webknjaz: On the second thought. It's probably better to use less negation in vars. So this one would be less confusing:
```suggestion
    spinner = not config.option.parallel_nospinner
```

src/tox/config/parallel.py
```diff
@@ -60,7 +66,6 @@ def add_parallel_flags(parser):
         help="connect to stdout while running environments",
     )
 
-
```
webknjaz: This change is unrelated and wrong. As per PEP 8, it must be two line separator between higher-level functions.

src/tox/config/parallel.py
```diff
@@ -60,7 +66,6 @@ def add_parallel_flags(parser):
         help="connect to stdout while running environments",
     )
```
webknjaz:
```suggestion
 
 
```

src/tox/session/commands/run/parallel.py
```diff
@@ -25,7 +26,7 @@ def run_parallel(config, venv_dict):
     semaphore = Semaphore(max_parallel)
     finished = Event()
 
-    show_progress = not live_out and reporter.verbosity() > reporter.Verbosity.QUIET
+    show_progress = not live_out and not nospinner and reporter.verbosity() > reporter.Verbosity.QUIET
```
webknjaz:
```suggestion
    show_progress = allow_spinner and not live_out and reporter.verbosity() > reporter.Verbosity.QUIET
```

webknjaz:
@gaborbernat I think that having both env var and CLI arg is important. It's a quite common CLI practice to do so and its a default expectation from the UX PoV. I think I'd want to use it locally sometimes...

themr0c:
Changelog would be a `docs/changelog/1184.bugfix.rst` ?
I need some guidance for the test and the environment variable.

webknjaz:
@themr0c I'd argue that it's probably a feature, not a bug... It's a UX improvement. A bug related to that isn't tox's, I'd say it's a bug in CIs.

docs/changelog/1184.bugfix.rst
```diff
@@ -0,0 +1 @@
+New option `--parallel-nospinner` to disable the spinner in parallel mode.
```
webknjaz: https://github.com/tox-dev/tox/pull/1220#issuecomment-477623744

webknjaz: > New option

kinda implies _feature_

themr0c: OK

gaborbernat:
For tests see as example test_parallel_error_report.

> I think that having both env var and CLI arg is important. It's a quite common CLI practice to do so and its a default expectation from the UX PoV. I think I'd want to use it locally sometimes...

Not really, few peoples want to use it and those who do are probably comfortable enough to set the env var, so I would need again to ask for the env var, no cli flags.

themr0c:
All that super quick feedback is awesome! Much appreciated by a first time contributor!

I will defend the cli option:

As a user, I expect that `tox --help` gives me all options. i see that cli options are far better documented than environment variables. That's why I didn't even think about environment variables here. As it is now all the parallel related options are grouped, which is quite a feature I believe.

```
tox --help
usage: tox [--version] [-h] [--help-ini] [-v] [-q] [--showconfig] [-l] [-a] [-c CONFIGFILE] [-e envlist] [--notest] [--sdistonly] [-p VAL] [--parallel-nospinner] [-o]
           [--parallel--safe-build] [--installpkg PATH] [--develop] [-i URL] [--pre] [-r] [--result-json PATH] [--hashseed SEED] [--force-dep REQ] [--sitepackages] [--alwayscopy] [-s [val]]
           [--workdir PATH]
           [args [args ...]]

tox options

positional arguments:
  args                             additional arguments available to command positional substitution (default: None)

optional arguments:
  --version                        report version information to stdout. (default: False)
  -h, --help                       show help about options (default: False)
  --help-ini, --hi                 show help about ini-names (default: False)
  -v                               increase verbosity of reporting output.-vv mode turns off output redirection for package installation, above level two verbosity flags are passed through
                                   to pip (with two less level) (default: 0)
  -q                               progressively silence reporting output. (default: 0)
  --showconfig                     show configuration information for all environments. (default: False)
  -l, --listenvs                   show list of test environments (with description if verbose) (default: False)
  -a, --listenvs-all               show list of all defined environments (with description if verbose) (default: False)
  -c CONFIGFILE                    config file name or directory with 'tox.ini' file. (default: None)
  -e envlist                       work against specified environments (ALL selects all). (default: None)
  --notest                         skip invoking test commands. (default: False)
  --sdistonly                      only perform the sdist packaging activity. (default: False)
  -p VAL, --parallel VAL           run tox environments in parallel, the argument controls limit: all, auto - cpu count, some positive number, zero is turn off (default: 0)
  --parallel-nospinner             disable spinner (default: False)
  -o, --parallel-live              connect to stdout while running environments (default: False)
  --parallel--safe-build           (deprecated) ensure two tox builds can run in parallel (uses a lock file in the tox workdir with .lock extension) (default: False)
  --installpkg PATH                use specified package for installation into venv, instead of creating an sdist. (default: None)
  --develop                        install package in the venv using 'setup.py develop' via 'pip -e .' (default: False)
  -i URL, --index-url URL          set indexserver url (if URL is of form name=url set the url for the 'name' indexserver, specifically) (default: None)
  --pre                            install pre-releases and development versions of dependencies. This will pass the --pre option to install_command (pip by default). (default: False)
  -r, --recreate                   force recreation of virtual environments (default: False)
  --result-json PATH               write a json file with detailed information about all commands and results involved. (default: None)
  --hashseed SEED                  set PYTHONHASHSEED to SEED before running commands. Defaults to a random integer in the range [1, 4294967295] ([1, 1024] on Windows). Passing 'noset'
                                   suppresses this behavior. (default: None)
  --force-dep REQ                  Forces a certain version of one of the dependencies when configuring the virtual environment. REQ Examples 'pytest<2.7' or 'django>=1.6'. (default: None)
  --sitepackages                   override sitepackages setting to True in all envs (default: False)
  --alwayscopy                     override alwayscopy setting to True in all envs (default: False)
  -s [val], --skip-missing-interpreters [val]
                                   don't fail tests for missing interpreters: {config,true,false} choice (default: config)
  --workdir PATH                   tox working directory (default: None)

Environment variables
TOXENV: comma separated list of environments (overridable by '-e')
TOX_SKIP_ENV: regular expression to filter down from running tox environments
TOX_TESTENV_PASSENV: space-separated list of extra environment variables to be passed into test command environments
PY_COLORS: 0 disable colorized output, 1 enable (default)
```

Moreover, as a not-a-python-developer-by-trade, i could learn how to write the "add the cli option" proposal, and I stay clueless on how to write it the environment variable way.

ssbarnea:
I myself find the presence of envars more useful for both CI and non CI, as I can keep them in config files (.envrc) and avoid cluttering the (tox) calls. I am not against cli commands, but I just wanted to state that I find ENVars easier to use.

tests/unit/session/test_parallel.py
```diff
@@ -51,6 +51,28 @@ def test_parallel_live(cmd, initproj):
...
+            commands=python -c "import sys; print(sys.executable)"
+        """,
+            "pyproject.toml": """
+            [build-system]
+            requires = ["setuptools >= 35.0.2"]
+            build-backend = 'setuptools.build_meta'
+                        """,
+        },
+    )
+    result = cmd("--parallel", "all", "--parallel-nopinner")
```
webknjaz:
```suggestion
    result = cmd("--parallel", "all", "--parallel-nospinner")
```

themr0c:
Would the 2 added tests be sufficient?

I would be left with the last struggle: how to declare and document the new envvar.

tests/unit/session/test_parallel.py
```diff
@@ -51,6 +51,28 @@ def test_parallel_live(cmd, initproj):
     result.assert_success()
 
 
+def test_parallel_nospinner(cmd, initproj):
+    initproj(
```
gaborbernat: do add that env var you've seen over at the test I've pointed to, no need to actually create the venv vars here; also I would suggest add some sleep (0.2s enough) as part of the command to allow the spinner to trigger

gaborbernat: If you look at that output at the bottom of the help you have all env-vars clearly documented...

themr0c:
I wouldn't call "all CI in the world" a very low usecase ...

gaborbernat:
Most CIs though are configured by environment variables rather than flags 🤔

themr0c:
Okay, I understand the idea, I just don't understand how to write it that way.

gaborbernat:
See how TOX_TESTENV_PASSENV is implemented and tested 👍

tests/unit/session/test_parallel.py
themr0c: Sorry here we clearly switched from "out of my comfort zone but I will manage" to "I don't understand anymore and I can't provide the requested work".

Please feel free to decline this PR and rewrite it a better way, I will read it and learn a lot for the next time. No offense, really, I already leaned a lot today and abused to much of your time.

webknjaz:
@themr0c in such cases I usually start with `git grep -n` or in case of GitHub UI it's here: https://github.com/tox-dev/tox/search?q=TOX_TESTENV_PASSENV&unscoped_q=TOX_TESTENV_PASSENV

This leads us to https://github.com/tox-dev/tox/blob/f52e24092be676be5740b0e8efe422f500a6ffc6/src/tox/config/__init__.py#L671, and https://github.com/tox-dev/tox/blob/984ce025ad8e75e323877fb5f34128abcfba950d/src/tox/session/commands/help.py#L10, and https://github.com/tox-dev/tox/blob/f52e24092be676be5740b0e8efe422f500a6ffc6/tests/unit/config/test_config.py#L1139.

From there you can inspect code around those lines and get more clarity on how it works.

Does this sound helpful?

themr0c:
Thanks @webknjaz, i am afk atm, i will look at it later.

themr0c:
I am getting confused because we will end up with something that is clearly inconsistent:

* activate parallelism: `-p VAL, --parallel VAL `, available as CLI option only,
* deactivate spinner: `TOX_PARALLEL_NO_SPINNER`, available as ENV variable.

So to get parallelism working in a CI environment, we will have to setup both one CLI option and one ENV variable? Is this really desirable?

Moreover, to my understanding, as a general practice, the user should have the choice between CLI and ENV to setup an option. As a good example I see the `-e` CLI / `TOXENV` ENV. Some people will have more a ENV culture, other more a CLI culture I guess. Does it make sense? Or should the choice be made by the tool ? (Sorry for the naïve question)

gaborbernat:
In our case, we made it more a tool decision. Advanced features are env powered, basic ones cli based.

themr0c:
i will step back on this PR. I was not able to achieve something useful in time, and I won't get time in the near future. I will let the room for someone with a little more expertise on the subject. It was a nice try, though :).

Issue #1223, opened 2019-03-29
[BUG] Spinner fails in CI on UnicodeDecodeError


tpazderka:
After upgrading tox to version 3.8.* my CI fails to run parallel test with the following traceback:

```
Traceback (most recent call last):
  File "/usr/local/bin/tox", line 11, in <module>
    sys.exit(cmdline())
  File "/usr/local/lib/python3.5/dist-packages/tox/session/__init__.py", line 42, in cmdline
    main(args)
  File "/usr/local/lib/python3.5/dist-packages/tox/session/__init__.py", line 63, in main
    retcode = session.runcommand()
  File "/usr/local/lib/python3.5/dist-packages/tox/session/__init__.py", line 187, in runcommand
    return self.subcommand_test()
  File "/usr/local/lib/python3.5/dist-packages/tox/session/__init__.py", line 213, in subcommand_test
    run_parallel(self.config, self.venv_dict)
  File "/usr/local/lib/python3.5/dist-packages/tox/session/commands/run/parallel.py", line 29, in run_parallel
    with Spinner(enabled=show_progress) as spinner:
  File "/usr/local/lib/python3.5/dist-packages/tox/util/spinner.py", line 47, in __init__
    if _file_support_encoding(self.UNICODE_FRAMES, sys.stdout)
  File "/usr/local/lib/python3.5/dist-packages/tox/util/spinner.py", line 27, in _file_support_encoding
    char.encode(encoding)
UnicodeEncodeError: 'ascii' codec can't encode character '\u280b' in position 0: ordinal not in range(128)
```

The spinner works fine when tox is downgraded to version 3.7.0.

gaborbernat:
Resolved via #1225

PR #1225, opened 2019-03-29
#1223 resolve unicode spinner error

gaborbernat:


docs/changelog/1223.bugfix.rst
```diff
@@ -0,0 +1 @@
+Spinner fails in CI on ``UnicodeDecodeError`` - by :user:`gaborbernat`
```
asottile: Encode isn't it ;)

gaborbernat: Indeed 👍

PR #1226, opened 2019-03-29
release 3.8.2

gaborbernat:
👍

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,6 +11,21 @@ with advance notice in the **Deprecations** section of releases.
 
 .. towncrier release notes start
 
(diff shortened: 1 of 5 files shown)


webknjaz:
@gaborbernat it looks like you've signed this commit with a key which is not added to your GitHub account `B22359CD0519A4C0`. So it labels the commit as `Unverified` @ https://github.com/tox-dev/tox/pull/1226/commits.

FYI

webknjaz:
@gaborbernat same for tag signatures: https://github.com/tox-dev/tox/tags

gaborbernat:
@webknjaz thanks, fixed 👍

PR #1233, opened 2019-03-29
release 3.8.3

gaborbernat:
A few more bug fixes before the weekend 👍 

https://dev.azure.com/toxdev/tox/_build/results?buildId=1431

Azure seems to be sluggish this week 😢 

[//]: # (rtdbot-start)

URL of RTD document: https://tox.readthedocs.io/en/release-3.8.3/ Documentation Status

[//]: # (rtdbot-end)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -11,6 +11,18 @@ with advance notice in the **Deprecations** section of releases.
 
 .. towncrier release notes start
 
+v3.8.3 (2019-03-29)
+-------------------
+
+Bugfixes
+^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


webknjaz:
@gaborbernat FTR this tag and commit aren't GPG-signed at all... Dunno whether it's intentional

gaborbernat:
It's autogenerated by a tool, which seems cannot pull in the gpg signing so disable it 🤔 that's alright though 👍

webknjaz:
@gaborbernat I think it's Git's job to put GPG signature, it depends on the git config (or explicit CLI arg), so that release script probably has nothing to do with it :)

PR #1261, opened 2019-04-15
move to latest images

gaborbernat:


asottile:
iirc pytest-timeout and pytest-xdist don't play well together, which is probably _some_ of the flakiness

gaborbernat:
I'm thinking of removing pytest-timeout 🤔

asottile:
> 
> 
> I'm thinking of removing pytest-timeout 🤔

wfm 👍 

I'll look into the windows-specific flake, I suspect that'll affect other `pytest` users as well

src/tox/_pytestplugin.py
```diff
@@ -176,9 +176,12 @@ def outlines(self):
         return err + out
 
     def __repr__(self):
-        return "RunResult(ret={}, args={}, out=\n{}\n, err=\n{})".format(
+        res = "RunResult(ret={}, args={}, out=\n{}\n, err=\n{})".format(
             self.ret, " ".join(str(i) for i in self.args), self.out, self.err
```
asottile: this will error on non-ASCII unicode args in py2, in my branch I Changed this to `args={!r}` (which is more readable anyway

src/tox/_pytestplugin.py
```diff
@@ -176,9 +176,12 @@ def outlines(self):
...
         )
+        if sys.version_info > (3, 0):
```
asottile: I used:

```python
if six.PY2:
    return res.encode('UTF-8')
else:
    return res
```

PR #1262, opened 2019-04-15
🔥 Cut comments off dependency lines

### LINKED ISSUE #1260: Dependency entries with comments are not parsed correctly
So the reproducer is
```ini
deps = 
    yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915
```

And it results in:
```console
ERROR: could not install deps [flake8>=3.6.0,<4, yamllint>=1.11.1,<2, mock==2.0.0, pytest==3.6.0, pytest-cov==2.6.0, pytest-helpers-namespace==2019.1.8, pytest-mock==1.10.0, pytest-verbose-parametrize==1.4.0, pytest-xdist==1.26.0, shade==1.22.2, yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915, ansible>=2.7,<2.8]; v = InvocationError("/home/travis/build/ansible/molecule/.tox/ansible27-unit/bin/python -m pip install 'flake8>=3.6.0,<4' 'yamllint>=1.11.1,<2' mock==2.0.0 pytest==3.6.0 pytest-cov==2.6.0 pytest-helpers-namespace==2019.1.8 pytest-mock==1.10.0 pytest-verbose-parametrize==1.4.0 pytest-xdist==1.26.0 shade==1.22.2 'yapf>=0.25.0,<0.27  # pyup: < 0.27 # disable updates, conflicts with flake8 as per https://github.com/ansible/molecule/pull/1915' 'ansible>=2.7,<2.8'", 1)
```

I think that it may be a `configparser` bug as I recall something like this mentioned in that PEP about selection of config format for pyproject. But it still seems relevant to tox...

webknjaz:
This enables users to have inline comments in deps lines in `tox.ini`. In particular, this allows having pyup bot integration.

Ref: https://github.com/pyupio/dparse/issues/34
Ref: https://github.com/ansible/molecule/issues/1973

Fixes #1260

## Contribution checklist:

(also see [CONTRIBUTING.rst](/CONTRIBUTING.rst) for details)

- [x] added/updated test(s)
      in message body
- [x] added news fragment in [changelog folder](/docs/changelog)
  * also see [examples](/docs/changelog/examples.rst)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -360,6 +360,20 @@ Complete list of settings that you can put into ``testenv*`` sections:
     (Experimentally introduced in 1.6.1) all installer commands are executed
     using the ``{toxinidir}`` as the current working directory.
 
+    .. note::
+
+        .. versionadded:: 3.9.0
+
+        As of tox v3.9.0 deps entries having inline comments are
+        supported. During post-processing stage, tox cuts off any
+        substring starting with any number of whitespaces followed by
+        a ``#`` character.
+
+        For example, if a dependency is ``octomachinery==0.0.13  # pyup:
+        < 0.1.0 # disable feature updates`` it will be turned into just
+        ``octomachinery==0.0.13`` which later will be used in the ``pip
+        install`` command.
+
 .. conf:: platform ^ REGEX
 
     .. versionadded:: 2.0
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -21,9 +21,6 @@ the *ini-style* format under the ``tool.tox.legacy_tox_ini`` key as a multi-line
 Below you find the specification for the *ini-style* format, but you might want to skim some
 :doc:`examples` first and use this page as a reference.
 
-.. _ConfigParser: https://docs.python.org/3/library/configparser.html
-
-
 tox global settings
 -------------------
 
@@ -45,10 +42,10 @@ Global settings are defined under the ``tox`` section as:
     .. versionadded:: 3.2.0
 
     Specify python packages that need to exist alongside the tox installation for the tox build
-    to be able to start. Use this to specify plugin requirements (or the version of ``virtualenv`` -
-    determines the default ``pip``, ``setuptools``, and ``wheel`` versions the tox environments
-    start with). If these dependencies are not specified tox will create :conf:`provision_tox_env`
-    environment so that they are satisfied and delegate all calls to that.
+    to be able to start (must be PEP-508_ compliant). Use this to specify plugin requirements
+    (or the version of ``virtualenv`` - determines the default ``pip``, ``setuptools``, and ``wheel``
+    versions the tox environments start with). If these dependencies are not specified tox will create
+    :conf:`provision_tox_env` environment so that they are satisfied and delegate all calls to that.
 
     .. code-block:: ini
 
@@ -152,10 +149,8 @@ Global settings are defined under the ``tox`` section as:
 
     Activate isolated build environment. tox will use a virtual environment to build
     a source distribution from the source tree. For build tools and arguments use
-    the ``pyproject.toml`` file as specified in
-    `PEP-517 <https://www.python.org/dev/peps/pep-0517/>`_ and
-    `PEP-518 <https://www.python.org/dev/peps/pep-0518/>`_. To specify the virtual
-    environment Python version define use the :conf:`isolated_build_env` config
+    the ``pyproject.toml`` file as specified in `PEP-517`_ and `PEP-518`_. To specify the
+    virtual environment Python version define use the :conf:`isolated_build_env` config
     section.
 
 .. conf:: isolated_build_env ^ string ^ .package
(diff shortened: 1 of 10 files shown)


src/tox/config/__init__.py
```diff
@@ -176,6 +177,10 @@ def _replace_forced_dep(self, name, config):
                 return forced_dep
         return name
 
+    @staticmethod
+    def _cut_off_dep_comment(name):
+        return re.sub(r'\s+#.*', '', name).strip()
```
gaborbernat: probably should be a global/class level regex compile 👍 here just sub on it.

webknjaz: Why? What do you mean?

asottile: @gaborbernat is suggesting a `re.compile(...)` at the module scope so this call doesn't sometimes-compile a regex (performance)

webknjaz: I followed the style of other code around  like in line 143: https://github.com/tox-dev/tox/pull/1262/files#diff-d33d19b1ff1730f9a176eed3f27fe64aR143

webknjaz: @asottile I'd argue that `re` caches compiled regexes internally making such optimization useless.

gaborbernat: 😄 we should probably fix those too I think 🤦‍♂️ but out of scope for this PR

webknjaz: Anyway. This suggestion isn't exactly in the scope of this PR either. It's a further refactoring.

gaborbernat: https://stackoverflow.com/a/12514276 so seems it's mostly ok 👍 though implementation detail the cache

webknjaz: Yep, that's what I was referring to

webknjaz:
@gaborbernat is it fine now?

src/tox/config/__init__.py
asottile: it's more that this call could _sometimes_ trigger a compile side-effect slowdown, whereas the module constant will never trigger a compile side-effect

caches aren't perfect, if we know it's going to be used more than once it's better to compile and keep a strong reference ourselves and not depend on implementation detail (especially when that implementation detail is _random eviction_ in <cpython3.6!)

webknjaz: Okay, it all makes sense but it's a refactoring which should be done separately

gaborbernat: Is it possible for environment markers to contain the comment marker? https://www.python.org/dev/peps/pep-0496/ How do we handle this case?

webknjaz: @gaborbernat the comment marker is `#` and env markers use `;`. What's the problem?

gaborbernat: ```pywin32 >=1.0 ; sys_platform == '#my-magic-platform' or sys_platform == 'your-#-platform' # so what now ``` 
https://www.python.org/dev/peps/pep-0496/#id9

gaborbernat: we probably want to strip to the right of the first non-quoted comment marker only

webknjaz: @gaborbernat this will cut off only comments with a space in front of the hash char

webknjaz: I've done it like this because URLs can also contain fragments.

gaborbernat: one would argue we should do it even if it does now have space, string literals inside the env marker can have space and then # too

webknjaz: It's the easiest solution possible. I don't see any case where it's impossible to avoid space in front of #. That's why I'd stop here instead of making this more complicated.

gaborbernat: I would argue this needs to support both ``a#comment`` and ``pywin32 >=1.0 ; sys_platform == '# my-magic-platform' or sys_platform == 'your- #-platform' # so what now ``, so it needs to be more complicated.

webknjaz: It'd be hard to make it work in URLs (outside quotes), then: 
git+https://github.com/manahl/pytest-plugins.git#egg=pytest-verbose-parametrize&subdirectory=pytest-verbose-parametrize.
In this case, you never know whether it's a fragment or a comment

gaborbernat: Sounds like we found why we never implemented this. Maybe we should reject this, @asottile?

asottile: hmmm yeah seems like there's potential for over-zealous splitting here -- though I'm surprised that the ini parsing isn't stripping comments already :thinking: 

I think the risk of this change is pretty low though, I can't think of any _common_ dependency markers this would break

webknjaz: I think it's moderately possible to implement "in quotes" check while keeping the whitespace requirement.

Oh btw I think configparser under python2 strips off those comments, but not under python 3

asottile: tox (unfortunately) (and pytest) use an alternate config parser from `pylib` :S

webknjaz: ah. so what do we do now?

gaborbernat:
I'll merge it after we fix our CI.

webknjaz:
@gaborbernat thanks!

gaborbernat:
Please do specify/document the exact behaviour in config.rst under deps.

webknjaz:
okay

decentral1se:
Thanks for picking this up @webknjaz :rocket:

gaborbernat:
master fixed now

docs/config.rst
```diff
@@ -360,6 +360,20 @@ Complete list of settings that you can put into ``testenv*`` sections:
     (Experimentally introduced in 1.6.1) all installer commands are executed
     using the ``{toxinidir}`` as the current working directory.
 
+    .. note::
+
+        .. versionadded:: 3.9.0
```
webknjaz: @gaborbernat I hope I've extrapolated the upcoming version correctly...

webknjaz:
Thanks! Can I expect a release soon?

gaborbernat:
Not much content for a new release 🤔 but why not!

webknjaz:
I'd appreciate that since it'd unblock issues in Molecule :)

gaborbernat:
https://github.com/tox-dev/tox/pull/1265

PR #1263, opened 2019-04-15
test commit please ignore

asottile:
ignore me, just trying to trigger some py27 CI w/ instrumentation since I can't seem to repro this locally


asottile:
CC @gaborbernat 

going to keep my notes here on this

After adjusting pytest to error, here's the trace:

(I clobbered the stacktrace, but it's while `pytest` is inspecting local variables):

```
INTERNALERROR>     localsrepr = self.repr_locals(entry.locals)
INTERNALERROR>   File "d:\a\1\s\.tox\py27\lib\site-packages\_pytest\_code\code.py", line 682, in repr_locals
INTERNALERROR>     raise ValueError("{} {} {}".format(name, type(value), e))
INTERNALERROR> ValueError: result <type 'instance'> 'ascii' codec can't encode character u'\u280b' in position 40: ordinal not in range(128)
```

In this particular case, it's an assertion while producing a repr of an old-style-class assigned to `result`.  The only place of this in `tox`'s testsuite is in tox's pytest plugin.

This implementation of `repr` is buggy, but also pytest shouldn't die here (so if my hunch is correct ~3 bugs: a flaky test, a bad repr, and pytest crash)

https://github.com/tox-dev/tox/blob/5cf41e7f2f945bb3ba679e0e72e395a957cc0b78/src/tox/_pytestplugin.py#L178-L181

asottile:
I'm now trying to narrow down the actual failure, it seems that xdist forces `-v` which tickles the bug -- trying to figure out how to turn that off :S

gaborbernat:
Let's fix the buggy repr 😄

asottile:
I can take that 👍

gaborbernat:
I think I already addressed it on #1261 just running it

asottile:
this is the pytest bug btw: https://github.com/pytest-dev/pytest/issues/5089

we fixed the repr in #1264 and the flaky test will be fixed in #1261 

good work team, these are my favorite types of bugs to debug and fix and this is why I love working on open source and software

PR #1264, opened 2019-04-16
Fix RunResult unicode repr errors

asottile:


src/tox/_pytestplugin.py
```diff
@@ -176,9 +176,13 @@ def outlines(self):
...
     def __repr__(self):
-        return "RunResult(ret={}, args={}, out=\n{}\n, err=\n{})".format(
-            self.ret, " ".join(str(i) for i in self.args), self.out, self.err
+        res = "RunResult(ret={}, args={!r}, out=\n{}\n, err=\n{})".format(
+            self.ret, self.args, self.out, self.err
         )
+        if six.PY2:
+            return res.encode("UTF-8")
+        else:
+            return res
```
gaborbernat: can you make it one return?

gaborbernat: one less line 

```
        if six.PY2:
            res = res.encode("UTF-8")
        return res
```

asottile: `if` `else` is easier for humans to parse as it's clearer that there are two disparate branches -- for maintenance reasons it's also easier to update if functionality were to be added as it's clearer it needs to happen at two places at a glance.

I've seen too many bugs from using your suggested style 😆

gaborbernat: something must be wrong with me then because for me two returns are a lot harder to understand, reason and parse 🤔 I think it's harder to maintain as now I need to update in two locations while I can clearly reduce the second to one, needing to worry only about a single case...

asottile: your suggested pattern also makes mypy unhappy since it changes the type of the variable

gaborbernat: works in my projects with mypy 🤔

gaborbernat: Seems our test is broken, can you fix it? I think on Windows py27 the order is inverted or undefined 🤔

gaborbernat:
Seems a flaky test 😢

src/tox/_pytestplugin.py
asottile: yeah happy to hack on the flakey test -- I knew there was one just the cascade of pytest / repr / etc. bugs was hiding it :laughing:

gaborbernat: fixed it on my PR 👍 now only to figure out why coverage html is not getting published

asottile: lol it's always something >.<

gaborbernat: it worked a while back now it's missing 💭

gaborbernat: https://github.com/Microsoft/azure-pipelines-tasks/pull/9962 turns out the ability to upload coverage report was removed 😆

gaborbernat: now is auto-generated instead https://docs.microsoft.com/en-us/azure/devops/release-notes/2019/sprint-150-update#cobertura-code-coverage-report-updates

asottile: whoa that's cool

PR #1269, opened 2019-04-21
Trim passenv (tox>=2.8)

asottile:
Committed via https://github.com/asottile/all-repos


asottile:
I'm not sure if `PROGRAMDATA` is needed either, I was removing it without issue on my repos 🤷‍♂️

asottile:
flakey test, merging

gaborbernat:
Let's mark these tests flaky in the future before the merge to confirm and fix.

PR #1272, opened 2019-04-23
log env-var changes with double verbosity and into log files

### LINKED ISSUE #1361: log environment variable changes
To debug some failures it's important to understand how the environment changes alongside calls. Add a way to log these at two plus verbosity level.

gaborbernat:
Resolves #1361.

### DIFF
--- a/docs/changelog/1361.feature.rst
+++ b/docs/changelog/1361.feature.rst
@@ -1,0 +1,1 @@
+log environment variable changes when we're at two or more verbosity - by :user:`gaborbernat`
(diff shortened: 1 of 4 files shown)


obestwalter:
There is a lot going on here that on first sight has not much to do with the title of the PR. Can you explain this a bit more?

webknjaz:
Also needs rebasing...

gaborbernat:
Everything else not related to the tile is just noise because of the lack of rebase. Needs tests though, hence why this is waiting here patiently. The free time I've had in the last month has been focused on rewriting virtualenv which in turn would essentially make so that we can delete the entire interpreters package. WIP

webknjaz:
@gaborbernat you may want to integrate https://github.com/apps/wip

gaborbernat:
Withdrawn 👍

MaineSoNewOrleans504:
Help 
Idk he to use github

PR #1277, opened 2019-04-25
[WIP] simplify CI/remove code climate

gaborbernat:


gaborbernat:
@asottile take a look at the new format and tell me what you think

azure-pipelines.yml
```diff
@@ -1,203 +1,73 @@
...
-    condition: succeededOrFailed()
+- template: run-tox-env.yml@tox-azure
+  parameters:
+    jobs:
+      check:
+        py: '3.7'
+        toxenvs:
+        ? fix_lint
+        ? docs
+        ? package_description
```
asottile: hmmm! how does this syntax work?

gaborbernat: yaml sets?

asottile: wow, I've *literal*ly never seen this

asottile: apparently it's a "key indicator" so it makes mappings with no values?

```pycon
>>> yaml.safe_load('''\
... x:
...   ? a
...   ? b
...   ? c
... ''')
{'x': {'a': None, 'b': None, 'c': None}}
```

gaborbernat: yes per standard defintion

gaborbernat: changed back to array now 👍

gaborbernat:
I think the current design is the cleanest, and I'm proposing it as the official way to use tox in Azure Pipelines.

PR #1279, opened 2019-04-25
disable pep-517 for develop install

### LINKED ISSUE #1276: develop mode broken with pip 19.1
due to https://github.com/pypa/pip/pull/6370 (follow up discussion on https://github.com/pypa/pip/issues/6434)

gaborbernat:
Resolves #1276. CI will fail until pip 19.1.1 is released.

### DIFF
--- a/tests/unit/package/builder/test_package_builder_isolated.py
+++ b/tests/unit/package/builder/test_package_builder_isolated.py
@@ -28,6 +28,27 @@ def test_verbose_isolated_build(initproj, mock_venv, cmd):
     assert "Writing example123-0.5{}setup.cfg".format(os.sep) in result.out, result.out
 
 
+def test_isolated_build_develop(initproj, cmd):
+    initproj(
+        "example123-0.5",
+        filedefs={
+            "tox.ini": """
+                    [tox]
+                    isolated_build = true
+                    [testenv]
+                    usedevelop = true
+                    """,
+            "pyproject.toml": """
+                    [build-system]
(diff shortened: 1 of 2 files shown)


tests/unit/package/builder/test_package_builder_isolated.py
```diff
@@ -28,6 +28,27 @@ def test_verbose_isolated_build(initproj, mock_venv, cmd):
...
+        "example123-0.5",
+        filedefs={
+            "tox.ini": """
+                    [tox]
+                    isolated_build = true
+                    [testenv]
+                    usedevelop = true
+                    """,
+            "pyproject.toml": """
+                    [build-system]
```
webknjaz: How about another test without build-system? And +1 with build system but no build-backend...

tests/unit/package/builder/test_package_builder_isolated.py
```diff
@@ -28,6 +28,27 @@ def test_verbose_isolated_build(initproj, mock_venv, cmd):
     assert "Writing example123-0.5{}setup.cfg".format(os.sep) in result.out, result.out
 
 
+def test_isolated_build_develop(initproj, cmd):
+    initproj(
+        "example123-0.5",
+        filedefs={
+            "tox.ini": """
+                    [tox]
+                    isolated_build = true
```
webknjaz: Maybe +1 regression test with this set to false would be nice

gaborbernat:
Per upstream pip will fix this on their side.

PR #1283, opened 2019-05-01
add no download

### LINKED ISSUE #448: package upgrades post virtualenv creation by default should be off
In virtualenv 15, virtualenv by default attempts to upgrade pip, setuptools, and wheel while creating a virtualenv.  It does this by using whatever `PIP_INDEX_URL` is set (or doesn't at all if `--no-download` or `VIRTUALENV_NO_DOWNLOAD` is set).

In my ideal world, I'd love for tox to default to `VIRTUALENV_NO_DOWNLOAD=1`, but I can see why that may be surprising or not a popular choice (since it would differ from the choice made in pypa/virtualenv).  The main reason I want this is to have repeatable builds without being surprised by external factors changing / going down (pypi goes down pretty frequently! and setuptools upgrades pretty frequently (and breaks things too often)).

Since this is probably not reasonable, I think a good compromise would be to export `PIP_INDEX_URL=tox['indexserver']['default']` if provided -- this would solve our primary concern and ensure repeatable builds given the state of our index server.

gaborbernat:
Resolves #448.

### DIFF
--- a/src/tox/venv.py
+++ b/src/tox/venv.py
@@ -646,7 +646,6 @@ def prepend_shebang_interpreter(args):
     return args
 
 
-NO_DOWNLOAD = False
 _SKIP_VENV_CREATION = os.environ.get("_TOX_SKIP_ENV_CREATION_TEST", False) == "1"
 
 
@@ -658,7 +657,7 @@ def tox_testenv_create(venv, action):
         args.append("--system-site-packages")
     if venv.envconfig.alwayscopy:
         args.append("--always-copy")
-    if NO_DOWNLOAD:
+    if not venv.envconfig.download:
         args.append("--no-download")
     # add interpreter explicitly, to prevent using default (virtualenv.ini)
     args.extend(["--python", str(config_interpreter)])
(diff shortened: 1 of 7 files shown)


src/tox/venv.py
```diff
@@ -658,7 +657,7 @@ def tox_testenv_create(venv, action):
         args.append("--system-site-packages")
     if venv.envconfig.alwayscopy:
         args.append("--always-copy")
-    if NO_DOWNLOAD:
+    if not venv.envconfig.download:
         args.append("--no-download")
```
asottile: this requires a new-enough `virtualenv` -- the older option was `--never-download` (if you want further compatibility) -- I believe `--no-download` was added in virtualenv 14.x (lol, says me to _the current maintainer of virtualenv_ 😆 )

gaborbernat: indeed, I'm tempted to just require at least 14.0 for virtualenv 👍 three/four year-old one seems a sane requirement

asottile: pre-commit requires >=15.2, I think it's probably fine for `tox` to require modern things too :D

The-Compiler:
This was missing docs, see #1494.

PR #1284, opened 2019-05-01
do not allow setting the ``TOXENV`` or the ``-e` flag to override the listed default environment variables, these show up under extra

### LINKED ISSUE #720: Not all environments shown when 'TOXENV' set
# Steps to reproduce

1. Export `TOXENV`

       export TOXENV=py27

2. List all environments

       tox -av

This was seen with the latest, master version of [Sphinx](https://github.com/sphinx-doc/sphinx/blob/b00f78311fd359435c7082e7e3b7590c05fb005b/tox.ini).

# Expected result

All environments are still listed.

```
$ tox -a
docs
flake8
mypy
coverage
py27
py34
py35
py36
pypy
du11
du12
du13
du14
pylint
```

# Actual result

Only a limited number of tox envs are shown. These correspond to the envs that are explicitly defined rather than the implicit (?) ones, as seen [here](https://github.com/sphinx-doc/sphinx/blob/b00f78311fd359435c7082e7e3b7590c05fb005b/tox.ini#L23-L42)

```
$ tox -a
py27
docs
flake8
mypy
pylint
```

gaborbernat:
Resolves #720.

### DIFF
--- a/docs/changelog/720.bugfix.rst
+++ b/docs/changelog/720.bugfix.rst
@@ -1,0 +1,1 @@
+fix for ``tox -l`` command: do not allow setting the ``TOXENV`` or the ``-e`` flag to override the listed default environment variables, they still show up under extra if non defined target - by :user:`gaborbernat`
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -1058,7 +1058,7 @@ class ParseIni(object):
         self.handle_provision(config, reader)
 
         self.parse_build_isolation(config, reader)
-        config.envlist, all_envs = self._getenvdata(reader, config)
+        config.envlist, all_envs, config.envlist_default = self._getenvdata(reader, config)
 
         # factors used in config or predefined
         known_factors = self._list_section_factors("testenv")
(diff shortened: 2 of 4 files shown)


docs/changelog/720.bugfix.rst
```diff
@@ -0,0 +1 @@
+do not allow setting the ``TOXENV`` or the ``-e`` flag to override the listed default environment variables, they still show up under extra if non defined target - by :user:`gaborbernat`
```
asottile: maybe mention that this is about `tox -l`, took me  a while of reading the code to figure that out :D

gaborbernat: done

rpkilby:
@gaborbernat is it safe for plugins to set the `envlist_default` to affect the `-l` option? Or would this not be recommended?

gaborbernat:
It's a sane thing to do in case of a good reason. Depends on the use case if it's best practice or not :-)

rpkilby:
Yeah, it might be better to create a separate command for outputting the matched testenvs, instead of modifying the output of `-l`, which *should* list all testenvs.

Something like `-ls` would be useful for tox-factor, since it matches a subset of testenvs (e.g., all `py37-*` testenvs), and it's nice to verify what is being matched before actually running.

PR #1286, opened 2019-05-02
move to improved Azure template

gaborbernat:
The first iteration of the template turned out to be not that flexible when you needed to customize things. This one also has support for inline tox installation (eat our own dog food) and separates the tox host from the target Python (in preparation of python 2 EOL). This separation also proved we can't detect when python is on path... but not python3.5... So fixed that with it.

### DIFF
--- a/src/tox/interpreters.py
+++ b/src/tox/interpreters.py
@@ -226,6 +226,7 @@ else:
             if spec.name == "python":
                 # The standard names are in predictable places.
                 candidates.append(r"c:\python{}{}\python.exe".format(spec.major, spec.minor))
+        print('candidates', candidates)
         return check_with_path(candidates, spec)
 
     _PY_AVAILABLE = []
--- a/src/tox/interpreters.py
+++ b/src/tox/interpreters.py
@@ -1,10 +1,12 @@
 from __future__ import unicode_literals
 
-import distutils.util
 import json
+import os
 import re
 import subprocess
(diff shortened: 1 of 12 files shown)


src/tox/interpreters.py
```diff
@@ -124,57 +124,83 @@ def __str__(self):
...
-                return actual
-        # Use py.exe to determine location - PEP-514 & PEP-397
-        if m:
-            return locate_via_py(*groups)
-
-    # Exceptions to the usual windows mapping
-    win32map = {"python": sys.executable, "jython": r"c:\jython2.5.1\jython.bat"}
+            py_exe = sys.executable
+        if py_exe is None:
+            py_exe = py.path.local.sysfind(envconfig.basepython)
```
asottile: at least for pre-commit, I've found that this picks up a garbage `python2.7.exe` (from chocolatey) on azure, I switched the find_by_py_launcher to be higher priority

gaborbernat: how did you switch it, or are you proposing to do this now?

asottile: I basically just moved it up higher than the "find by exe name" bit: https://github.com/pre-commit/pre-commit/pull/1012/files#diff-235f2434ddb56aaeef64d44558e82943R112

gaborbernat: did it now 👍

PR #1298, opened 2019-05-15
show config now shows all config, filter-able, contains host tox python and package versions

gaborbernat:
``--showconfig`` overhaul:

 - now fully generated via the config parser, so anyone can load it by using the built-in python config parser
- the ``tox`` section contains all configuration data from config
- the ``tox`` section contains a ``host_python`` key detailing the path of the host python
- the ``tox:version`` section contains the versions of all packages tox depends on with their version
- passing ``-l`` now allows only listing default target envs
- allows showing config for a given set of tox environments only via the ``-e`` cli flag or the ``TOXENV`` environment variable, in this case the ``tox`` and ``tox:version`` section is only shown if at least one verbosity flag is passed

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -392,7 +392,8 @@ def tox_addoption(parser):
     parser.add_argument(
         "--showconfig",
         action="store_true",
-        help="show configuration information for all environments. ",
+        help="show live configuration (by default all env, with -l only default targets,"
+        " specific via TOXENV/-e)",
     )
     parser.add_argument(
         "-l",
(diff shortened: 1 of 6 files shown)


src/tox/config/__init__.py
```diff
@@ -1368,6 +1370,11 @@ def __init__(self, name, url=None):
         self.name = name
         self.url = url
 
+    def __str__(self):
+        return "IndexServerConfig(name={}, url={})".format(self.name, self.url)
+
+    __repr__ = __str__
```
asottile: if you just define `__repr__`, `__str__` will default to that

src/tox/session/commands/show_config.py
```diff
@@ -1,31 +1,76 @@
...
-from tox.version import __version__
+from six import StringIO
+
+from tox import reporter
+
+try:
+    import ConfigParser
+except ImportError:
+    # noinspection PyPep8Naming
+    import configparser as ConfigParser
```
asottile: six.moves.configparser

gaborbernat:
Argh Azure decided to go down again in West Europe 🤦‍♂

PR #1311, opened 2019-05-22
Add environment variable to disable parallel spinner for CI tools

### LINKED ISSUE #1184: option to disable parallel spinner for CI envs

zeroshift:
Adding TOX_PARALLEL_NO_SPINNER environment variable to disable the spinner for parallel runs. The purpose is to avoid issues with the spinner causing excess logging in CI tools such as Jenkins.

Closes #1184

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -60,6 +60,7 @@ Mikhail Kyshtymov
 Monty Taylor
 Morgan Fainberg
 Nick Douma
+Nick Prendergast
 Oliver Bestwalter
 Paweł Adamczak
 Philip Thiem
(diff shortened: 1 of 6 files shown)


zeroshift:
Should we rerun tests on this? I tested locally on OSX before submitting the PR. I'm not sure why that one test would fail. How should I proceed?

gaborbernat:
Can you amend your last commit to retrigger your build?

gaborbernat:
released via 3.12.0, thanks for your contribution 
thanks

PR #1317, opened 2019-05-23
fix windows selects incorrect spec

gaborbernat:
🤦‍♂ yeah, an overriding variable, quite sneaky

### DIFF
--- a/docs/changelog/1317.bugfix.rst
+++ b/docs/changelog/1317.bugfix.rst
@@ -1,0 +1,1 @@
+Fix Windows selects incorrect spec on first discovery - by :user:`gaborbernat`
--- a/src/tox/interpreters/windows/__init__.py
+++ b/src/tox/interpreters/windows/__init__.py
@@ -44,8 +44,7 @@ def locate_via_pep514(spec):
         if not _PY_AVAILABLE:
             from . import pep514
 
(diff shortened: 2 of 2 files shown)


asottile:
we should make sure that the tests pass on master since they haven't been running in a while heh

gaborbernat:
```
[5156] D:\a\1\s\.tox$ 'C:\hostedtoolcache\windows\Python\3.7.2\x64\python.exe' -m virtualenv --no-download --python 'C:\Python27amd64\python.exe' 
``` 
damn 🤔

Juanlu001:
Comment from the peanut gallery: I have been staring at this diff for 10 minutes and still don't understand what's the difference :man_facepalming:

gaborbernat:
the py variable is used in the second part of the function, by using it as a loop variable we override it , breaking the logic ☹️ on python 2 the loop variables leak.. 👌

Juanlu001:
:open_mouth: Thanks for clarifying!

itzdan:
Are you going to resolve this?

asottile:
> 
> 
> Are you going to resolve this?

? this patch is merged and released

ihabpalamino:
> this patch is merged and released

what the solution?

PR #1329, opened 2019-05-28
Documents envlist effect on interpreter lookup

ssbarnea:
Adds paragraph to tox documentation related to how it determines which python interpreter to use when it creates a new virtualenv.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -68,6 +68,7 @@ Pierre-Luc Tessier Gagné
 Ronald Evers
 Ronny Pfannschmidt
 Selim Belhaouane
+Sorin Sbarnea
 Sridhar Ratnakumar
 Stephen Finucane
 Sviatoslav Sydorenko
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -113,6 +113,10 @@ Global settings are defined under the ``tox`` section as:
         (e.g. ``py27.*`` means **don't** evaluate environments that start with the key ``py27``).
         Skipped environments will be logged at level two verbosity level.
 
+    ``envlist`` is also used to determine the preferrence list of python interpreters. For example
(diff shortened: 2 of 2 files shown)


asottile:
this seems false:

```
$ tox -e lint --notest
lint create: /tmp/x/.tox/lint
  lint: skipped tests
  congratulations :)
$ .tox/lint/bin/python --version
Python 3.6.7
$ cat tox.ini 
[tox]
envlist = lint,py{37,36}
skipsdist = true
```

in your case you probably have `tox` installed to a `python3.7` interpreter

ssbarnea:
@asottile I tested on two different system but I do not rule out being fooled by their setup.

If my statement is false, what is the correct way for assuring that tox will pick up the correct interpreter?

Clearly `basepython=` approach does not work because:

* if `basepython=python3` and the projects supports python3.5+, it will pick python3.4 if that is the default interepreter in the developer machine.
* if you set it to specific version, like `python3.5` it will fail to run on systems that do only have `python3.6` installed.

Mainly we need to tell tox to look for compatible interpreters based on a preference list.

gaborbernat:
At the moment there's no preference option for the base python (the python used to create a venv). It always is deterministic 100%. python3 - means you want whatever python3 on path is or what py -3 tells you on Windows.

gaborbernat:
You can achieve what you want via a plugin 🤔 we could add it to the core, but we don't have a good syntax yet for whatever the latest available interpreter it is.

ssbarnea:
I ended up creating a bug that describes this issue. Please check https://github.com/tox-dev/tox/issues/1330 and let us know how we can implement this missing feature.

This is a real issue for developer user experience for bigger projects which support many platforms and where is unrealistic to expect for ocasional developers to be able to install all possible python interpreters.

gaborbernat:
It's a feature request, not a bug. But presented some option that could work.

PR #1333, opened 2019-06-01
Improve --devenv help and ensure `--devenv ` does not delete

asottile:


obestwalter:
I don't quite get what this is supposed to do.

Calling `tox --devenv` already failed in the last version as it needs an argument. If I call `tox --devenv .` this should still delete the project I guess or doesn't it? 

So my comment mentioning the problem with envdir wasn't about accidentally the whole project. It was about breaking the promise that venvs end up in .tox by default. Overwriting arbitrary folders with a an unqualified venv name is still perfectly possible, isn't it?

obestwalter:
I thought about this again. For you having the venv in the project root is a **feature** so you want this to be the default. I understand that. For me this is a potential danger, because I am giving courses for beginners now and then and I see the footgun potential. This is the conflict and I would be interested to hear your thoughts on that. To me people like you and @gaborbernat are the prototypical power users who know their tools in and out (and they also reprogram them if they don't do what they shuold), but maybe sometimes you forget that the average user has not the kind of insight like you have?

gaborbernat:
I actually think exposing development environments at root level is what beginners expect, in my experience with my other colleagues.

obestwalter:
My experience is more along the lines of people not even having a clue yet what a venv is. So again there might be some slight bias here. Not all of us are deeply embedded in the Python ecosystem working in companies together with Python core devs (say hello to @pablogsal for me if you bump into him) :D

gaborbernat:
My experience is actually not based on core devs,just average python developers.

asottile:
> I don't quite get what this is supposed to do.
> 
> Calling `tox --devenv` already failed in the last version as it needs an argument. If I call `tox --devenv .` this should still delete the project I guess or doesn't it?

In the previous version `tox --devenv ''` (the empty string as an argument) would act as if `--devenv` wasn't passed at all, but it should have acted as if devenv was passed as the empty string (which should trigger an error about deleting the project).  `tox --devenv .` also now triggers the same error 👍

obestwalter:
Yeah ... IMO it's already treating symptoms we wouldn't have if we'd cure the illness though. See https://github.com/tox-dev/tox/pull/1326#issuecomment-498004831

pablogsal:
> (say hello to @pablogsal for me if you bump into him) :D

Hi @obestwalter :hand:  :)

PR #1345, opened 2019-06-21
Add note that directory is created if does not exist yet

### LINKED ISSUE #1342: changedir has a potentially surprising side-effect
If the directory to change into does not exist yet, it is created. I am not sure if this is even a good idea, but this is the behaviour for a long time now.

We should either update the [documentation](https://tox.readthedocs.io/en/latest/config.html?highlight=changedir#conf-changedir) or change the behaviour to crash if the user tries to change into a directory that does not exist.

What do you think?

obestwalter:
fixes #1342

Also fixed some surrounding typos.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -320,7 +320,7 @@ Complete list of settings that you can put into ``testenv*`` sections:
 
 .. conf:: whitelist_externals ^ MULTI-LINE-LIST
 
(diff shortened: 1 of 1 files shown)


obestwalter:
Do those worried bots mean anything?

gaborbernat:
Neah it's alright.

PR #1360, opened 2019-07-01
fix non-canonical names don't work with provisioning

### LINKED ISSUE #1359: non canonical names within requires break provisining
😢

gaborbernat:
Resolves #1359


src/tox/config/__init__.py
```diff
@@ -1161,7 +1161,7 @@ def ensure_requires_satisfied(config, requires, min_version):
             # noinspection PyBroadException
             try:
                 package = requirements.Requirement(require)
-                package_name = canonicalize_name(package.name)
+                package_name = package.name
                 if package_name not in exists:
                     deps.append(DepConfig(require, None))
                     exists.add(package_name)
```
asottile: I think the `exists` set still needs canonicalization -- but the `dist` does not (?).  The intention here was to use the canonicalization as a deduping mechanims (essentially what `pkg_resources.Requirement.key` was doing in tox 1.12) and it looks like I goofed and passed that to `distribution(...)` -- I think the minimal patch is to change the `distribution` call from `package_name` to `package.name` (1 character change)

gaborbernat: well we still need tests to validate and it's nice to break infinite loop provisioning, so a bit more characters 😄 but yeah what you say can work 👍

asottile: oh yeah of course :D

PR #1362, opened 2019-07-01
release 3.13.2

gaborbernat:
👍


webknjaz:
@gaborbernat @asottile interesting case: how do you think the bot should proceed? Ignore it in with neutral status report when there are additions to the changelog and removals of the fragments?

gaborbernat:
I think that's a very sensible case 😀 so yes 😁

PR #1364, opened 2019-07-04
Fix CI

gaborbernat:


gaborbernat:
@webknjaz so how do you tag this as internal change, does not need changelog 👍

gaborbernat:
seems like it 😢

webknjaz:
@gaborbernat *Did you mean https://github.com/sanitizers/chronographer-github-app/issues/11?*

webknjaz:
Also: you might consider having a special dev changelog fragments type for this case and render it elsewhere...

gaborbernat:
I was thinking of something even more simple, like maybe if the title is [skip changelog] 🤔 I don't really want changelogs for internal stuff. There's already git log for that.

webknjaz:
This check is very simplistic and it doesn't fetch commits related to the PR. So maybe it's possible to add it to the PR title. File a feature request under the bot repo then.

OTOH, I imaging see random people abusing it so it may need something like ACL allowing this to be done on PRs from the Core team only for example.

Why do you think that include/exclude won't solve the problem? Tests-only or docs-only PRs could be skipped automatically this way. How do you define "internal"? Can it be encoded as some set of rules/DSL?

gaborbernat:
@webknjaz I can't see this as a hard rule. I suppose as a maintainer I can always just ignore the error. It's a judgment call from the maintainers. Maybe support comments from maintainers?

webknjaz:
@gaborbernat comments would mean the need to scan all historic comments in the PR on each event to make a decision. This is at least one API call more (more calls in case of pagination). I'm trying to be minimalistic here. How about labels? They are bundled in the PR metadata in events, maybe except one case. Could you please rerequest the check at this page https://github.com/tox-dev/tox/pull/1364/checks so I'd check if labels are there...

gaborbernat:
will do 👍

webknjaz:
Ouch... Got a traceback :)

webknjaz:
Wow, there's no PR data bundled in the event. PR list is empty. It's probably a bug.

asottile:
cpython uses labels and it seems to work pretty well

webknjaz:
I'm concerned why the PR details aren't bundled in the check_run rerequest event in this org. It works well for another test org.

webknjaz:
@gaborbernat what exactly did you click? I cannot reproduce that in the test org...

gaborbernat:
Just the rerun check button 😁

webknjaz:
There's a few of them. Can you show a screenshot?

gaborbernat:
Can do tomorrow, I'm no longer besides a computer, maybe @asottile can help you 😄

asottile:
this button I assume:

webknjaz:
There's one/two more in the left sidebar. That's why I'm asking.

webknjaz:
@gaborbernat GitHub says I'm the author of the commit you submitted. Why?

gaborbernat:
Like to give you some extra credit 😁

webknjaz:
I just don't remember suggesting this change :)

webknjaz:
I just don't understand: was it intentional? :)

gaborbernat:
Not intentional 😅

webknjaz:
Alright. JFYI if you want to fix that it's `git commit --amend --reset-author` :)

PR #1371, opened 2019-07-11
Enforce rst file extension @ change fragments

webknjaz:


.github/config.yml
```diff
@@ -1,2 +1,5 @@
+chronographer:
+  enforce_name:
+    suffix: .rst
```
asottile: We could do this through pre-commit as well, check out how pytest does this

webknjaz: Well, the bot does a generic thing and it belongs there...

PR #1373, opened 2019-07-20
[RFC/WIP] _getdict: turn escaped newlines into real ones

blueyed:
This is meant for `getdict_setenv` only, so should be moved maybe (or
driven through an argument).

It is meant to support passing literal newlines to e.g. pytest's `-o
filterwarnings=` value (which expects a list of values separated by
newlines).

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -1432,6 +1432,8 @@ class SectionReader:
         for line in value.split(sep):
             if line.strip():
                 name, rest = line.split("=", 1)
+                # Turn escaped newlines into real newlines.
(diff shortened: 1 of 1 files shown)


webknjaz:
Mind adding a changelog fragment?

blueyed:
@webknjaz 
Sure, but I'd like to get it accepted first - also CI is failing.  Have not checked, but should either fix an existing test or/and add a new one then.

gaborbernat:
This feels a breaking change we cannot do as it stands. At the very least one would need to provide an escape hatchet. For this to be part of core we would need to do it in general, not just for the newlines. Otherwise, I'm tempted to reject this and encourage you to do it via a plugin.

PR #1380, opened 2019-07-27
Work around broken pyparsing release

asottile:


asottile:
related: https://github.com/pypa/packaging/issues/170

asottile:
oops, looks like this broke everything else, I wasn't quite ready to merge -- I'll go fix that 👍

gaborbernat:
Thanks did not notice that 😁

PR #1382, opened 2019-07-27
Fix testsuite and honor environment markers in requires

asottile:
accidentally broken in #1380

I also made tox support environment specifiers in the `requires` section

unfortunately, I still can't use that in the testsuite itself because tox tries to install tox which goes to public pypi and ends up with the current version which doesn't support environment specifiers 🤦‍♂. 
 instead of trying to move around the testing code for the `tox` wheel, I opted for a comment with clean up instructions 🤷‍♂

### DIFF
--- a/docs/changelog/1380.bugfix.rst
+++ b/docs/changelog/1380.bugfix.rst
@@ -1,0 +1,1 @@
+Honor environment markers in ``requires`` list - by :user:`asottile`
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -1165,6 +1165,9 @@ class ParseIni(object):
             # noinspection PyBroadException
             try:
                 package = requirements.Requirement(require)
+                # check if the package even applies
+                if package.marker and not package.marker.evaluate({"extra": ""}):
+                    continue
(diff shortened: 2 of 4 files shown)


gaborbernat:
I have a test somewhere that injects a devpi server part of the test suite, that way tox installing itself actually installs itself, not upstream itself.

asottile:
> I have a test somewhere that injects a devpi server part of the test suite, that way tox installing itself actually installs itself, not upstream itself.

yeah I saw that code but didn't want to rework the whole provision tests to incorporate that for what is essentially a temporary hack -- the other provision tests have this in place and it seems to be quite a bit of work :S

gaborbernat:
Overall it is, but made it a fixture so if you just request the fixture all should automatically work 🤔

asottile:
> Overall it is, but made it a fixture so if you just request the fixture all should automatically work 🤔

oh! where is it?

asottile:
I'm going to merge this for now to unblock the testsuite -- we can use that fixture later 👍

gaborbernat:
Damn is in one of rewritee branches 🤕 Maybe later we add it then.

PR #1383, opened 2019-07-27
Improve recreate check by looking for .tox-config1

asottile:


src/tox/venv.py
```diff
@@ -707,6 +707,8 @@ def cleanup_for_venv(venv):
         not venv.path.exists()
         # does exist, but it's empty => OK
         or not dir_items
+        # tox has marked this as an environment it has created in the past
+        or ".tox-config1" in dir_items
         # it exists and we're on windows with Lib and Scripts => OK
```
gaborbernat: Can't we refuse to delete anything without the tox config file? Simplify checks.

asottile: I wanted to continue to support a usecase like:

```bash
virtualenv venv
tox -r --devenv venv
```

gaborbernat: Seems very arcane use case 🤔

asottile: I think it's a pretty realistic one, maybe not directly in succession but consider someone who set up a project which used `make` to manage a virtualenv, but then wanted to switch to use `tox --devenv` to manage that same environment instead.

gaborbernat: Any ideas why the ci fails now?

gaborbernat: https://dev.azure.com/toxdev/tox/_build/results?buildId=2045&view=results take 2

asottile: weird

rpkilby: For `--devenv` can't we also add a `.tox-config1` file to the venv?

asottile: it does, yes -- `--devenv` invokes the normal tox environment provisioning mechanisms

asottile:
I manually triggered this run of the same PR, it didn't update the PR status but it did pass: https://dev.azure.com/toxdev/tox/_build/results?buildId=2048&view=results

src/tox/venv.py
rpkilby: Gotcha - missed the part where the env was created prior to devenv. Need ☕️

PR #1385, opened 2019-07-30
Remove dependence on md5

### LINKED ISSUE #1384: Consider removing or making MD5 optional
Hello,
recently I have been investigating an issue running some tests with Tox on a system configured in FIPS mode.
The tests failed immediately because Tox seem to be using MD5 which is a forbidden hash in FIPS mode as it is [considered truly broken](https://en.wikipedia.org/wiki/MD5#Security).

Given tox already has support for generating a SHA-256 Hash, I was wondering if you could consider either removing the use of MD5 completely or simply making it optional.

That would make it possible to run tests for components like python-cryptography in FIPS mode which is currently [blocked on tox](https://github.com/pyca/cryptography/pull/4960) and would be really useful in many settings where use of FIPS mode is mandatory in general.

I've found 2 usages of md5 in tox:

-----

https://github.com/tox-dev/tox/blob/1908b198be5fd6e92143a78b43caf636965bbc6e/src/tox/_pytestplugin.py#L506

This seems selfcontained and possibly replacable by:

    base_resolved_python_sha256=getdigest(sys.executable, hashtype="sha256")

-----

https://github.com/tox-dev/tox/blob/a8b34cc5690cef34fa91ca0f1950a1fdde0f8b72/src/tox/logs/env.py#L39

Here the md5 hash is computed togteher with sha256 and maybe it can be omitted entirely or made optional (if MD5 initialization raises an exception, you skip it).

-----

Would you accept a pull request?

asottile:
Resolves #1384

### DIFF
--- a/src/tox/venv.py
+++ b/src/tox/venv.py
@@ -23,7 +23,7 @@ from .config import DepConfig
 class CreationConfig:
     def __init__(
         self,
-        base_resolved_python_md5,
+        base_resolved_python_sha256,
         base_resolved_python_path,
         tox_version,
         sitepackages,
@@ -31,7 +31,7 @@ class CreationConfig:
         deps,
         alwayscopy,
     ):
-        self.base_resolved_python_md5 = base_resolved_python_md5
+        self.base_resolved_python_sha256 = base_resolved_python_sha256
         self.base_resolved_python_path = base_resolved_python_path
         self.tox_version = tox_version
         self.sitepackages = sitepackages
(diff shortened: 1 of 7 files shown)


src/tox/venv.py
```diff
@@ -266,11 +266,11 @@ def _getliveconfig(self):
         alwayscopy = self.envconfig.alwayscopy
         deps = []
         for dep in self.get_resolved_dependencies():
-            dep_name_md5 = getdigest(dep.name)
-            deps.append((dep_name_md5, dep.name))
-        base_resolved_python_md5 = getdigest(base_resolved_python_path)
+            dep_name_sha256 = getdigest(dep.name)
```
hroncok: If we call such variables just `dep_name_hash` etc., the name would not need to change again when `getdigest` is updated (for example once sha256 is considered "old and dangerous" in 10 years from now) :thinking:

gaborbernat: I don't expect this code to be around in 10 years 👍

PR #1393, opened 2019-08-08
Add quote marks so readers are not tripped up by spaces in their path…

bu-tegan:
tox ran beautifully on my laptop but when I moved to my work PC with a different version of python to test my project without installing five versions of python on my laptop it just wouldn't work. Careful inspection of the error message was not really helpful other than tox being unable to find my tests but I've had the spaces in path names issue with other software so I tried adding quotes in my tox.ini and it worked. I have made a trivial update to the documentation adding quotes on the page I was following. There may be more places where adding quotes would be advisable. Alternatively, tox should quote file paths internally making the documentation changes irrelevant.

### DIFF
--- a/docs/example/pytest.rst
+++ b/docs/example/pytest.rst
@@ -53,7 +53,7 @@ and the following ``tox.ini`` content:
     changedir = tests
     deps = pytest
     # change pytest tempdir and add posargs from command line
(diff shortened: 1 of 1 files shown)


gaborbernat:
I agree the change here should be internal not documentation side. For now will fly with the documentation change though.

PR #1400, opened 2019-08-20
WIP - tox 4

gaborbernat:


asottile:
lmk when you want me to take a look at this (@-me), going to mute until then o7

gaborbernat:
@asottile probably not in the next two weeks 👍 no worries

src/tox/log/env.py
```diff
@@ -0,0 +1,35 @@
...
+from copy import copy
+
+from .command import CommandLog
+
+
+class EnvLog:
+    """Report the status of a tox environment"""
+
+    def __init__(self, result_log, name, content):
+        self.reportlog = result_log
```
blueyed: JFI: this is unused (also now already (27d7a6a))

obestwalter:
I have some time and more importantly mind space now and would be like to work my way into this, would this make sense at the moment @gaborbernat?

gaborbernat:
It's definitely possible; though mind you the state of this branch is a very early alpha, with not all platforms supported yet 👍 Maybe first look at the code and tell me your thoughts 👍 I do plan to pick this up eventually but first would like to roll out the virtualenv rewrite.

obestwalter:
o.k. I am especially interested atm on how this can help us integrating docker better and how we can improve reporting. I am not 100% sure yet if a rewrite is a really good idea or a more incremental approach (which my gut feeling rather would tend to), but I'll definitely have a look to get a feeling for the direction you are heading.

gaborbernat:
As someone who maintains a POC version of the docker thing my experience is that a major rewrite is necessariy to make it sustainable. We want to end up with something that's also fast, not just correct; and does not require patching in 5 different locations. In the current state of things none of these three are easily achievable. 😢 The rewrite will make this possible though I feel like.

gaborbernat:
Will close for now, 👍

PR #1401, opened 2019-08-20
Fix reporting exit due to (real) signal

blueyed:
`subprocess` reports a negative exit code for signals (the assumption in
https://github.com/tox-dev/tox/pull/760#discussion_r165831005 was
wrong).

TODO:

- [x] faster test (takes >3s)

### DIFF
--- a/src/tox/exception.py
+++ b/src/tox/exception.py
@@ -17,18 +17,24 @@ def exit_code_str(exception_name, command, exit_code):
     """
     str_ = "{} for command {}".format(exception_name, command)
     if exit_code is not None:
-        str_ += " (exited with code {:d})".format(exit_code)
-        if (os.name == "posix") and (exit_code > 128):
+        if (exit_code < 0 or (os.name == "posix" and exit_code > 128)):
             signals = {
                 number: name for name, number in vars(signal).items() if name.startswith("SIG")
             }
-            number = exit_code - 128
-            name = signals.get(number)
-            if name:
-                str_ += (
-                    "\nNote: this might indicate a fatal error signal "
-                    "({:d} - 128 = {:d}: {})".format(number + 128, number, name)
-                )
+            if exit_code < 0:
+                # Signal reported via subprocess.Popen.
+                sig_name = signals.get(-exit_code)
+                str_ += " (exited with code {:d} ({}))".format(exit_code, sig_name)
+            else:
+                str_ += " (exited with code {:d})".format(exit_code)
+                number = exit_code - 128
+                name = signals.get(number)
+                if name:
+                    str_ += (
+                        ")\nNote: this might indicate a fatal error signal "
+                        "({:d} - 128 = {:d}: {})".format(exit_code, number, name)
+                    )
+        str_ += " (exited with code {:d})".format(exit_code)
     return str_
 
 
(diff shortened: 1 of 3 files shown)


src/tox/exception.py
```diff
@@ -17,18 +17,24 @@ def exit_code_str(exception_name, command, exit_code):
...
-            number = exit_code - 128
-            name = signals.get(number)
-            if name:
-                str_ += (
-                    "\nNote: this might indicate a fatal error signal "
-                    "({:d} - 128 = {:d}: {})".format(number + 128, number, name)
-                )
+            if exit_code < 0:
+                # Signal reported via subprocess.Popen.
+                sig_name = signals.get(0 - exit_code)
```
asottile: also just `-exit_code` or `-1 * exit_code`

tests/unit/test_exit_signal.py
```diff
@@ -0,0 +1,37 @@
...
+    p.wait()
+    out = p.stdout.read().decode()
+    assert p.returncode == 1
+    assert (
+        "ERROR: InvocationError for command {} tfile.py (exited with code -15 (SIGTERM))".format(
+            sys.executable
+        )
+        in out
+    )
+    assert p.stderr.read() == b""
```
blueyed: @asottile 
Is there a better pattern for this test?  (mainly with regard to speed)

asottile: there's a pytest plugin that's used elsewhere -- it won't be much faster but in most cases that's ok

blueyed: Do you mean the "initproj" fixture?
(that got too much in the way)

I think tweaking tox.ini would be good to just run the command if possible.

tests/unit/test_exit_signal.py
```diff
@@ -0,0 +1,37 @@
+import subprocess
```
blueyed: Should be moved to tests/integration.

gaborbernat:
Definitely use initproj and cmd instead of this import. Let's not make the project a lot slower for just this test.

tests/unit/test_exit_signal.py
asottile: ???

It should be as simple as:

```python
def test...(initproj, cmd):
    initproj(
        "pkg",
        filedefs={
            "tox.ini": """\
                [tox]
                skipsdist = true
                skips_install = true
                [testenv]
                commands = python tfile.py
            """,
            "tfile.py": """\
                import os, signal
                os.kill(os.getpid(), signal.SIGTERM)
            """,
        },
    )
    result = cmd("-e", "py")

    ...
```

blueyed:
> initproj/cmd

ok.
I think I've mainly skipped it since initially I wanted to kill the process from outside - but it can kill itself.

blueyed:
It fails to install tox therein when not using minversion?!  (using pytest directly)

> ERROR: Could not find a version that satisfies the requirement tox>=3.13.3.dev18+g3725c93.d20190820

Added test_exit_code_signal_2 - it is as slow, and fails due to installing 3.7.0, not using the dev version apparently?

tests/unit/test_exit_signal.py
blueyed: Yeah, but it's not.
Dropping the integration test.

(btw: should have been "skip_install")

blueyed:
Ping.

gaborbernat:
This now has been released under https://tox.readthedocs.io/en/latest/changelog.html#v3-14-1-2019-11-13

PR #1411, opened 2019-09-01
Add initial GitHub Actions CI/CD linters workflow

webknjaz:

PR #1412, opened 2019-09-01
Add GitHub Actions CI/CD tests workflow

webknjaz:


webknjaz:
Demo: https://github.com/webknjaz/tox/runs/209288285

obestwalter:
Hi, just saw the twitter review :) 

I am not really involved at the moment and this will still be the case for a while. My current focus is on teaching Python and testing with Python, so I am afraid to say that I simply don't have bandwith to make any non-trivial contributions (and that includes reviewing PRs). If push comes to shove I'll be there if called, but that's all I can promise at the moment. 

IIUC github actions is just wrapper around azure pipelines anyway and a not very mature one yet either (please correct me if i am wrong). So I'd opt against changing anything fundamental regarding our CI/CD pipeline, especially because @gaborbernat put a gargantuan effort in just last year to make it all work :)

gaborbernat:
I'm fine with having this if we make it work nicely 👍 Needs some fine-tunning for now.

webknjaz:
@obestwalter yes, I tagged you because it may be interesting to you from a testing perspective. Even if you don't do reviews I just wanted to share what I've got :)

I don't think that it's just a wrapper around Azure Pipelines. They seem to use the same or similar components in the UI at glance and probably some back-end but the config syntax is almost completely different AFAICS.

Not very mature? I don't disagree with such assessment :) They're going live (out of beta) on Nov 13. I mostly like what I see now, yet hitting some issues too. I also like that folks from GitHub react to issues/reports/calls for help on Twitter :)

@gaborbernat
It looks like it's hard to do 1-1 mapping of the features. There's no UI for displaying test results and coverage, though this could be worked around using Checks API.

I'm seeing this as maybe a possible extension to the existing CI in terms of env diversity: I've tried it in a several projects now and in most of them I hit corner cases. tox is probably the only one which didn't reveal any env-related problems (I was expecting to see a number of those).

This can also be seen as a free resource for working around parallelism limits in other CIs, e.g. just for running linters (#1411) or other automations that are tightly bound to GitHub event/workflows (commenting on PRs/issues of the first-time contributors with info pointing to contribution instructions, etc.)

This PR doesn't add all envs present in the current setup. It can be extended to do so.

I think that the main downsides are UIs for tests and coverage, everything else could be set up the same way.

Anyway, I don't insist on merging, just wanted to demo some of the possibilities of using gh workflows.

gaborbernat:
I want to come back to this, definitely interested in it once it matures 👍 for now I'll close it though as would cause regression in some features we have.

PR #1419, opened 2019-09-10
fix typo in doc on factored configuration

freddrake:
First example in the "Factors and factor-conditional settings" section pulls in ``unittest2``, but the explanation refers to ``unittest``.


asottile:
thanks!

Issue #1437, opened 2019-10-16
Extend default passenv to include SSL proxy vars


ssbarnea:
There is a list of 3 environment variables that are also needed for SSL proxies. Tox is currently passing proxy vars but missing to pass the SSL vars and the side effect is clear: called code will choke trying to verify the SSL certificates.

```
CURL_CA_BUNDLE
REQUESTS_CA_BUNDLE
SSL_CERT_FILE
```

I am fully aware that use can manually add them to their tox.ini file but this does not scale well at all when you have hundreds of repositories (like openstack for example).

gaborbernat:
This now has been released under https://tox.readthedocs.io/en/latest/changelog.html#v3-14-1-2019-11-13

PR #1439, opened 2019-10-16
Adds SSL proxy variables to default passenv

ssbarnea:
This shouls avoid SSL failures with HTTP requests from pip, requests
or other libraries.

Also sorts existing values in order to make it easier to update the
list in the future.

Fixes: #1437

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/docs/changelog/1437.bugfix.rst
+++ b/docs/changelog/1437.bugfix.rst
@@ -1,0 +1,1 @@
+Adds ``CURL_CA_BUNDLE``, ``REQUESTS_CA_BUNDLE``, ``SSL_CERT_FILE`` to the default passenv values. - by :user:`ssbarnea`
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -663,11 +663,14 @@ def tox_addoption(parser):
         value = list(itertools.chain.from_iterable([x.split(" ") for x in value]))
 
         passenv = {
-            "PATH",
-            "PIP_INDEX_URL",
+            "CURL_CA_BUNDLE",
             "LANG",
             "LANGUAGE",
             "LD_LIBRARY_PATH",
+            "PATH",
+            "PIP_INDEX_URL",
+            "REQUESTS_CA_BUNDLE",
+            "SSL_CERT_FILE",
             "TOX_WORK_DIR",
             str(REPORTER_TIMESTAMP_ON_ENV),
             str(PARALLEL_ENV_VAR_KEY),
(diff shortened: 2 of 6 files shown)


src/tox/config/__init__.py
```diff
@@ -663,11 +663,14 @@ def passenv(testenv_config, value):
...
         passenv = {
-            "PATH",
-            "PIP_INDEX_URL",
+            "CURL_CA_BUNDLE",
             "LANG",
             "LANGUAGE",
             "LD_LIBRARY_PATH",
+            "PATH",
+            "PIP_INDEX_URL",
+            "REQUESTS_CA_BUNDLE",
```
asottile: I don't think tox should know things about `requests`

ssbarnea: @asottile If I remember well, even pip is using requests more recently. If we do not include it, it will likely fail to run even pre-commit -- in fact that was why drove me to add them, I was not able to `tox -e lint` without them.

asottile: pip vendors a ca-cert bundle, which should be active without needing anything right?

gaborbernat: Well not really behind a corporate proxy; as the ca-cert bundle obviously doesn't know about corporate proxy certificates. These usually are set as env-vars though.

docs/changelog/1437.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Adds CURL_CA_BUNDLE, REQUESTS_CA_BUNDLE, SSL_CERT_FILE to the default
```
gaborbernat: would prefer to see these escaped, as 
```  
``CURL_CA_BUNDLE`` 
````

File: tox.ini [hunk 3]
```diff
@@ -19,7 +19,14 @@ description = run the tests with pytest under {basepython}
 setenv = PIP_DISABLE_VERSION_CHECK = 1
          COVERAGE_FILE = {env:COVERAGE_FILE:{toxworkdir}/.coverage.{envname}}
          VIRTUALENV_NO_DOWNLOAD = 1
-passenv = http_proxy https_proxy no_proxy SSL_CERT_FILE PYTEST_*
+passenv =
+    CURL_CA_BUNDLE
```
gaborbernat: why would you need it to set it there if it's done internally 🤔

ssbarnea: Because the tox version used to test tox itself is an older version, one that is not able to
pass these vars. I had to do this locally in order to be able to run tox-on-tox. If I would
install from source before it could have worked.

Issue #1444, opened 2019-10-21
tox in parallel leaks TOXENV to subprocess


hroncok:
When running tox from within tox in parallel mode, the tox env of the inner tox is changed unlike when running it without parallel mode. This used to work with detox.

## Get the latest tox

```
[totox]$ python3.7 -m venv venv
[totox]$ . venv/bin/activate
(venv) [totox]$ pip install tox
...
Successfully installed filelock-3.0.12 importlib-metadata-0.23 more-itertools-7.2.0 packaging-19.2 pluggy-0.13.0 py-1.8.0 pyparsing-2.4.2 six-1.12.0 toml-0.10.0 tox-3.14.0 virtualenv-16.7.6 zipp-0.6.0
```

## Create an inner project with simple tox.ini

```
(venv) [totox]$ mkdir inner
(venv) [totox]$ cat > inner/tox.ini
[tox]
envlist = py36,py37,py38
skipsdist = True

[testenv]
commands =
    python -c 'import sys; print(sys.version)'

(venv) [totox]$ cd inner
(venv) [inner]$ tox
py36 create: /home/churchyard/tmp/totox/inner/.tox/py36
py36 run-test-pre: PYTHONHASHSEED='3754984691'
py36 run-test: commands[0] | python -c 'import sys; print(sys.version)'
3.6.9 (default, Jul  3 2019, 15:08:07) 
[GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]
py37 create: /home/churchyard/tmp/totox/inner/.tox/py37
py37 run-test-pre: PYTHONHASHSEED='3754984691'
py37 run-test: commands[0] | python -c 'import sys; print(sys.version)'
3.7.5 (default, Oct 17 2019, 12:09:47) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]
py38 create: /home/churchyard/tmp/totox/inner/.tox/py38
py38 run-test-pre: PYTHONHASHSEED='3754984691'
py38 run-test: commands[0] | python -c 'import sys; print(sys.version)'
3.8.0 (default, Oct 16 2019, 12:47:36) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]
____________________________________________________ summary ____________________________________________________
  py36: commands succeeded
  py37: commands succeeded
  py38: commands succeeded
  congratulations :)
(venv) [inner]$ cd ..
```

## See that invoking the inner dir's tox via subprocess works

```
[...]

PR #1446, opened 2019-10-21
Use _TOX_PARALLEL_ENV environment variable to control the behavior

hroncok:
...but keep the TOX_PARALLEL_ENV variable for others to consume.

The problem with using environment variables to let tox know it's being run
as a parallel worker is the following:

Either you don't pass the environment variable, but the test frameworks and
test suites are unaware that tox is running in parallel.
This makes it hard for cases like pytest-django:
https://github.com/tox-dev/tox/pull/1139

Or you pass the environment variable, but it makes tox invoked within tox
think it is a parallel worker even when it isn't.
This makes it hard to test various tox plugins that invoke tox in their
integration test suite:
Fixes https://github.com/tox-dev/tox/issues/1444

By introducing two variables we can use one ("private", _TOX_PARALLEL_ENV) to
control the behavior of tox (and don't pass it) and another ("public",
TOX_PARALLEL_ENV) to let the tests and test frameworks know tox is running
in parallel (and pass it by default).

The integration test is a tad complicated invoking subprocess.Popen instead
of subprocess.run, to support Python 2.7 and 3.4.

## Contribution checklist:

- [x] added/updated test(s)
- [x] updated/extended the documentation - changelog only, I've checked and the current doc for the public variable still applies
- [x] added yourself to `CONTRIBUTORS` (preserving alphabetical order) -- already there

### DIFF
--- a/docs/changelog/1444.bugfix.rst
+++ b/docs/changelog/1444.bugfix.rst
@@ -1,0 +1,5 @@
+Fix nested tox execution in the parallel mode by separating the environment
+variable that let's tox know it is invoked in the parallel mode
+(``_TOX_PARALLEL_ENV``) from the variable that informs the tests that tox is
+running in parallel mode (``TOX_PARALLEL_ENV``).
+— by :user:`hroncok`
--- a/tests/integration/test_parallel_inception.py
+++ b/tests/integration/test_parallel_inception.py
@@ -1,0 +1,52 @@
+def test_parallel_inception(initproj, cmd):
+    initproj(
+        "inception-1.2.3",
+        filedefs={
+            # the outer config just has one env: graham
+            "tox.ini": """
+            [tox]
+            envlist = graham
+            skipsdist = True
+
+            [testenv]
+            commands =
+                python runner.py
+            """,
+            # the inner config has 3 different envs, 1 of them is graham
+            "inner": {
+                "tox.ini": """
+                [tox]
+                envlist = graham,john,terry
+                skipsdist = True
+
+                [testenv]
+                commands =
+                    python -c 'pass'
+                """
+            },
+            # the outer test runs the inner tox and asserts all 3 envs were run
+            "runner.py": """
+            import os
+            import subprocess
+            import sys
+
+            os.chdir("inner")
+            p = subprocess.Popen(("tox"), stdout=subprocess.PIPE, universal_newlines=True)
+            stdout, _ = p.communicate()
+            sys.stdout.write(stdout)
+            assert "graham" in stdout
+            assert "john" in stdout
+            assert "terry" in stdout
+            """,
+        },
+        add_missing_setup_py=False,
+    )
+
+    result = cmd("-p", "all", "-o")
+    result.assert_success()
+
+    # 1 from the outer, 1 from the inner
+    assert result.out.count("graham: commands succeeded") == 2
+    # those gentlemen are only inside
+    assert result.out.count("john: commands succeeded") == 1
+    assert result.out.count("terry: commands succeeded") == 1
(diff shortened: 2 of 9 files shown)


docs/changelog/1444.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Separate the environment variable that let's tox know it is invoked in
+the parallel mode (``_TOX_PARALLEL_ENV``) from the variable that informs
+the tests that tox is running in parallel mode (``TOX_PARALLEL_ENV``).
+— by :user:`hroncok`
```
gaborbernat: I feel like you should mention here what this delivers to users; aka allowing to run nested parallel executions?

hroncok:
The absence of a test is the reason this is marked as draft. It was the plan to add one.

I'll amend the changelog entry.

hroncok:
Added the test. I hate subprocess compat code :(

tests/integration/test_parallel_inception.py
```diff
@@ -0,0 +1,79 @@
+from __future__ import print_function, unicode_literals
+
+import subprocess
```
gaborbernat: I appreciate you've tried to do everything from scratch, but can you instead use our battle tested pytest fixture; that significantly simplifies the amount of code - see https://github.com/tox-dev/tox/blob/master/tests/unit/session/test_parallel.py#L15-L36 e.g.

hroncok: I've only seen initproj and it seemed like it does files I don't need. Will try with cmd

hroncok: amended

tests/integration/test_parallel_inception.py
```diff
@@ -0,0 +1,69 @@
...
+        sys.stdout.write(stdout)
+        assert "graham" in stdout
+        assert "john" in stdout
+        assert "terry" in stdout
+    """
+    )
+    runner = tmpdir / "runner.py"
+    runner.write_text(content, encoding="utf-8")
+
+    with tmpdir.as_cwd():
```
gaborbernat: you can pass this as cwd arg to cmd, no need to change

tests/integration/test_parallel_inception.py
```diff
@@ -0,0 +1,69 @@
+from __future__ import print_function, unicode_literals
+
+import textwrap
+
+from tox.util.main import MAIN_FILE
+
+
+def test_parallel_inception(tmpdir, cmd):
+    # the outer config just has one env: graham
+    content = textwrap.dedent(
```
gaborbernat: I think you can transform this to initproj; which would make the file structure declaration a simple dictionary (no textrap + write_text) would be needed

tests/integration/test_parallel_inception.py
hroncok: TypeError: run() got an unexpected keyword argument 'cwd'

tests/integration/test_parallel_inception.py
hroncok: I'll try.

hroncok: done

hroncok:
My problem is that I've started by copypasting parallel_interrupt :( Anyway, now it uses the fixtures (and those are really nice!).

hroncok:
I don't understand the test failures on the CI. I don't think they are related.

gaborbernat:
Hmm, is pytest-xdist broken? 🤔 seems the order of the items throw the error 🤔

hroncok:
What can I do to continue here?

hroncok:
Thanks, Bernát!

gaborbernat:
I'm traveling now and need to fix the CI before I release; so most likely will get a release on Wednesday or so; unless @asottile has time before 👍

hroncok:
I don't need a release. I just didn't want this to get stalled and eventually conflict.

Issue #148, opened 2016-09-17
__PYVENV_LAUNCHER__ causing issues on OS X with Python 3.3


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/148
- Originally reported by: [r1chardj0n3s](http://bitbucket.org/r1chardj0n3s)
- Originally created at: 2014-01-21T06:52:29.147

If the environment variable **PYVENV_LAUNCHER** is set when tox is launched then the pip subprocesses that tox fires off to install dependencies get confused and install everything into the virtualenv that tox comes from, rather than the test envs.

The "fix" is to add **PYVENV_LAUNCHER** to the list of environment variables deleted from the environment in the _pcall method (along with VIRTUALENV_PYTHON and PYTHONDONTWRITEBYTECODE).

Sorry I can't provide a better report than this but I'm not sure what **PYVENV_LAUNCHER** does - only that it causes trouble sometimes and is the specific cause of the problem in this case.

I've attached the output/logs of trying to run tox with flake8, with what I hope is a reasonable explanation of what's going on. It's kinda verbose, sorry, but it does show the failure and then success after I make the edit above.

PR #1452, opened 2019-11-06
Fix provisioning from pyvenv interpreter

Kentzo:
The bug is related to #148.

When the original interpreter [0] tox is called with is actually a pyvenv
it will set the `__PYVENV_LAUNCHER__` environment variable the path of pyvenv
interpreter.

If tox configuration specifies `requires` and the requirements are not
satifsfied by [0], tox will create an intermediate virtual environment [1].

After setting up [1] tox will delegate the original call to it by spawning
a new [1] interpreter with the list of arguments identical to [0].

It would also pass the value of `__PYVENV_LAUNCHER__` if present. This in turn
causes `sys.executable` to resolve to [0] instead of [1] therefore ignoring all
modifications including installed requirements.

The fix is to make sure that when [0] spawns [1], `__PYVENV_LAUNCHER__` is removed
from the environment.

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -31,6 +31,7 @@ Hazal Ozturk
 Henk-Jaap Wagenaar
 Ian Stapleton Cordasco
 Igor Duarte Cardoso
+Ilya Kulakov
 Ionel Maries Cristian
 Itxaka Serrano
 Jake Windle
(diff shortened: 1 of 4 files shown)


asottile:
@Kentzo could you fix the tests so this can be merged? thanks!

Kentzo:
@asottile Are you sure the failure is related to the change? I may take a look, but I'm completely unfamiliar with the codebase.

gaborbernat:
I don't think the ci failure is related, but let's not merge until we fix master and rebase. Thanks!

gaborbernat:
This now has been released under https://tox.readthedocs.io/en/latest/changelog.html#v3-14-1-2019-11-13

PR #1456, opened 2019-11-11
fix the failing CI on master

gaborbernat:
The PR also bumps the linter tool versions, and drop python3.4 as the CI no longer supports it.


gaborbernat:
That was long but succeeded 👍

Issue #1450, opened 2019-10-26
Add --durations flag


blueyed:
Durations are reported with `-vv`, but it might be useful to have it just by itself via `--durations`, since `-vv` is very verbose indeed (including pip progress etc).

Related issue: https://github.com/tox-dev/tox/issues/470

blueyed:
Any input here?
I would really like to use this to trim build logs down (my current workaround it split install and tests).
Happy to give a new `--durations` flag a shot myself.

gaborbernat:
This has been brought up earlier and agreed that this should be a plugin not part of core.

blueyed:
IIRC it was you creating the plugin (https://github.com/tox-dev/tox-durations), but then it moved into the core.  It is just that it is hidden/used with verbosity 2 now.

gaborbernat:
I'm not settled on having a dedicated core option for this, maybe other maintainers can chime in 🤔

blueyed:
What about displaying it with verbosity 1 already?

gaborbernat:
Maybe 🤔

PR #1457, opened 2019-11-13
Display durations with verbosity 1 already

blueyed:
Fixes https://github.com/tox-dev/tox/issues/1450.

### DIFF
--- a/docs/changelog/1450.feature.rst
+++ b/docs/changelog/1450.feature.rst
@@ -1,0 +1,1 @@
+Durations of actions are displayed with verbosity 1 already - by :user:`blueyed`.
--- a/src/tox/reporter.py
+++ b/src/tox/reporter.py
@@ -60,7 +60,7 @@ class Reporter(object):
         start = time.time()
         yield
         duration = time.time() - start
-        self.verbosity2(
+        self.verbosity1(
             "{} finish: {} after {:.2f} seconds".format(name, msg, duration), bold=True
         )
 


gaborbernat:
I'd rather keep it at level 2. If anything let's move pip verbose one level deeper.

blueyed:
Well, it was based on https://github.com/tox-dev/tox/issues/1450#issuecomment-552219848.

> If anything let's move pip verbose one level deeper

Might make sense in general, but there's no `verbosity3` yet at least.
It would also break the expected outcome for people using `-vv` and suddenly missing it.

> Also, such behaviour changes should be accompanied by tests to defend against regressions.

:+1:

I will go with a custom tox version I think, given that an explicit `--durations` flag to enable this always was also rejected already.

gaborbernat:
Still not sure why you don't want to do this via a plugin 🤔

blueyed:
> just bump the level of the print from above to use the verbosity1 instead of verbosity2.

Not sure what you mean - I've looked into creating a plugin that just monkeypatches `timed_operation` to use `verbosity1`, but that's likely not what you mean / is recommended.

Apart from that I'd rather use a fork with that simple patch, where I can also add TERM and CI to passenv by default.. :)

gaborbernat:
As I said before for both of those features I have some reservations. If you can convince another maintainer (@asottile or @obestwalter) that we should do this I'll accept; otherwise, I'll recommend to do it via a plugin.

asottile:
plug-in seems the best approach here :+1:

blueyed:
@gaborbernat
I've said already that I'm fine with a fork.

I would still be interested to know what you've meant with "just bump the level of the print from above to use the verbosity1 instead of verbosity2" in the context of a plugin though.

For reference: https://github.com/tox-dev/tox-durations

obestwalter:
> As I said before for both of those features I have some reservations. If you can convince another maintainer (@asottile or @obestwalter) that we should do this I'll accept

I just want to mention that I am more a hibernating maintainer and a lot of the finer things elude me as I haven't been able to follow discussions and development for quite a while ... again *sigh* :(

To give my opinion more from a user perspective, I would prefer completely decoupling durations from verbosity and instead have a  `--durations` flag as proposed in #1450 I simply had not seen this, otherwise I'd already would have said so.
Best explanation I can come up with why I feel that way:

1. IMO It's hard to decide what is the "correct" verbosity level to show durations - I'd say it's impossible to find the right level, because the right level doesn't exist. durations are just a different kind of information that I better explicitly aks for rather than having them pop up from a certain level onwards (and yes: this viewpoint is also a matter of taste :)). The more I teach people in the use of tox, the more I am aware of the shortcomings of our output in general and I really hope I can help improve that, so I'd like to start with us thinking about disentangling the kind of information we burden the user with in any given tox run. Durations would be an easy win to unclutter the output (even if it is verbose) by saying: "durations" is something you can have, if you explicitly ask for it.

2. Thinking across project borders: pytest has `--durations` and why should we do it differently from our sister project? Learn how to get durations in one project and you know how to get them in the other ...

PR #1460, opened 2019-11-13
release 3.14.1

gaborbernat:


docs/changelog.rst
```diff
@@ -11,6 +11,36 @@ with advance notice in the **Deprecations** section of releases.
 
 .. towncrier release notes start
 
+v3.14.1 (2019-11-13)
```
asottile: should this be 3.15 given the dep bump and py34 drop?

gaborbernat: the py34 is automatic via python-requires; so don't think so 😄 and similarly with the dep bump it's a bug fix, so no need major IMHO

asottile: sg!

PR #1462, opened 2019-11-15
use get_python_info for consistency and caching

blueyed:
I've came across this when looking at the code - it obviously needs some cleanup, but passes all tests and got me a cache hit at least with a typical run.  (IIRC it saves 0.2+s at least due to this).

Let me know if you like it and I/we can finish it.


obestwalter:
Thanks! Seems like a no brainer to me :)

blueyed:
ok, added a changelog.

gaborbernat:

PR #1471, opened 2019-12-02
fix colouring on windows

### LINKED ISSUE #1468: Color support on Windows: TerminalWriter (PY_COLORS), colorama (PYCHARM_HOSTED)
While running several tests with `tox` on GitHub Actions environments, tox's output is not coloured by default (see actions/starter-workflows#114). Setting `PY_COLORS: 1` (#1009) works for jobs running on GNU/Linux (`ubuntu-latest`). Unfortunately, it seems not to be effective on `windows-latest`.

- `ubuntu-latest`: https://github.com/1138-4EB/vunit/runs/326860648#step:5:29
- `windows-latest`: https://github.com/1138-4EB/vunit/runs/326860617#step:7:31

Note that, in those examples, the output of the commands executed by tox is coloured because `--color=yes` is explicitly passed to `pytest` through `{posargs}`.

1138-4EB:
Ref #1468

Currently, `colorama` is required in order to get coloured output on Windows, even if it is not used. This is because of:

https://github.com/pytest-dev/py/blob/1e99d20f31ef511fc2f1a9e49ba970acad441d4e/py/_io/terminalwriter.py#L16-L24

Since neither tox nor TerminalWriter do install colorama explicitly, it falls back to `win32_and_ctypes`, which does not work. In this PR, colorama is added to `install_requires`, for it to be installed on Windows only.

At first, I was not convinced about adding the dependency, because it is not required for minimal usage of tox (i.e., without colours). However, I saw that it will be included for all platforms in #1400, so I guess that it is acceptable.

### DIFF
--- a/docs/changelog/1471.bugfix.rst
+++ b/docs/changelog/1471.bugfix.rst
@@ -1,0 +1,1 @@
+Fix colouring on windows: colorama is a dep. - by :user:`1138-4EB`
--- a/setup.cfg
+++ b/setup.cfg
@@ -46,6 +46,7 @@ install_requires =
     virtualenv >= 16.0.0
     toml >=0.9.4
     filelock >= 3.0.0, <4
+    colorama >= 0.4.1 ;platform_system=="Windows"
 
 [options.packages.find]
 where = src


src/tox/session/commands/help.py
```diff
@@ -10,5 +10,8 @@ def show_help(config):
         "TOX_TESTENV_PASSENV: space-separated list of extra environment variables to be "
         "passed into test command environments"
     )
-    reporter.line("PY_COLORS: 0 disable colorized output, 1 enable (default)")
     reporter.line("TOX_PARALLEL_NO_SPINNER: 1 disable spinner for CI, 0 enable (default)")
+    reporter.line(
+        "tox depends on py.io.TerminalWriter to handle colouring, which supports forcing "
```
gaborbernat: I don't think we want to disclose this implementation behaviour as a report, please remove this change 👍

1138-4EB: Done.

gaborbernat:

PR #1476, opened 2019-12-03
Relax importlib_metadata pin

chyzzqo2:


gaborbernat:
Can demonstrate with an use case why this is needed?

chyzzqo2:
You don't actually require version 1.1. Everything works fine with older versions, in my case using 0.23. There are other libraries that actually break with version > 1. Meaning I either need to pin tox back for my builds or get pip to ignore your pin. So while I can work around this, I still think that it's best to only pin when you actually need to because of breaking APIs or other dependency requirements. Having said that, it's your project so if you disagree feel free to close this and I'll work around the issues it causes.

gaborbernat:
If you're running into compatibility issues with other packages I would recommend using https://github.com/pipxproject/pipx, that installs tox within its own virtual environment avoiding such issues.

PR #1482, opened 2019-12-07
Correct statement on PEP-518

nuno-andre:
As explicitly stated in [PEP-518](https://www.python.org/dev/peps/pep-0518/#build-system-table):

> _Tools should not require the existence of the `[build-system]` table. A `pyproject.toml` file may be used to store configuration details other than build-related data and thus lack a `[build-system]` table legitimately. It is up to the tool to decide how to handle this case._

Also removed _(if you want to avoid an extra file)_ since is not uncommon for a `pyproject.toml` to have the configuration of a single tool that otherwise would have its own file.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -65,6 +65,7 @@ Monty Taylor
 Morgan Fainberg
 Nick Douma
 Nick Prendergast
+Nuno André Novo
 Oliver Bestwalter
 Pablo Galindo
 Paweł Adamczak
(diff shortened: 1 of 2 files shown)


gaborbernat:
I don't think tox follows that advise, or pip for that matter. I'm pretty sure everything would break without it. Furthermore tox is encouraging best practices, and having the build part defined is definitely encouraged and best practice. So I think the current statement is the good one.

asottile:
I'm -1 as well, the claim that pyproject.toml is "the standard for Python tools" simply isn't correct.  It is a specced standard for build tools yes, but that doesn't mean every python tool must put their configuration there

di:
I think this should probably be revisited, I filed https://github.com/tox-dev/tox/issues/2429 to start a new conversation.

PR #1493, opened 2020-01-08
Allow to config the behavior on CTRL+C

sileht:
By default, we run SIGKILL after 0.5 seconds. Most of the time is
enough. But if the interrupted command have a complex processes tree,
it might not be enough to propagate the signal.

In such case processes are left behind and never killed.
If theses processes use static network port or keep file open. Next
call of tox will fail until the all processes left behind are manually
killed.

This change adds some configuration to be able to config the timeout
before signals are sent.

If the approach work for you, I will polish the PR (doc+test)

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/docs/changelog/1493.feature.rst
+++ b/docs/changelog/1493.feature.rst
@@ -1,0 +1,1 @@
+Add ``interrupt_timeout`` and ``terminate_timeout`` that configure delay between SIGINT, SIGTERM and SIGKILL when tox is interrupted. - by :user:`sileht`
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -54,6 +54,9 @@ Import hookimpl directly from tox instead.
 
 WITHIN_PROVISION = os.environ.get(str("TOX_PROVISION")) == "1"
 
+INTERRUPT_TIMEOUT = 0.3
+TERMINATE_TIMEOUT = 0.2
+
 
 def get_plugin_manager(plugins=()):
     # initialize plugin manager
@@ -799,6 +802,20 @@ def tox_addoption(parser):
     parser.add_testenv_attribute_obj(DepOption())
 
     parser.add_testenv_attribute(
+        name="interrupt_timeout",
+        type="float",
+        default=INTERRUPT_TIMEOUT,
+        help="timeout before sending SIGTERM after SIGINT",
+    )
+
+    parser.add_testenv_attribute(
+        name="terminate_timeout",
+        type="float",
+        default=TERMINATE_TIMEOUT,
+        help="timeout before sending SIGKILL after SIGTERM",
+    )
+
+    parser.add_testenv_attribute(
         name="commands",
         type="argvlist",
         default="",
@@ -1231,7 +1248,16 @@ class ParseIni(object):
         for env_attr in config._testenv_attr:
             atype = env_attr.type
             try:
-                if atype in ("bool", "path", "string", "dict", "dict_setenv", "argv", "argvlist"):
+                if atype in (
+                    "bool",
+                    "float",
+                    "path",
+                    "string",
+                    "dict",
+                    "dict_setenv",
+                    "argv",
+                    "argvlist",
+                ):
                     meth = getattr(reader, "get{}".format(atype))
                     res = meth(env_attr.name, env_attr.default, replace=replace)
                 elif atype == "basepython":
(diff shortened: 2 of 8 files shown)


src/tox/session/__init__.py
```diff
@@ -37,6 +37,9 @@
 from .commands.show_config import show_config
 from .commands.show_env import show_envs
 
+WAIT_INTERRUPT = 0.3
```
gaborbernat: why are these not using the config values?

src/tox/config/__init__.py
```diff
@@ -798,6 +798,20 @@ def develop(testenv_config, value):
 
     parser.add_testenv_attribute_obj(DepOption())
 
+    parser.add_testenv_attribute(
+        name="interrupt_timeout",
+        type="float",
+        default=0.3,
```
gaborbernat: shouldn't the defaults come from the sessions constants?

sileht: I used the constants here.

src/tox/session/__init__.py
sileht: Here, `Action()` is created for internal purpose in the legacy builder, and the config values are only available in the envconfig.

docs/changelog/1493.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``interrupt_timeout`` and ``terminate_timeout`` that configure delay between SIGINT, SIGTERM and SIGKILL when tox is interrupted. - by :user:`sileht`
```
gaborbernat: you can use the `` :conf:`interrupt_timeout ` `` to make them references please do so

sileht: done

gaborbernat:
Thanks!

stefanholek:
Hi People,

I have been dealing with a similar issue the other day, and my solution is different. 😎 

I have a test runner which install unittests's [signal handlers](https://docs.python.org/3/library/unittest.html#signal-handling). When hitting Ctrl+C in tox, the test runner sometimes receives a SIGINT while in the process of shutting down. This seemed wrong, and I traced it down to tox unconditionally forwarding any received SIGINT. The situation however is that _all_ foreground processes receive the SIGINT, not just tox, and when tox forwards it the test runner receives the SIGINT twice. My solutions is to make tox wait for processes to go away on their own before starting to send signals.

I see you are talking about SIGKILL, but maybe that's just a side-effect of tox SIGINT'ing immediately? Also, my issue is not fixed by the changes in #1493.

Anyway, here is my patch (which no longer applies now, but you get the idea):

```diff
diff --git a/src/tox/action.py b/src/tox/action.py
index 10707b4..9f826a0 100644
--- a/src/tox/action.py
+++ b/src/tox/action.py
@@ -18,6 +18,7 @@ from tox.reporter import Verbosity
 from tox.util.lock import get_unique_file
 from tox.util.stdlib import is_main_thread
 
+WAIT_SUICIDE = 0.5
 WAIT_INTERRUPT = 0.3
 WAIT_TERMINATE = 0.2
 
@@ -177,7 +178,7 @@ class Action(object):
     def handle_interrupt(self, process):
         """A three level stop mechanism for children - INT -> TERM -> KILL"""
         msg = "from {} {{}} pid {}".format(os.getpid(), process.pid)
-        if process.poll() is None:
+        if self._wait(process, WAIT_SUICIDE) is None:
             self.info("KeyboardInterrupt", msg.format("SIGINT"))
             process.send_signal(signal.CTRL_C_EVENT if sys.platform == "win32" else signal.SIGINT)
             if self._wait(process, WAIT_INTERRUPT) is None:
@@ -193,7 +194,7 @@ class Action(object):
         if sys.version_info >= (3, 3):
             # python 3 has timeout feature built-in
             try:
-                process.communicate(timeout=WAIT_INTERRUPT)
+                process.communicate(timeout=timeout)
             except subprocess.TimeoutExpired:
                 pass
         else:
```

sileht:
#1493 does not change anything about signal handling by default. But just allow to configure different timeouts.

I got your double SIGINT issue. But my issue was a bit different.
My tooling don't care about receiving SIGINT/SIGTERM manytimes, any additional signals are just ignored during the processes shutdown procedure. My patch was to allow to increase the delay between the first SIGINT and the final deadly SIGKILL, so my test runner can cleanly shutdown everything, other signals sent between are ignored by my tooling.

I think your issue/use-case is valid too, feel free to open an issue to track it.

stefanholek:
Ok, thanks.

PR #1519, opened 2020-02-12
Avoid incompatible six

ssbarnea:
Avoids problem introduced by newer virtualenv which broke installion
of tox on systems that already had a six version >1.0 installed but
which was older than minimal needed by tox.

Avoids installing tox with a combination of virtualenv-six that is broken, mainly introduced by newer virtualenv which dropped vendored six and which requires six>=1.14.0. 

Pip fails to install correct dependency because we list six already as a dependecy. We cannot drop it but we can bump it to assure we avoid ending with incompatible setup.

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/docs/changelog/1519.bugfix.rst
+++ b/docs/changelog/1519.bugfix.rst
@@ -1,0 +1,2 @@
+Bump minimal six version needed to avoid using one incompatible with newer
+virtualenv. - by :user:`ssbarnea`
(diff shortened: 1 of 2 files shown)


ssbarnea:
@gaborbernat I do see few CI failures which seem fully unrelated to my change. Are they known and not counting for final result?

gaborbernat:
They are known but need to be fixed for merges. Can be this or another pr 🤷‍♂️ the reason I did not do this myself is that I don't have time at the moment to fix the ci.

PR #1521, opened 2020-02-12
fix pypy CI failure

ssbarnea:
Fixes CI error seen with other patches.

### DIFF
--- a/src/tox/session/commands/run/parallel.py
+++ b/src/tox/session/commands/run/parallel.py
@@ -35,6 +35,7 @@ def run_parallel(config, venv_dict):
 
         def run_in_thread(tox_env, os_env, processes):
             output = None
+            print_out = None
             env_name = tox_env.envconfig.envname
             status = "skipped tests" if config.option.notest else None
             try:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -55,7 +55,12 @@ jobs:
     coverage:
       with_toxenv: 'coverage' # generate .tox/.coverage, .tox/coverage.xml after test run
       for_envs: [py38, py37, py36, py35, py27, pypy3, pypy]
-
+    before:
+      - task: UsePythonVersion@0
+        condition: and(succeeded(), in(variables['TOXENV'], 'pypy'))
+        displayName: provision pypy 3
+        inputs:
+          versionSpec: 'pypy3'
 - ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/') }}:
   - template: publish-pypi.yml@tox
     parameters:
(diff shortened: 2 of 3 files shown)


ssbarnea:
@gaborbernat While current PR is fixing one bug it does not fix the test failure. I need your help regarding what needs to be done regarding https://dev.azure.com/toxdev/tox/_build/results?buildId=2318&view=ms.vss-test-web.build-test-results-tab&runId=1025542&resultId=100172&paneView=debug

Mainly the test fails because no warning is displayed.

src/tox/session/commands/run/parallel.py
```diff
@@ -75,7 +75,7 @@ def collect_process(process):
                 elif status is not None:
                     outcome = spinner.fail
                 outcome(env_name)
-                if print_out and output is not None:
+                if vars().get("print_out", None) and output is not None:
```
asottile: I don't think this is a correct solution, probably just need to move the definition of `print_out` outside of the try block

gaborbernat:
I'm not sure why it fails, would need to debug to understand why it's not triggering 👍

src/tox/session/commands/run/parallel.py
ssbarnea: Sure. The hard question is about the missing warning, which causes that specific test to fail.

ssbarnea:
@gaborbernat @asottile I was able to narrow down the bug, it is CI specific and happens if you do not also have pypy3 installed when trying to run pypy environment. This unique test will fail, but it should be skipped.

It is easy to replicate locally as I didn't have any of the pypy versions installed. I installed pypy and run `tox -e pypy`, failed like on CI with this single test. After this I installed pypy3 and run the tests again, this time it passed.

I think that test should be change to be skipped when pypy3 executable is not found.

gaborbernat:
This can be done as here https://github.com/pypa/virtualenv/blob/master/azure-pipelines.yml#L97-L101

ssbarnea:
Maybe you can explain me why https://github.com/pypa/virtualenv/blob/master/azure-pipelines.yml#L102-L106 lines did not also install pypy3. Based on what I read there, that task was supposed to also add the pypy3 near the default pypy2.

Anyway, I made a change to the test to do an xfail when both executable are not found, as that is the only case where the conflict warning is expected to happen.

ssbarnea:
🙀 green!

ssbarnea:
@gaborbernat Please make a suggestion/edit with the alternative implementation as I am used to that discovery logic. I just want to make CI green ASAP to merge and release https://github.com/tox-dev/tox/pull/1519 -- as tox itself is affected by the virtualenv20 issues. Since I did not do anything else, and I seen a dozen others around me doing the same.

gaborbernat:
Will address this in a few hours.

gaborbernat:
Sigh pypy3 is very-very-very slow... 😢

PR #1525, opened 2020-02-16
Remove Python 3.4 from the supported Python version list

rpkilby:
Removes Python 3.4 from the supported Python versions list. refs #1483, #1456

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,7 +29,6 @@ classifiers =
     Programming Language :: Python :: 2
     Programming Language :: Python :: 2.7
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.4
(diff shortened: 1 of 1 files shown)


asottile:
thanks!

gaborbernat:
Wait I did not want to remove this support here yet...

asottile:
oh uh, I assumed because it was already excluded from python_requires

gaborbernat:
Nm 🤷‍♂️

rpkilby:
Yeah, I ran into this while updating tests for tox-venv. Reading #1483, it doesn't look like the intent was to drop support for 3.4, but it was (at least temporarily) dropped in #1456 / v3.14.1 to fix test suite breakages.

PR #1537, opened 2020-03-16
Exclude virtualenvs vers w/ regressions 20.0.[0-7]

webknjaz:
virtualenv 20 series has had a number of regressions. This PR explicitly excludes versions with known issues that are hard to debug.

## Contribution checklist:

- [ ] added/updated test(s)
      in message body

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,7 +43,7 @@ install_requires =
     pluggy >= 0.12.0, <1
     py >= 1.4.17, <2
     six >= 1.14.0, <2  # required when virtualenv>=20
-    virtualenv >= 16.0.0
+    virtualenv >= 16.0.0, !=20.0.0, !=20.0.1, !=20.0.2, !=20.0.3, !=20.0.4, !=20.0.5, !=20.0.6, !=20.0.7
     toml >=0.9.4
     filelock >= 3.0.0, <4
     colorama >= 0.4.1 ;platform_system=="Windows"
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -43,7 +43,7 @@ install_requires =
     pluggy >= 0.12.0, <1
     py >= 1.4.17, <2
     six >= 1.14.0, <2  # required when virtualenv>=20
-    virtualenv >= 16.0.0
+    virtualenv >= 16.0.0, !=20.0.0, !=20.0.1, !=20.0.2, !=20.0.3, !=20.0.4, !=20.0.5, !=20.0.6, !=20.0.7
```
gaborbernat: Why is this needed? Shouldn't install just pick the latest?

asottile: additionally, though each of these have bugs they "work" for the 95-99% case and it's probably not helpful to exclude it for those already using those versions

webknjaz: It'll only install latest if no existing virtualenv is present. So if you do `pip install tox` and there's `virtualenv==20.0.2` there, it'll not upgrade it because it matches the requirement.

webknjaz: I can't think of a better solution atm

gaborbernat: 😢 you're right

gaborbernat:
released via https://pypi.org/project/tox/3.14.6/

PR #1542, opened 2020-03-25
Fix parsing of architecture in python interpreter name

bruchar1:
The wrong group was captured in the regexp, causing the arch part of the spec to include the dash.
It prevented selecting the architecture for the Python interpreted (e.g. python3.7-32).


gaborbernat:

heads up

PR #1529, opened 2020-02-19
Fix ValueError: max() arg is an empty sequence on

jquast:
Closes #1343. When running `tox -l` with a tox.ini file not containing any default environments, instead of an empty output, an exception is raised:

```
$ tox -l
Traceback (most recent call last):
  File "/Users/jq/.pyenv/versions/py38/bin/tox", line 8, in <module>
    sys.exit(cmdline())
  File "/Users/jq/.pyenv/versions/3.8.1/envs/py38/lib/python3.8/site-packages/tox/session/__init__.py", line 44, in cmdline
    main(args)
  File "/Users/jq/.pyenv/versions/3.8.1/envs/py38/lib/python3.8/site-packages/tox/session/__init__.py", line 68, in main
    exit_code = session.runcommand()
  File "/Users/jq/.pyenv/versions/3.8.1/envs/py38/lib/python3.8/site-packages/tox/session/__init__.py", line 189, in runcommand
    self.showenvs(all_envs=False, description=show_description)
  File "/Users/jq/.pyenv/versions/3.8.1/envs/py38/lib/python3.8/site-packages/tox/session/__init__.py", line 293, in showenvs
    show_envs(self.config, all_envs=all_envs, description=description)
  File "/Users/jq/.pyenv/versions/3.8.1/envs/py38/lib/python3.8/site-packages/tox/session/commands/show_env.py", line 14, in show_envs
    max_length = max((len(env) for env in (default + extra)))
ValueError: max() arg is an empty sequence
```

This change resolves the issue by setting the value of `max_length` to 0 when default and extra are empty.

PR #1543, opened 2020-03-25
Fix changelog for 3.14.6

asottile:


.pre-commit-config.yaml
```diff
@@ -38,3 +38,11 @@ repos:
     rev: v1.4.2
     hooks:
     -   id: rst-backticks
+-   repo: local
+    hooks:
+    -   id: changelogs-rst
+        name: changelog filenames
+        language: fail
+        entry: 'changelog files must be named ####.(bugfix|feature|deprecation|breaking|doc|misc).rst'
+        exclude: ^docs/changelog/(\d+\.(bugfix|feature|deprecation|breaking|doc|misc).rst|README.rst|template.jinja2)
```
webknjaz, line 47: FTR chronographer was showing the red status: https://github.com/tox-dev/tox/pull/1529/checks?check_run_id=527182252. It uses the `pyproject.toml` as a source of truth for the allowed types...

asottile: yes, but it's been pretty noisy and doesn't give people readable signal so it's unfortunately gotten ignored and slipped through -- not really your fault though, github's check results is not the most readable output :(

webknjaz: Well, I mean there's an area for improvements there even on my side.

PR #1544, opened 2020-03-26
Fix exeception when command is empty

bruchar1:
Under some circumstances (e.g. commands = {posargs} ), tox try to run an empty command.
It causes an exception when looking for the first character of an empty string.


gaborbernat:

heads up

PR #1545, opened 2020-03-26
expand section names

bruchar1:
This is a proposition to allow generative section names.

In conjecture with factors, it will allow to write more succinct config files, when using factors in the default [testenv] is not sufficient.

Let me know if you like it.

### DIFF
--- a/docs/changelog/1545.feature.rst
+++ b/docs/changelog/1545.feature.rst
@@ -1,0 +1,1 @@
+Allow generative section name expansion. - by :user:`bruchar1`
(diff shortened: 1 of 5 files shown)


bruchar1:
For `tox -l`, it should not change anything, as this is determined by envlist.
For `tox -a`, it seems to work as expected, since sections are generated just after reading the ini file, before any parsing.
It shouldn't break non-generative sections, as they are left unchanged.
And all tests pass.

bruchar1:
I added documentation. Feel free to suggest improvements. I'm not just sure want to put for the version number.

gaborbernat:

heads up

PR #1556, opened 2020-04-18
Fix CI

gaborbernat:
- drop python3.5
- os.linesep correct


ssbarnea:
@gaborbernat Where is this CI running because I can only see 3 jobs an dnone of them seems to be what I was expecting.

gaborbernat:
Check then auctions tab, probably a GitHub bug 🤷‍♂️

Issue #1418, opened 2019-09-09
tox -p without number should enable auto mode


ssbarnea:
It would a good improvement in user experience if when adding just `-p`, tox would run in parallel mode, like it does not with `tox -p auto`. This would save 5 keystrokes when calling tox.

I think that `auto` is likely what most people want when they want to run in parallel, so why to ask the to type more for getting it.

awecx:
This would be consistent with makefile `-j` where, if no argument is given, would select a number of parallel recipes to build itself.

JulienPalard:
> This would be consistent with makefile `-j` where, if no argument is given, would select a number of parallel recipes to build itself.

I'm not sure for every make implemn, but quoting the GNU make manpage:

> If the -j  option  is  given  without an argument, make will not limit the number of jobs that can run simultaneously.

Which looks more like `all` than `auto`.

(This typically make `make -j` a bit violent when compiling cpython.)

I'd say `all` is not a sane default, and defaulting to `auto` looks a better choice.

While thinking about it, I **bet** (it's only a bet, so don't mind me) `auto` can be faster than `all` as both run the same quantity of instructions, but in the `all` case there may be more context switches (or worse, swapping).

gaborbernat:
Feel free to open a PR against master and fix it. My available efforts at the moment are aimed at fixing this as part of https://github.com/tox-dev/tox/issues/1394, but that probably will take a while (ETA September).

gaborbernat:
Hello, this now has been released via https://pypi.org/project/tox/3.15.0/ 

![heads up](https://media.tenor.co/images/0118006f847acfcfd22e8bf0fd405e42/tenor.gif)

PR #1559, opened 2020-04-20
Allow parallel mode without arguments

ssbarnea:
Allows calling tox with `tox -p` in order to use auto mode. Previously
it was mandatory to mention `auto` in order to enable parallel mode.

This will not change behavior of tox when called without `-p`, which
will still keep the non-parallel mode.

Fixes: #1418
Signed-off-by: Sorin Sbarnea <ssbarnea@redhat.com>

### DIFF
--- a/src/tox/config/parallel.py
+++ b/src/tox/config/parallel.py
@@ -45,9 +45,11 @@ def add_parallel_flags(parser):
     parser.add_argument(
         "-p",
         "--parallel",
+        nargs="?",
+        const="auto",
         dest="parallel",
         help="run tox environments in parallel, the argument controls limit: all,"
-        " auto - cpu count, some positive number, zero is turn off",
+        " auto or missing argument - cpu count, some positive number, 0 to turn off",
         action="store",
         type=parse_num_processes,
         default=DEFAULT_PARALLEL,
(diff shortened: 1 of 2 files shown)


src/tox/config/parallel.py
```diff
@@ -45,9 +45,11 @@ def add_parallel_flags(parser):
     parser.add_argument(
         "-p",
         "--parallel",
+        nargs="?",
+        const="auto",
         dest="parallel",
         help="run tox environments in parallel, the argument controls limit: all,"
-        " auto - cpu count, some positive number, zero is turn off",
+        " auto or missing argument - cpu count, some positive number, 0 to turn off",
```
rpkilby, line 52: I would append `(default: auto)` to the end of the help string, which is more consistent with other arguments. e.g.

```
--workdir PATH                   tox working directory (default: None)
```

ssbarnea: It would be incorrect to do this because the default is `0` which is also added automatically by argparse libary. There is a bit of confusion beween the two types of defaults: the default when opion is not mentioned which is 0, and the default value when the option is mentioned (auto), known as `const`.  Maybe in the future we may want to change default to be auto, but this change has serious implications and should be posponed till a major version change. Mainly, if you want parallel, you must add at least `-p` for the moment.

rpkilby: ah, thanks, I was conflating the two.

PR #1562, opened 2020-04-21
Passenv PIP_EXTRA_INDEX_URL by default

### LINKED ISSUE #1561: Pass PIP_EXTRA_INDEX_URL by default
tox currently passes `PIP_INDEX_URL` as a default environment config. This parameter allows the user to specify a package index.

pip supports another configuration parameter for additional indexes; `PIP_EXTRA_INDEX_URL`.

I think it would be beneficial to pass `PIP_EXTRA_INDEX_URL` by default in addition to `PIP_INDEX_URL` to support index configuration parameters transparently

georgealton:
This Pull Request passes the `PIP_EXTRA_INDEX_URL` environment variable to tox environments by default. Allowing pip index configuration to pass to tox environments without having to set `passenv` configuration. 

I believe this is in spirit with the current `PIP_INDEX_URL` default passing behaviour. 

This closes #1561

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -168,7 +168,7 @@ Here is a simple overview, with tox-specific bits:
 
     $ tox -e dev
 
-   To get information about all environements, type:
+   To get information about all environments, type:
 
    $ tox -av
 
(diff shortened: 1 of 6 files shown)


ssbarnea:
Is there any reason why we would not want to whitelist `PIP_*` as pip has a very long list of options. This should also be more future proof.

PR #1579, opened 2020-05-11
Fix CI due to latest pip output change

gaborbernat:


hroncok:
Would you mind if this was changed to work with older pip as well?

gaborbernat:
Sure send the PR.

hroncok:
https://github.com/tox-dev/tox/pull/1594

PR #1588, opened 2020-05-21
Add option to delay propegating SIGINT to child process 

### LINKED ISSUE #1497: Immediately forwarding Ctrl+C to child processes considered harmful
I have a test runner which install unittests's [signal handlers](https://docs.python.org/3/library/unittest.html#signal-handling). When hitting Ctrl+C in tox, the test runner sometimes receives a SIGINT while in the process of shutting down. This seemed wrong, and I traced it down to tox unconditionally forwarding any received SIGINT. The situation however is that _all_ foreground processes receive the SIGINT, not just tox, and when tox forwards it the test runner receives the SIGINT twice. My solution is to make tox wait for processes to go away on their own before starting to send signals.

Anyway, here is my patch (which no longer applies after #1493, but you get the idea):

```diff
diff --git a/src/tox/action.py b/src/tox/action.py
index 10707b4..9f826a0 100644
--- a/src/tox/action.py
+++ b/src/tox/action.py
@@ -18,6 +18,7 @@ from tox.reporter import Verbosity
 from tox.util.lock import get_unique_file
 from tox.util.stdlib import is_main_thread
 
+WAIT_SUICIDE = 0.5
 WAIT_INTERRUPT = 0.3
 WAIT_TERMINATE = 0.2
 
@@ -177,7 +178,7 @@ class Action(object):
     def handle_interrupt(self, process):
         """A three level stop mechanism for children - INT -> TERM -> KILL"""
         msg = "from {} {{}} pid {}".format(os.getpid(), process.pid)
-        if process.poll() is None:
+        if self._wait(process, WAIT_SUICIDE) is None:
             self.info("KeyboardInterrupt", msg.format("SIGINT"))
             process.send_signal(signal.CTRL_C_EVENT if sys.platform == "win32" else signal.SIGINT)
             if self._wait(process, WAIT_INTERRUPT) is None:
@@ -193,7 +194,7 @@ class Action(object):
         if sys.version_info >= (3, 3):
             # python 3 has timeout feature built-in
             try:
-                process.communicate(timeout=WAIT_INTERRUPT)
+                process.communicate(timeout=timeout)
             except subprocess.TimeoutExpired:
                 pass
         else:
```

jhesketh:
Fixes #1497

Ctrl+C sends a SIGTERM to both tox and the running process, causing the
running process to receive the signal twice once tox passes it along.
This can often cancel or interfer with the process's cleanup.

Instead, add an option to allow a process to suicide before sending the
SIGTERM.

Co-Authored-By: Stefan H. Holek <stefan@epy.co.at>

Also includes a minor fix for correcting the location of the documentation for the existing timeout options: The interrupt_timeout and terminate_timeout options are set on the environment, not globally.

## Contribution checklist:
- [x] added/updated test(s)

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -160,22 +160,6 @@ Global settings are defined under the ``tox`` section as:
     Name of the virtual environment used to create a source distribution from the
     source tree.
 
-.. conf:: interrupt_timeout ^ float ^ 0.3
-
-    .. versionadded:: 3.15.0
-
-    When tox is interrupted, it propagates the signal to the child process,
-    waits :conf:``interrupt_timeout`` seconds, and sends it a SIGTERM if it hasn't
-    exited.
-
-.. conf:: terminate_timeout ^ float ^ 0.2
-
-    .. versionadded:: 3.15.0
-
-    When tox is interrupted, it propagates the signal to the child process,
-    waits :conf:``interrupt_timeout`` seconds, sends it a SIGTERM, waits
-    :conf:``terminate_timeout`` seconds, and sends it a SIGKILL if it hasn't exited.
-
 Jenkins override
 ++++++++++++++++
 
@@ -596,6 +580,22 @@ Complete list of settings that you can put into ``testenv*`` sections:
        ``depends`` does not pull in dependencies into the run target, for example if you select ``py27,py36,coverage``
        via the ``-e`` tox will only run those three (even if ``coverage`` may specify as ``depends`` other targets too -
        such as ``py27, py35, py36, py37``).
+
+.. conf:: interrupt_timeout ^ float ^ 0.3
+
+    .. versionadded:: 3.15.0
+
+    When tox is interrupted, it propagates the signal to the child process,
+    waits :conf:``interrupt_timeout`` seconds, and sends it a SIGTERM if it hasn't
+    exited.
+
+.. conf:: terminate_timeout ^ float ^ 0.2
+
+    .. versionadded:: 3.15.0
+
+    When tox is interrupted, it propagates the signal to the child process,
+    waits :conf:``interrupt_timeout`` seconds, sends it a SIGTERM, waits
+    :conf:``terminate_timeout`` seconds, and sends it a SIGKILL if it hasn't exited.
 
 Substitutions
 -------------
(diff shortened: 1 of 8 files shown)


docs/config.rst
```diff
@@ -160,22 +160,6 @@ Global settings are defined under the ``tox`` section as:
     Name of the virtual environment used to create a source distribution from the
     source tree.
 
-.. conf:: interrupt_timeout ^ float ^ 0.3
```
gaborbernat, line 163: Why move these?

jhesketh: This PR bundles up two commits. I can separate them out if needed, but it is explained in the earlier message:

> Move timeout config options into correct section
> 
> The interrupt_timeout and terminate_timeout options are set on the
> environment, not globally.

jhesketh: (I'll update the PR message at least)

jhesketh:
So the failing test (`test_parallel_interrupt`) is marked as flaky, and will fail (even without my patch) if you put a sleep at the beginning of `Action.handle_interrupt`. That likely explains the flakiness, but I am unsure where in the test it is expecting things to be cleaned up so quickly (adding sleeps or checking the processes has not helped in my debugging, so I'm probably missing something).

jheld:
Are the master tests consistently passing?

gaborbernat:
Almost always.

jheld:
I'm confused. The docs suggest that tox already supports something like this.

interrupt_timeout, terminate_timeout.

Is that not what's being asked for?

jhesketh:
> I'm confused. The docs suggest that tox already supports something like this.
> 
> interrupt_timeout, terminate_timeout.
> 
> Is that not what's being asked for?

The problem is that Ctrl+C will send a SIGINT to the foreground processes. This means that tox receives it along with pytest (or whatever it may be). Because tox then sends a SIGINT to the child processes it receives it twice. This goes unnoticed in most cases, but it often means that while pytest is cleaning up it receives another SIGINT and then aborts the cleanup.

This patch gives the running process a chance to exit before sending a SIGINT. If it hasn't exited, it'll receive a SIGINT from tox (which may be the first one it has received, or the second).

jhesketh:
So I've set the default suicide timeout to 0.

This ensures that the behaviour has not changed from previously, and that programs get (at least one) SIGINT immediately. This also fixes the flaky test that appear to be sensitive to timing.

PR #1601, opened 2020-06-16
Add alias "allowlist_externals" to "whitelist_externals"

### LINKED ISSUE #1491: Editorial: consider offering alternative to whitelist_externals
Using whitelist (and its counterpart) isn't exactly welcoming. While migrating away from it entirely might be hard due to compatibility considerations, it'd be nice if configuration files from projects that build on this technology could avoid it.

See https://tools.ietf.org/html/draft-knodel-terminology for additional context.

dajose:
closes #1491

### DIFF
--- a/docs/changelog/1491.deprecation.rst
+++ b/docs/changelog/1491.deprecation.rst
@@ -1,0 +1,1 @@
+Add allowlist_externals alias to whitelist_externals (whitelist_externals is now deprecated). - by :user:`dajose`
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -325,15 +325,20 @@ Complete list of settings that you can put into ``testenv*`` sections:
 
     Don't set this option if your :conf:`install_command` does not use pip.
 
-.. conf:: whitelist_externals ^ MULTI-LINE-LIST
+.. conf:: allowlist_externals ^ MULTI-LINE-LIST
 
     Each line specifies a command name (in glob-style pattern format)
     which can be used in the ``commands`` section without triggering
     a "not installed in virtualenv" warning.  Example: if you use the
-    unix ``make`` for running tests you can list ``whitelist_externals=make``
-    or ``whitelist_externals=/usr/bin/make`` if you want more precision.
+    unix ``make`` for running tests you can list ``allowlist_externals=make``
+    or ``allowlist_externals=/usr/bin/make`` if you want more precision.
     If you don't want tox to issue a warning in any case, just use
-    ``whitelist_externals=*`` which will match all commands (not recommended).
+    ``allowlist_externals=*`` which will match all commands (not recommended).
+
+    .. note::
+
+        ``whitelist_externals`` has the same meaning and usage as ``allowlist_externals``
+        but it is now deprecated.
 
 .. conf:: changedir ^ PATH ^ {toxinidir}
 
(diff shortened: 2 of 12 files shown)


dajose:
I think the macOS checks failure is not related with the PR, can someone give me a hand there?

src/tox/config/__init__.py
```diff
@@ -747,8 +747,13 @@ def passenv(testenv_config, value):
         "eventual passenv setting.",
     )
 
+    # FIXME remove whitelist in favor of allowlist_external
     parser.add_testenv_attribute(
-        name="whitelist_externals",
+        name="whitelist_externals", type="line-list", help="DEPRECATED: use allowlist_external",
```
pawamoy, line 752: Missing `s` at the end.

src/tox/config/__init__.py
```diff
@@ -747,8 +747,13 @@ def passenv(testenv_config, value):
         "eventual passenv setting.",
     )
 
+    # FIXME remove whitelist in favor of allowlist_external
```
pawamoy, line 750: (Missing `s` at the end)

docs/changelog/1491.deprecation.rst
```diff
@@ -0,0 +1 @@
+Whitelist (as used on whitelist_externals) has a racial weight, allowlist_externals should be preferred. - by :user:`dajose`
```
gaborbernat, line 1: This needs to expand on the fact that whitelist_externals is still supported, however now the allowlist_externals alias has been added.

docs/config.rst
```diff
@@ -325,15 +325,15 @@ Complete list of settings that you can put into ``testenv*`` sections:
...
     Each line specifies a command name (in glob-style pattern format)
     which can be used in the ``commands`` section without triggering
     a "not installed in virtualenv" warning.  Example: if you use the
-    unix ``make`` for running tests you can list ``whitelist_externals=make``
-    or ``whitelist_externals=/usr/bin/make`` if you want more precision.
+    unix ``make`` for running tests you can list ``allowlist_externals=make``
+    or ``allowlist_externals=/usr/bin/make`` if you want more precision.
     If you don't want tox to issue a warning in any case, just use
-    ``whitelist_externals=*`` which will match all commands (not recommended).
+    ``allowlist_externals=*`` which will match all commands (not recommended).
```
gaborbernat, line 336: This needs to still mention that whitelist_external is still is accepted.

src/tox/config/__init__.py
gaborbernat: We don't use fixmes in this project but instead prefer the issue tracker, please remove this. We could not remove it either way before the next major version as that would be breaking change.

tests/unit/test_venv.py
```diff
@@ -433,6 +433,10 @@ def test_install_command_not_installed(newmocksession):
 
 
 def test_install_command_whitelisted(newmocksession):
+    """
+    Deprecated: Remove in favor of test_install_command_allowlisted when
```
gaborbernat, line 437: Remove this docstring, the migration complete will not happen before the rewrite.

src/tox/venv.py
```diff
@@ -223,7 +223,7 @@ def is_allowed_external(self, p):
         if tox.INFO.IS_WIN:
             tryadd += [os.path.normcase(x) for x in os.environ["PATHEXT"].split(os.pathsep)]
             p = py.path.local(os.path.normcase(str(p)))
-        for x in self.envconfig.whitelist_externals:
+        for x in self.envconfig.allowlist_externals + self.envconfig.whitelist_externals:
```
gaborbernat, line 226: I'm not happy to allow specifying both of them. We should instead allow only one, and raise if both set. And then pick only the one that is set.

dajose:
@gaborbernat can you give me a hand on why it is failing? 

I think it is not related with the changes: https://dev.azure.com/toxdev/tox/_build/results?buildId=2498&view=logs&j=6f785b48-ed19-5349-e6cb-bb6d29a19c8b&t=4621fbdb-f74d-5491-ece0-e6cd5b96369c&l=704

asottile:
re-triggered CI, seems to be doing better now

PR #1604, opened 2020-06-24
Remove usage of threads for parsing ini

brettlangdon:
The GIL prevents us from using threads to optimize this process. So we end up paying the costs of threading coordination without any of the benefits.

On small config files this makes almost no difference in performance, but for large config files it does.

For our large config file (1199 envs) it drops runtime by about 1.5s.

```
$ python ./src/tox/__main__.py -c ~/example/tox.ini -l | wc -l
1199

$ time python ./src/tox/__main__.py -c ~/example/tox.ini -l > /dev/null
python ./src/tox/__main__.py -c ~/example/tox.ini -l > /dev/null  8.39s user 1.18s system 109% cpu 8.712 total

$ time python ./src/tox/__main__.py -c ~/example/tox.ini -l > /dev/null
python ./src/tox/__main__.py -c ~/example/tox.ini -l > /dev/null  6.82s user 0.40s system 99% cpu 7.258 total
```

## Contribution checklist:

- [ ] added/updated test(s)
      in message body
  - Added in a previous approved PR

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -13,7 +13,6 @@ import warnings
 from collections import OrderedDict
 from fnmatch import fnmatchcase
 from subprocess import list2cmdline
-from threading import Thread
 
 import pluggy
 import py
@@ -1147,16 +1146,8 @@ class ParseIni(object):
                 known_factors.update(env.split("-"))
 
         # configure testenvs
-        to_do = []
         failures = OrderedDict()
         results = {}
-        cur_self = self
-
-        def run(name, section, subs, config):
-            try:
-                results[name] = cur_self.make_envconfig(name, section, subs, config)
-            except Exception as exception:
-                failures[name] = (exception, traceback.format_exc())
 
         order = []
         for name in all_envs:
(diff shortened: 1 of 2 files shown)


src/tox/config/__init__.py
```diff
@@ -1147,17 +1146,10 @@ def line_of_default_to_zero(section, name=None):
                 known_factors.update(env.split("-"))
 
         # configure testenvs
-        to_do = []
         failures = OrderedDict()
         results = {}
         cur_self = self
```
asottile, line 1151: can also probably remove this aliasing too now that it's unused

was this threaded maybe to avoid some subprocess overhead?

brettlangdon: will tox open any subprocesses during parsing?

asottile: I believe it does (or used to) to figure out the python versions

gaborbernat:
Add a configuration that needs envsitepackages and we call subprocess to figure that out. This case is much more common than 1k+ environments, so this optimization actually would slow us down in general.

PR #1614, opened 2020-07-13
Implement support for PEP517 backend-path setting

### LINKED ISSUE #1575: Support PEP517 in-tree build backend
I'm getting
```python-traceback
ERROR: invocation failed (exit code 1), logfile: ~/src/github/ansible/pylibssh/.tox/.package/log/.package-2.log
=================================== log start ====================================
  File "~/.pyenv/versions/3.7.1/lib/python3.7/site-packages/tox/helper/build_requires.py", line 7, in <module>
    backend = __import__(backend_spec, fromlist=[None])
ModuleNotFoundError: No module named 'pep517_backend'
```
because of
```console
$ cat pyproject.toml                            
[build-system]
...
backend-path = ["bin"]  # requires 'Pip>=20' or 'pep517>=0.6.0'
build-backend = "pep517_backend"
```
and
```console
$ ls -l bin/pep517_backend.py      
-rw-r--r-- 1 me me 6.2K May  5 00:43 bin/pep517_backend.py
```

I think this should be considered a bug. Pip supports it starting v20 and pep517 supports it with v0.7 or so.

webknjaz:
This change implements support for `backend-path` in the `[build-system]` section of
`pyproject.toml` as described @ https://www.python.org/dev/peps/pep-0517/#in-tree-build-backends.

Resolves #1575.

## Contribution checklist:

- [ ] added/updated test(s)
      in message body

### DIFF
--- a/src/tox/helper/build_isolated.py
+++ b/src/tox/helper/build_isolated.py
@@ -8,6 +8,24 @@ tarball.
 import os
 import sys
 
+
+def _ensure_module_in_paths(module, paths):
+    """Verify that the imported backend belongs in-tree."""
+    if not paths:
+        return
+
+    module_path = os.path.normcase(os.path.abspath(module.__file__))
+    normalized_paths = (os.path.normcase(os.path.abspath(path)) for path in paths)
+
+    if any(os.path.commonprefix((module_path, path)) == path for path in normalized_paths):
+        return
+
+    raise SystemExit(
+        "build-backend ({!r}) must exist in one of the paths "
+        "specified by backend-path ({!r})".format(module, paths),
+    )
+
+
 dist_folder = sys.argv[1]
 backend_spec = sys.argv[2]
 backend_obj = sys.argv[3] if len(sys.argv) >= 4 else None
@@ -16,6 +34,7 @@ backend_paths = sys.argv[4].split(os.path.pathsep) if sys.argv[4] else []
 sys.path[:0] = backend_paths
 
 backend = __import__(backend_spec, fromlist=["_trash"])
+_ensure_module_in_paths(backend, backend_paths)
 if backend_obj:
     backend = getattr(backend, backend_obj)
 
--- a/src/tox/helper/build_isolated.py
+++ b/src/tox/helper/build_isolated.py
@@ -1,8 +1,19 @@
+"""PEP 517 build backend invocation script.
+
+It accepts externally parsed build configuration from `[build-system]`
+in `pyproject.toml` and invokes an API endpoint for building an sdist
+tarball.
+"""
+
+import os
 import sys
 
 dist_folder = sys.argv[1]
 backend_spec = sys.argv[2]
 backend_obj = sys.argv[3] if len(sys.argv) >= 4 else None
+backend_paths = sys.argv[4].split(os.path.pathsep) if sys.argv[4] else []
+
+sys.path[:0] = backend_paths
 
 backend = __import__(backend_spec, fromlist=["_trash"])
 if backend_obj:
--- a/src/tox/package/builder/isolated.py
+++ b/src/tox/package/builder/isolated.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import json
+import os
 from collections import namedtuple
 
 import six
@@ -11,7 +12,9 @@ from tox import reporter
 from tox.config import DepConfig, get_py_project_toml
 from tox.constants import BUILD_ISOLATED, BUILD_REQUIRE_SCRIPT
 
-BuildInfo = namedtuple("BuildInfo", ["requires", "backend_module", "backend_object"])
+BuildInfo = namedtuple(
+    "BuildInfo", ["requires", "backend_module", "backend_object", "backend_paths"],
+)
 
 
 def build(config, session):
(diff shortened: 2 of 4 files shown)


src/tox/helper/build_isolated.py
```diff
@@ -3,6 +3,9 @@
 dist_folder = sys.argv[1]
 backend_spec = sys.argv[2]
 backend_obj = sys.argv[3] if len(sys.argv) >= 4 else None
+backend_paths = sys.argv[4].split(':')
```
webknjaz, line 6: @gaborbernat should I serialize this as JSON instead?

gaborbernat: I think it's fine, though you might want to check here if len is at least 4 👍

src/tox/helper/build_isolated.py
```diff
@@ -3,6 +3,9 @@
 dist_folder = sys.argv[1]
 backend_spec = sys.argv[2]
 backend_obj = sys.argv[3] if len(sys.argv) >= 4 else None
+backend_paths = sys.argv[4].split(":")
+
+sys.path[:0] = backend_paths
```
gaborbernat, line 8: wouldn't this be optional?

src/tox/package/builder/isolated.py
```diff
@@ -103,6 +124,7 @@ def perform_isolated_build(build_info, package_venv, dist_dir, setup_dir):
                 str(dist_dir),
                 build_info.backend_module,
                 build_info.backend_object,
+                ":".join(str(p) for p in build_info.backend_paths),
```
gaborbernat, line 127: can we use instead the os specific variable here?

src/tox/helper/build_isolated.py
webknjaz: @gaborbernat this script is always called with a fixed number of args in only one place in code. I'd say that the check above is also redundant.

My concern with `:` delimiter is that it's not escaped and paths having it would break things.

src/tox/helper/build_isolated.py
webknjaz: `''.split(':')` returns `[]` and `sys.path[:0] = []` is noop

src/tox/package/builder/isolated.py
webknjaz: Ah, fair enough. Though, `:` in paths is unescapable per POSIX using the standard `$PATH` delimiter.

src/tox/helper/build_isolated.py
gaborbernat: quite cryptic behaviour instead of being explicit if you ask me

src/tox/package/builder/isolated.py
```diff
@@ -47,6 +50,17 @@ def build(config, session):
     return perform_isolated_build(build_info, package_venv, config.distdir, config.setupdir)
 
 
+def _ensure_importable_in_path(python_path, importable):
+    """Figure out if the importable exists in the given path."""
+    mod_chunks = importable.split(".")
+    pkg_path = python_path.join(*mod_chunks[:-1])
+    if not pkg_path.exists():
+        return False
+    if pkg_path.join(mod_chunks[-1], "__init__.py").exists():
```
gaborbernat, line 59: You should check here for pyc too. If you delete ``__init__.py`` but keep ``__init__.pyc`` the import will still work. Same below.

src/tox/package/builder/isolated.py
```diff
@@ -47,6 +50,17 @@ def build(config, session):
...
+        return True
+    return pkg_path.join(mod_chunks[-1] + ".py").exists()
```
gaborbernat, line 61: use '{}.py'.format instead of +

src/tox/package/builder/isolated.py
webknjaz: I'm not sure if this is reasonable: the build backend is supposed to be committed along with the source and so it'd be a bad idea to encourage checking-in pre-compiled files... WDYT?

src/tox/helper/build_isolated.py
webknjaz: I just wanted to keep it clean. If you think about it, it's pretty straightforward. The code basically says "inject any extra build paths in front of `sys.path`". And it shouldn't matter whether it's one path, two or none — the interface is the same.

src/tox/package/builder/isolated.py
gaborbernat: Checking in the compiled files might be used/useful for obfuscation of the build-backend and to save space, not?

gaborbernat: Just a thought, but I consider this point nice to have. The rest though are non-negotiable.

webknjaz: Well, I get the same impression as if I were to check-in a binary executable honestly... Though, it's not explicitly forbidden in PEP.

webknjaz: Let's check what `pep517` does and follow the lead. Shall we?

src/tox/helper/build_isolated.py
gaborbernat: Might be cleaner but if you ask me it's more cryptic because you have to think about it, why it's correct rather than just be obvious. Not?

src/tox/package/builder/isolated.py
gaborbernat: Sure.

webknjaz: So `pypa/pep517` does not check the importable on the preparation stage (https://github.com/pypa/pep517/blob/3f1949a/pep517/wrappers.py#L131-L135) but it does so later after actually importing the backend (https://github.com/pypa/pep517/blob/3f1949a/pep517/_in_process.py#L90-L95) by checking `__file__` against those paths.

Also, I'd like to point out that PEP's backend-path section implies that it facilitates the src code use-case explicitly.

> In certain circumstances, projects may wish to include *the source code* for the build backend directly in the source tree

Given this, I'm not sure what to do. Any preferences? I should probably drop this part of the check here.

webknjaz: @gaborbernat I've unresolved this thread (you managed to close it while I went out to search for clues in `pep517` and in the PEP text). I'd like to agree on what's better and I'm not completely against your idea, just trying to be as precise as possible wrt to what PEP says.

gaborbernat: I'm happy to do just as much as pep517 does here.

webknjaz:
Hey! I was going to polish it... This PR needs more tests. Will have to send them separately, I guess...

PR #1616, opened 2020-07-14
Use a HOSTNAME environment variable instead of socket.getfqdn() in ResultLog

hroncok:
This makes it possible to prevent an undesired DNS query.

Fixes https://github.com/tox-dev/tox/issues/1615

## Thanks for contributing a pull request!

If you are contributing for the first time or provide a trivial fix don't worry too
much about the checklist - we will help you get started.

## Contribution checklist:

(also see [CONTRIBUTING.rst](../tree/master/CONTRIBUTING.rst) for details)

- [x] wrote descriptive pull request text
- [x] added/updated test(s)
- [ ] updated/extended the documentation
- [x] added relevant [issue keyword](https://help.github.com/articles/closing-issues-using-keywords/) in message body
- [x] added news fragment in [changelog folder](../tree/master/docs/changelog)
- [x] added yourself to `CONTRIBUTORS` (preserving alphabetical order)

Not quite sure where to document this behavior.

PR #1618, opened 2020-07-14
Fix tests when HOSTNAME is set, but empty string

hroncok:
See https://github.com/tox-dev/tox/pull/1616#issuecomment-658437312

This is a fixup of cfe66fd0fe8a81bd65f79f589486b7721632a4cb.

- [x] added/updated test(s)

### DIFF
--- a/tests/unit/test_result.py
+++ b/tests/unit/test_result.py
@@ -1,3 +1,4 @@
+import functools
 import os
 import signal
 import socket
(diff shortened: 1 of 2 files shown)


tests/unit/test_result.py
```diff
@@ -20,7 +20,7 @@ def create_fake_pkg(tmpdir):
 @pytest.fixture()
 def no_hostname_envvar():
     hostname = os.getenv("HOSTNAME")
```
asottile, line 22: perhaps a better alternative is to use pytest's builtin `monkeypatch.setenv` ?

hroncok: Right, I forgot it exists. I can get to it tomorrow, I'm already away from keyboard.

hroncok: I've rebased the PR to use it, thanks for the suggestion.

hroncok:
Thanks for the suggestion and merge!

gaborbernat:
This now is done under 3.17.1, thanks for the contribution!

PR #1619, opened 2020-07-15
release 3.17.1

gaborbernat:


docs/changelog.rst
```diff
@@ -11,6 +11,16 @@ with advance notice in the **Deprecations** section of releases.
...
 .. towncrier release notes start
 
+v3.17.1 (2020-07-15)
+--------------------
+
+Bugfixes
+^^^^^^^^
+
+- Fix tests when the ``HOSTNAME`` environment variable is set, but empty string - by :user:`hroncok`
+  `#1616 <https://github.com/tox-dev/tox/issues/1616>`_
```
asottile, line 21: the change here was test-only, I don't think there needs to be another release

gaborbernat: I think it's needed because otherwise, Linux distributions wanting to repackage can't release.

Issue #1621, opened 2020-07-17
[docs][level:easy] "tty" substitution example doesn't show up as a code block


webknjaz:
I noticed that the code block @ https://tox.readthedocs.io/en/latest/config.html#interactive-shell-substitution doesn't show up as code, like other similar examples above.

The fix should be easy — `s/:$/::/` @ https://github.com/tox-dev/tox/blob/3c10fd9/docs/config.rst#L722.

I'm not sending a PR hoping that this would be a good first issue for beginners.

PR #1622, opened 2020-07-17
Fix `tty` substitution to displays as code block

slsh1o:
It's trivial fix so I hope there is no need to follow up the checklist.
See: #1621


webknjaz:
@gaborbernat mind merging? This doesn't need a change fragment.

PR #1637, opened 2020-07-30
Fix `:user:` role syntax in the #1631 change note

webknjaz:


docs/changelog/1631.doc.rst
```diff
@@ -1 +1 @@
-Fixed grammar in top-level documentation. :user:tfurf
+Fixed grammar in top-level documentation - by :user:`tfurf`
```
webknjaz, line 1: Hey @tfurf, here's the correct syntax for RST roles FYI ^

tfurf: Thanks! A pre-commit hook would be nice addition here.

webknjaz: Yeah... Not sure if it should be in pre-commit or in the bot tho.

Issue #1045, opened 2018-10-08
do not use setup.cfg file that does not have tox:tox namespace


williamjamir:
Hi,

I would like to suggest a check to ensure that a `tox.ini` file is present when calling the `tox` command.

I'm saying this because by mistake I've named the file `tox.ini` as `.tox.ini` and took me a time to figure it out the problem.

The build had five stages:
```
py27, py36, py37, linting, docs
```
[Link for the Travis build with all stages listed](https://travis-ci.org/williamjamir/test-cookiecutter/builds/438427625)

The stages py27, py36, py37 was marked as "successful", even though nothing was tested.
[Link for the job py37](https://travis-ci.org/williamjamir/test-cookiecutter/jobs/438427637)

While the stages `linting` and `docs` were raising an `Error`
```
ERROR: unknown environment 'linting'
```

I believe that this type of information leads to misinterpretation since the user could believe that the configuration is ok (because the py* was marked as successful) and the problem is related with the "linting/docs" stage.

gaborbernat:
@williamjamir at the moment if we fail to find a configuration fail we stop with error and the message:

```
tox
ERROR: tox config file (either pyproject.toml, tox.ini, setup.cfg) not found
```

If you did not get it means you most probably had a tox.ini file discovered by following the rules from https://tox.readthedocs.io/en/latest/config.html#configuration-discovery. 

Furthermore, if you call tox with a verbose flag you'll see explicitly stated what config is used:

```
tox -v
using tox.ini: /home/bernat/git/tox/tox.ini
using tox-3.5.0 from /home/bernat/.local/lib/python3.7/site-packages/tox/__init__.py
```

The tools/checks are already there I think, you just need to use it. Let me know if you need anything else.

williamjamir:
Thanks for the answer @gaborbernat =)
I've added the flag `-v` and as you have mentioned, another file was being used as the configuration for `tox`.

```
$ tox -v
using tox.ini: /home/travis/build/williamjamir/test-cookiecutter/setup.cfg
ERROR: unknown environment 'linting'
```

On the documentation the following note is added:
> Note that setup.cfg requires the content to be under the tox:tox section. 

Perhaps the existence of the `tox:tox` section inside the `setup.cfg` should not be verified as well?

gaborbernat:
Indeed it should, not sure how trivial it is to added will have a look. Feel free to shoot a pr if you have time for it.

hroncok:
[...]

PR #1642, opened 2020-08-06
Do not use setup.cfg file that does not have tox:tox namespace

hroncok:
Do not use setup.cfg file that does not have tox:tox namespace

When tox is invoked without a tox configuration file (and without command line
options that could be used instead), it fails with:

    $ tox
    ERROR: tox config file (either pyproject.toml, tox.ini, setup.cfg) not found

This is useful: If there is no config, test didn't run and such error should
not pass silently.

However, until this commit, when there was a setup.cfg file it was considered
as a tox config even when there was no `[tox:tox]` section:

    $ tox
    GLOB sdist-make: .../setup.py
    python create: .../.tox/python
    python inst: ...
    python installed: ...
    python run-test-pre: PYTHONHASHSEED='1234567890'
      python: commands succeeded
      congratulations :)

That is not safe, because setup.cfg can exist even if there is no tox
configuration in it and the success creates a false sense of safety.

To fix this and to match the behavior of pyproject.toml without a tox section,
tox now skips setup.cfg, if there is no `[tox:tox]` section.

Fixes https://github.com/tox-dev/tox/issues/1045

## Contribution checklist:

- [x] added/updated test(s)

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -271,7 +271,10 @@ def parseconfig(args, plugins=()):
                 content = toml_content["tool"]["tox"]["legacy_tox_ini"]
             except KeyError:
                 continue
-        ParseIni(config, config_file, content)
+        try:
+            ParseIni(config, config_file, content)
+        except SkipThisIni:
+            continue
         pm.hook.tox_configure(config=config)  # post process config object
         break
     else:
@@ -1045,6 +1048,10 @@ def make_hashseed():
     return str(random.randint(1, max_seed))
 
 
+class SkipThisIni(Exception):
+    """Internal exception to indicate the parsed ini file should be skipped"""
+
+
 class ParseIni(object):
     def __init__(self, config, ini_path, ini_data):  # noqa
         config.toxinipath = ini_path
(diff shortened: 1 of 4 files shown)


asottile:
heheh I actually noticed this behaviour recently and considered it a feature (at least for `tox --devenv` which _just works_ even without tox configuration)

will try this branch out and poke around a bit 👍

hroncok:
Well, if it is useful to allow devenv without configuration, let's do that explicitly and not just with setup.cfg?

src/tox/config/__init__.py
```diff
@@ -1045,13 +1048,22 @@ def make_hashseed():
     return str(random.randint(1, max_seed))
 
 
+class SkipThisIni(Exception):
```
gaborbernat, line 1051: I'd probably derive this from ValueError as the file is present but we're unhappy with its content, rather than directly from the exception.

hroncok: Actually, I wanted to use `ValueError` directly, but there is another `ValueError` raise by this `__init__` a bit bellow and I didn't want to interfere with it.

hroncok: Oh, I thought this conversation is still going on. All good?

gaborbernat: Was a slight suggestion, not deal-breaker.

hroncok: Understood.

hroncok:
@asottile https://github.com/tox-dev/tox/issues/1643

PR #1647, opened 2020-08-06
Fix: accept hyphens in generative section names

### LINKED ISSUE #1636: Generative section names don't work with empty factor
If I have a `tox.ini` like this:

```ini
[tox]
skipsdist = true

[testenv:foo]
commands =
    {envpython} -c "print('Hello World')"

[testenv:foo-factor]
commands =
    {envpython} -c "print('Factor given')"
    {envpython} -c "print('Hello World')"
```

Now I'd like to simplify this to use generative section names. I'd expect to either `testenv:foo{,-factor}` or `testenv:foo-{,factor}` to work. It looks like neither does.

This:

```ini
[tox]
skipsdist = true

[testenv:foo,{-factor}]
commands =
    factor: {envpython} -c "print('Factor given')"
    {envpython} -c "print('Hello World')"
```

results in:

```
$ tox -e foo-factor
ERROR: unknown environment 'foo-factor'
```

(or an empty run without any commands if `foo-factor` is added to `envlist`), while this:

```ini
[tox]
skipsdist = true

[testenv:foo-{,factor}]
commands =
    factor: {envpython} -c "print('Factor given')"
    {envpython} -c "print('Hello World')"
```

results in:

```
$ tox -e foo       
ERROR: unknown environment 'foo'
```

(as it defines an environment called `foo-`, not `foo`).

Am I missing something, or is the currently no way to do this kind of thing?

Tyagdit:
Generative section names wouldn't work if there were hyphens within the braces. the regex caused everything after the hyphen to be omitted resulting in not being able to use empty factors like `[testenv:foo{,-factor}]` and forcing users to put the hyphens outside the braces.

Fixes #1636 

## Contribution checklist:

- [x] added/updated test(s)

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -94,6 +94,7 @@ Stephen Finucane
 Sviatoslav Sydorenko
 Thomas Grainger
 Tim Laurence
+Tyagraj Desigar
 Ville Skyttä
 Xander Johnson
 anatoly techtonik
--- a/docs/changelog/1636.bugfix.rst
+++ b/docs/changelog/1636.bugfix.rst
@@ -1,0 +1,1 @@
+Allow hyphens and empty factors in generative section name. - by :user:`tyagdit`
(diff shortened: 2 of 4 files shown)


The-Compiler:
Yay! Thanks @Tyagdit and @gaborbernat :tada:

PR #1668, opened 2020-09-01
Support environment files and comments in setenv

### LINKED ISSUE #1667: Support environment files within setenv
👍

gaborbernat:
Resolves #1667.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -404,6 +404,14 @@ Complete list of settings that you can put into ``testenv*`` sections:
         setenv   =
             PYTHONPATH = {env:PYTHONPATH}{:}{toxinidir}
 
+    .. versionadded:: 3.20
+
+    Support for comments. Lines starting with ``#`` are ignored.
+
+    Support for environment files. Lines starting with the ``file|`` contain path to a environment
+    file to load. Rules within the environment file are the same as within the ``setenv``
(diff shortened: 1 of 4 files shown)


gaborbernat:
Released via https://pypi.org/project/tox/3.20.0/

docs/config.rst
```diff
@@ -404,6 +404,14 @@ Complete list of settings that you can put into ``testenv*`` sections:
         setenv   =
             PYTHONPATH = {env:PYTHONPATH}{:}{toxinidir}
 
+    .. versionadded:: 3.20
+
+    Support for comments. Lines starting with ``#`` are ignored.
+
+    Support for environment files. Lines starting with the ``file|`` contain path to a environment
```
webknjaz, line 411: @gaborbernat any specific reason you implemented it with a prefix vs a substitution? If I were to design this, I'd probably go for `{dotenv:file_path}` there... I'm not judging, it just seems untypical for tox's UX.

gaborbernat: Didn't want to worry about conflicts with conditional factors 🤷‍♂️

gaborbernat: And it's not really substitution at all happening here, but different load mechanism. Substitution implies we just string replace that expression with content of the file, but actually here we never do that. Conceptually.

PR #1683, opened 2020-10-05
Fix negated factor condition example for not-a-and-not-b

markferry:
test_factor_ops makes it clear that `!a,!b` is "not-a OR not-b" and that
the "not-a AND not-b" (implied by the docs) should be `!a-!b`.

## Contribution checklist:

- [ ] added/updated test(s)
      in message body


gaborbernat (commented):
I think the current format is the correct one, and not the new proposed one.

markferry:
I wasn't explicit enough. The problem is that this:
```
!py34,!py36: enum34
```
still adds enum34 as a dependency when using *either* python3.4 *or* python3.6.
(Because `py34` matches `!py36`, and `py36` matches `!py34`!)

...which is why the example should be corrected to `!py34-!py36`.
(e.g. add enum34 as a dependency when not using *neither*)

gaborbernat (approved):

Issue #1690, opened 2020-10-17
Curly braces cannot be escaped in setenv definition since 3.20.0


jayvdb:
Somewhat similar to https://github.com/tox-dev/tox/issues/1502 , but this is a regression caused by https://github.com/tox-dev/tox/pull/1656 , which has a test which is precisely what is *not* wanted.  It appears there wasnt an issue created about that problem, but it appears to be that the runtime environment for the invoked commands contains backslashes.   That should have been solved by leaving the tox substitution framework working as-is, and post-processing the substitutions when they are injected into the new process environment, to whatever extent that is possible (seems feasible to me, but maybe I have missed something).

The following emits `\{a\}`
```ini
[testenv]
setenv =
  FOO=\\\{a\\\}
commands =
  echo "{env:FOO}"
```

Removing the `\\` results in an startup error.
```py
[testenv]
setenv =
  FOO=\{a\}
commands =
  echo "{env:FOO}"
```

```py
using tox.ini: /home/jayvdb/projects/tox/my-tox-tests/tests/tox.ini (pid 27501)
  removing /home/jayvdb/projects/tox/my-tox-tests/tests/.tox/log
Traceback (most recent call last):
  File "/home/jayvdb/.cache/tox/my-tox-tests/py38-tox3_20/bin/tox", line 8, in <module>
    sys.exit(cmdline())
  File "/home/jayvdb/.cache/tox/my-tox-tests/py38-tox3_20/lib/python3.8/site-packages/tox/session/__init__.py", line 44, in cmdline
    main(args)
  File "/home/jayvdb/.cache/tox/my-tox-tests/py38-tox3_20/lib/python3.8/site-packages/tox/session/__init__.py", line 65, in main
    config = load_config(args)
  File "/home/jayvdb/.cache/tox/my-tox-tests/py38-tox3_20/lib/python3.8/site-packages/tox/session/__init__.py", line 81, in load_config
    config = parseconfig(args)
  File "/home/jayvdb/.cache/tox/my-tox-tests/py38-tox3_20/lib/python3.8/site-packages/tox/config/__init__.py", line 278, in parseconfig
    ParseIni(config, config_file, content)
  File "/home/jayvdb/.cache/tox/my-tox-tests/py38-tox3_20/lib/python3.8/site-packages/tox/config/__init__.py", line 1249, in __init__
    raise tox.exception.ConfigError(
[...]

PR #1691, opened 2020-10-17
Fix escaping brackets in setenv

jayvdb:
Fixs regression in v3.20.0 that caused escaped curly braces in setenv to break usage of the variable elsewhere in tox.ini.

Fixes https://github.com/tox-dev/tox/issues/1690

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -390,9 +390,7 @@ class SetenvDict(object):
                 return os.environ.get(name, default)
             self._lookupstack.append(name)
             try:
-                res = self.reader._replace(val)
-                res = res.replace("\\{", "{").replace("\\}", "}")
-                self.resolved[name] = res
+                self.resolved[name] = res = self.reader._replace(val)
             finally:
                 self._lookupstack.pop()
             return res
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -408,6 +408,19 @@ class SetenvDict(object):
         self.definitions[name] = value
         self.resolved[name] = value
 
+    def items(self):
+        return ((name, self[name]) for name in self.definitions)
+
+    def export(self):
+        # post-process items to avoid internal syntax/semantics
+        # such as {} being escaped using \{\}, suitable for use with
+        # os.environ .
+        return {
+            name: Replacer._unescape(value)
+            for name, value in self.items()
+            if value is not self._DUMMY
+        }
+
 
 @tox.hookimpl
 def tox_addoption(parser):
@@ -1785,6 +1798,10 @@ class Replacer:
 
         return expanded
 
+    @staticmethod
+    def _unescape(s):
+        return s.replace("\\{", "{").replace("\\}", "}")
+
     def _replace_match(self, match):
         g = match.groupdict()
         sub_value = g["substitution_value"]
@@ -1924,7 +1941,7 @@ class _ArgvlistReader:
                 new_arg = ""
                 new_word = reader._replace(word)
                 new_word = reader._replace(new_word)
-                new_word = new_word.replace("\\{", "{").replace("\\}", "}")
+                new_word = Replacer._unescape(new_word)
                 new_arg += new_word
                 newcommand += new_arg
         else:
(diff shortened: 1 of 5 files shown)


jayvdb:
The macOS CI error appears to be unrelated.

src/tox/config/__init__.py
```diff
@@ -410,6 +408,19 @@ def __setitem__(self, name, value):
         self.definitions[name] = value
         self.resolved[name] = value
 
+    def items(self):
+        return [(name, self.__getitem__(name)) for name in self.definitions]
```
asottile, line 412:
```suggestion
        return ((name, self[name]) for name in self.definitions)
```

1. more closely match the python3 `.items()` api
2. don't call double-double-under methods

src/tox/config/__init__.py
```diff
@@ -410,6 +408,19 @@ def __setitem__(self, name, value):
...
+
+    def export(self):
+        # post-process items to avoid internal syntax/semantics
+        # such as {} being escaped using \{\}, suitable for use with
+        # os.environ .
+        return {
+            name: re.sub(r"\\({|})", r"\1", value)
```
asottile, line 419: regex is overkill here, I'd just use `.replace` -- it looks like the original code did this why did you change it? (the original code is much more readable imo)

tests/unit/test_venv.py
```diff
@@ -799,6 +803,10 @@ def test_env_variables_added_to_pcall(tmpdir, mocksession, newconfig, monkeypatc
         assert env is not None
         assert "ENV_VAR" in env
         assert env["ENV_VAR"] == "value"
+        assert env["ESCAPED_VAR"] == "{value}"
+        assert env["ESCAPED_VAR2"] == r"\{value\}"
+        assert env["COLON"] == ";" if sys.platform == "win32" else ":"
+        assert env["TTY_VAR"] == "OFF_VALUE"
```
asottile, line 809: these seem unrelated to the patch

asottile (commented):
it's unclear to me what this is actually fixing, why would you still want the backslashes in the variable?

can you show a concrete usecase?

tests/unit/test_venv.py
jayvdb: these are testing other parts of substitution internals which are not tested in this context.  If `export()` is to work correctly with a stable API, all of the substitution internal values need to be converted into externally understandable values.

For the record, using `{posargs}` in setenv is another case of special handling in the substitution system, and it fails strangely in setenv.

And another internal voodoo not tested is `file|` - I havent tried that yet.

src/tox/config/__init__.py
jayvdb: I didnt pay attention to the original code, as it was in a very different area of the code exercised by different tests.  I'm happy to use `.replace`.

jayvdb:
> why would you still want the backslashes in the variable?

backslashes are how tox escapes `{` so that `{` can be a literal in a variable value.
This is in the docs at https://tox.readthedocs.io/en/latest/config.html#substitutions

If you are asking about literal backslashes in the variable value, there are many reasons people might want them.
https://www.google.com/search?q=backslashes+in+environment+variables

src/tox/config/__init__.py
gaborbernat: Yeah please do.

gaborbernat (changes requested):

tests/unit/test_venv.py
jayvdb: I've tested with `file|` - no problems here, but https://github.com/tox-dev/tox/issues/1694 raised for other issues.

I also retested with `{posargs}`, `{packages}` and `{opts}` and very odd behaviour raised in https://github.com/tox-dev/tox/issues/1695

src/tox/config/__init__.py
jayvdb: This was done.  Please review again when you have time, as this PR merge conflicts with other PRs, and is also a prerequisite for testing in other PRs using brackets in settings and their effect on the runtime environment variables present while running `commands`.

gaborbernat (approved):

Issue #763, opened 2018-02-10
tox fails when running in a path containing a hash


The-Compiler:
On Archlinux with 2.9.1, when trying to run it in a folder called `test#foo`, I get:

```
$ tox -e flake8
flake8 create: /home/florian/proj/qutebrowser/test#foo/.tox/flake8
flake8 installdeps: -r/home/florian/proj/qutebrowser/test#foo/requirements.txt, -r/home/florian/proj/qutebrowser/test#foo/misc/requirements/requirements-flake8.txt
flake8 installed: attrs==17.3.0,colorama==0.3.9,cssutils==1.0.2,flake8==3.5.0,flake8-bugbear==17.12.0,flake8-builtins==1.0.post0,flake8-comprehensions==1.4.1,flake8-copyright==0.2.0,flake8-debugger==3.0.0,flake8-deprecated==1.3,flake8-docstrings==1.1.0,flake8-future-import==0.4.3,flake8-mock==0.3,flake8-per-file-ignores==0.4,flake8-polyfill==1.0.1,flake8-string-format==0.2.3,flake8-tidy-imports==1.1.0,flake8-tuple==0.2.13,Jinja2==2.10,MarkupSafe==1.0,mccabe==0.6.1,pep8-naming==0.4.1,pycodestyle==2.3.1,pydocstyle==2.1.1,pyflakes==1.6.0,Pygments==2.2.0,pyPEG2==2.15.2,PyYAML==3.12,six==1.11.0,snowballstemmer==1.2.1
flake8 runtests: PYTHONHASHSEED='2112495524'
flake8 runtests: commands[0] | /home/florian/proj/qutebrowser/test
ERROR: invocation failed (errno 2), args: ['/home/florian/proj/qutebrowser/test'], cwd: /home/florian/proj/qutebrowser/test#foo
Traceback (most recent call last):
  File "/bin/tox", line 11, in <module>
    load_entry_point('tox==2.9.1', 'console_scripts', 'tox')()
  File "/usr/lib/python3.6/site-packages/tox/session.py", line 40, in main
    retcode = Session(config).runcommand()
  File "/usr/lib/python3.6/site-packages/tox/session.py", line 392, in runcommand
    return self.subcommand_test()
  File "/usr/lib/python3.6/site-packages/tox/session.py", line 583, in subcommand_test
    self.runtestenv(venv)
  File "/usr/lib/python3.6/site-packages/tox/session.py", line 592, in runtestenv
[...]

Issue #924, opened 2018-07-24
shlexing happens _after_ substitution causing space paths to be multiple arguments


asottile:
```ini
[tox]
skipsdist = true

[testenv]
commands =
    python -c 'import sys; print(sys.argv[1:])' -- {toxinidir}
```

In `/private/tmp/t t`:

```
$ tox -e py37 -qq
['--', '/private/tmp/t', 't']
```

Expected:

```
$ tox -e py37 -qq
['--', '/private/tmp/t t']
```

If I quote `{toxinidir}` it works though:

```ini
[tox]
skipsdist = true

[testenv]
commands =
    python -c 'import sys; print(sys.argv[1:])' -- '{toxinidir}'
```

```
$ tox -e py37 -qq
['--', '/private/tmp/t t']
```

Intuitively, I think `{toxinidir}` should probably be considered a specific argument, though if people are already depending on this behaviour it would potentially break their configuration.

Related, if we treat it as an indivisible argument, there would have to probably be some special handling for things like `{toxinidir}/.coveragerc` 🤔

jayvdb:
Confirmed it is still a problem, and is the same underlying problem as #763 .
i.e. using `{envpython}` with a path containing space fails in the same way.
While using `#` in path does fail in more confusing ways, the manual quoting workaround used in this issue, like putting `'{toxinidir}'` in the .ini file, also fixes the problem with `#`.

jayvdb:
There is one difference to #763 that makes spaces 'harder', the word splitter in `config/__init__.py`:`CommandParser.words` adds another level of whitespace handling in addition to the shlex'ing.

masenf:
Requiring users to quote substitutions is essentially how shell scripting has handled this problem: https://www.shellcheck.net/wiki/SC2086

Not saying shell scripting is the epitome of style and safety, but keeping the substitution expansion separate from any sort quoting or splitting makes the configuration language more flexible and composable (albeit adding some footguns to watch for)

PR #1698, opened 2020-10-19
Quote path settings containing # and spaces

jayvdb:
This is limited to settings of type "path".
While it is limited to `#` and ` ` (space), other characters which break `shlex` could be added.

Fixes https://github.com/tox-dev/tox/issues/763 and https://github.com/tox-dev/tox/issues/924

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/docs/changelog/1700.feature.rst
+++ b/docs/changelog/1700.feature.rst
@@ -1,0 +1,1 @@
+Allow {/} to refer to os.sep. - by :user:`jayvdb`
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -62,6 +62,7 @@ _FACTOR_LINE_PATTERN = re.compile(r"^([\w{}\.!,-]+)\:\s+(.+)")
 _ENVSTR_SPLIT_PATTERN = re.compile(r"((?:\{[^}]+\})+)|,")
 _ENVSTR_EXPAND_PATTERN = re.compile(r"\{([^}]+)\}")
 _WHITESPACE_PATTERN = re.compile(r"\s+")
+_UNESCAPED_DOUBLEQUOTE = re.compile(r"((?<!\{1})'){2}")
 
 
 def get_plugin_manager(plugins=()):
@@ -1946,8 +1947,11 @@ class _ArgvlistReader:
                     continue
 
                 new_arg = ""
+                had_dual_quote = re.search(_UNESCAPED_DOUBLEQUOTE, word)
                 new_word = reader._replace(word, unquote_path=False)
                 new_word = reader._replace(new_word, unquote_path=False)
+                if not had_dual_quote:
+                    new_word = re.sub(_UNESCAPED_DOUBLEQUOTE, "'", new_word)
                 new_word = new_word.replace("\\{", "{").replace("\\}", "}")
                 new_arg += new_word
                 newcommand += new_arg
@@ -1982,8 +1986,14 @@ class CommandParser(object):
                     and ps.word
                     and ps.word[-1] not in string.whitespace
                 )
-                or (cur_char == "{" and ps.depth == 0 and not ps.word.endswith("\\"))
-                or (ps.depth == 0 and ps.word and ps.word[-1] == "}")
+                or (
+                    cur_char == "{"
+                    and ps.depth == 0
+                    and not ps.word.endswith("\\")
+                    and ps.word != "'"
+                )
+                or (ps.depth == 0 and ps.word and ps.word[-1] == "}" and peek() != "'")
+                or (ps.depth == 0 and ps.word and ps.word[-2:] == "}'")
                 or (cur_char not in string.whitespace and ps.word and ps.word.strip() == "")
             )
 
@@ -1997,6 +2007,12 @@ class CommandParser(object):
             if word_has_ended():
                 yield_this_word()
 
+        def peek():
+            try:
+                return self.command[_i + 1]
+            except IndexError:
+                return ""
+
         def accumulate():
             ps.word += cur_char
 
@@ -2006,7 +2022,7 @@ class CommandParser(object):
         def pop_substitution():
             ps.depth -= 1
 
-        for cur_char in self.command:
+        for _i, cur_char in enumerate(self.command):
             if cur_char in string.whitespace:
                 if ps.depth == 0:
                     yield_if_word_ended()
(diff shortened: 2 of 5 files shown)


src/tox/config/__init__.py
```diff
@@ -390,7 +391,7 @@ def get(self, name, default=None):
                 return os.environ.get(name, default)
             self._lookupstack.append(name)
             try:
-                res = self.reader._replace(val)
+                res = self.reader._replace(val, unquote_path=False)
```
jayvdb, line 394: How quoted paths should appear in the setenv results is an aspect I havent explored yet, and would likely build on https://github.com/tox-dev/tox/pull/1691

Despite stripping quotes as early as possible, I expect there are quite a few corner cases where the quotes are going to causes new complications, which is why I have ensured the quotes are only used when `#` is in the path, which already didnt work, so corner cases breaking isnt a regression.  Adding ` ` to the quoted path chars is less clear, as that probably worked for many simple cases, and only failed in corner cases as seen at https://github.com/tox-dev/tox/issues/924.

src/tox/config/__init__.py
```diff
@@ -1591,7 +1592,7 @@ def addsubstitutions(self, _posargs=None, **kw):
             self.posargs = _posargs
 
     def getpath(self, name, defaultpath, replace=True):
-        path = self.getstring(name, defaultpath, replace=replace)
+        path = self.getstring(name, defaultpath, replace=replace, unquote_path=True)
         if path is not None:
             toxinidir = self._subs["toxinidir"]
             return toxinidir.join(path, abs=True)
```
jayvdb, line 1598: I was doing the unquoting here, before the join.  It worked well, but didnt feel tight enough.

There is still one aspect which may make this necessary, and at least needs a test case.
In addition to problematic chars in preset paths like `{toxinidir}`, the space or hash might appear in derived paths, such as `{toxinidir}/foo bar.txt`

jayvdb: Added tests for `'{toxinidir}'/'foo bar.txt'` , and I am pretty sure that the unquoting cant occur here.

src/tox/config/__init__.py
```diff
@@ -1895,7 +1915,7 @@ def getargvlist(cls, reader, value, replace=True):
             current_command += line
 
             if is_section_substitution(current_command):
-                replaced = reader._replace(current_command, crossonly=True)
+                replaced = reader._replace(current_command, crossonly=True, unquote_path=False)
```
jayvdb, line 1918: `unquote_path=False` is added liberally to aid reviewing, so all invocations of `_replace(..)` can be seen.
IMO they should removed before being merged.

src/tox/config/__init__.py
```diff
@@ -1924,8 +1944,11 @@ def processcommand(cls, reader, command, replace=True):
                     continue
 
                 new_arg = ""
-                new_word = reader._replace(word)
-                new_word = reader._replace(new_word)
+                has_dual_quote = "''" in word
+                new_word = reader._replace(word, unquote_path=False)
+                new_word = reader._replace(new_word, unquote_path=False)
+                if not has_dual_quote:
+                    new_word = new_word.replace("''", "'")
```
jayvdb, line 1951: this has the potential for changing the behaviour of any existing `tox.ini` which used quotes and also had quotes inserted via substitution.  The most obvious case would be an env var of `'` and commands using the envvar and quotes.  Then, does stripping dual quotes change the behaviour of those previous usages?  In theory, this removal would be simply merging quoted strings.  Need to think more about possible breakages.

jayvdb: Ah, this simple replace would break `\\''`.  Easy fixed with a `re`.

src/tox/config/__init__.py
```diff
@@ -1864,6 +1878,12 @@ def _replace_substitution(self, match):
             val = self._substitute_from_other_section(sub_key)
         if callable(val):
             val = val()
+        if isinstance(val, PathBase):
+            val = str(val)
+            # XXX handle ' and " in paths
```
jayvdb, line 1883: This has partially been done.
I was thinking of switching quoting char to avoid clashes with `'`.  The quoting could even be done with an invalid filename char (e.g. non-printables), as long as it makes it through the regex engine.

tests/unit/config/test_config.py
```diff
@@ -1531,6 +1645,47 @@ def test_rewrite_simple_posargs(self, tmpdir, newconfig):
...
+            [testenv:py27]
+            args_are_paths = True
+            changedir = test#dir
+            commands = cmd1 {posargs:hello}
+        """
+        conf = newconfig([], inisource).envconfigs["py27"]
+        argv = conf.commands
+        assert argv[0] == ["cmd1", "hello"]
+
+        if sys.platform != "win32":
```
jayvdb, line 1659: Noting windows specific behaviour is happening here, cutting the dir at `#`.  May be a bug in test harness code, given that the next stanza is OK.

src/tox/config/__init__.py
jayvdb: `\\''` now fixed.  No test coverage yet.  Other obscure oddities may still exist.

src/tox/config/__init__.py
jayvdb: Tests for setting paths in `setenv` have been added.

Spaces in paths also added.

Merging #1691 first would allow for more exploration of possible edge cases using escaping.

docs/changelog/1700.feature.rst
```diff
@@ -0,0 +1 @@
+Allow {/} to refer to os.sep. - by :user:`jayvdb`
```
jayvdb, line 1: This is https://github.com/tox-dev/tox/pull/1701

Without this, the tests need to do hacks to convert `/` to `\` on Windows.  It gets messy because the bugs related to this are about real native paths, including constructed path based on preset paths.

gaborbernat (changes requested):
Can you rebase this?

jayvdb:
Sure, but now that #1691 is merged (thx), I'd like to also explore how all of these chars appear in the commands subprocess runtime environment variables.
Also now worth verifying that `{` in paths doesnt do anything odd.

jayvdb:
https://github.com/tox-dev/tox/pull/1713 solves many of these issues, in as much as that `type=path` args pass through without escaping except anything inside `{..}`.  There are a few tests from this PR which were omitted, related to round tripping paths through `setenv`, but it feels it can solve the underlying issues much cleaner and clearer (and thus easier to document), but will have some problems unsolved.

gaborbernat:

gaborbernat:

Issue #1708, opened 2020-10-22
Backslash-ed curly bracket in settings


jayvdb:
config.rst states

> You can escape curly braces with the \ character if you need them, for example:
> `commands = echo "\{posargs\}" = {posargs}`

A minor note, the `= {posargs}` should be moved so that it is clearer that it is output, and not part of the `commands`.

The docs don't indicate where this is possible, which led to https://github.com/tox-dev/tox/pull/1656 and https://github.com/tox-dev/tox/issues/1690 to get it working in `setenv` and in the subprocess runtime environment variables.

And there is one very obvious omission from the docs.  Any time an escaping character is introduced, there needs to be a way to escape the escape character.  The typical way to escape `\` is `\\`, and this is de-facto implemented because of use of `shlex`.  https://github.com/tox-dev/tox/pull/1691 added the first test for `\\{` escaping the `\` to become literal `\{`.  `\\` should be mentioned in the docs.  However, `\\` is not implemented consistently; all of the non-shlex parsing ignores it.

e.g. `commands = echo "\\{posargs}"` should be `\` followed by substitution of `{posargs}`.  Instead it results in literal `\{posargs}`.

An additional complication is that `\` is valid in many contexts, e.g.
```ini
[testenv:env\bar]
foo\ = bar
\foo = bar
setenv =
  \ENV_VAR = {[testenv:env\bar]\foo}
  ENV_VAR\ = {[testenv:env\bar]foo\}
commands = echo '{}' {env:\ENV_VAR} {env:ENV_VAR\}
```
`\` is valid in paths on Linux, so the above works fine on Linux.  The above would probably have strange results on Windows as `\` is a path separator.  I think and hope it would be reasonable to prohibit `\` in envname and any paths on all platforms.

https://github.com/tox-dev/tox/pull/1698 goes some of the way to handle paths a bit more intelligently, but there are some very low level problems, and there is at least one case where better docs can help.  Because the implementation uses `shlex`, the behaviour of `\` depends on whether it is within quotes or not.

e.g. the following will put literal `\{` into the setting, and almost every other setting is based on that setting, so that `\{` will appear in many other paths.

```ini
toxworkdir = {toxinidir}/.tox\{dir
```

It also causes subprocess env to contain
```
'TOX_WORK_DIR': '/path/to/tox{foo/tests/.tox\\{dir'
'VIRTUAL_ENV': '/path/to/tox{foo/tests/.tox\\{dir/bracketpath2'
```

[...]

PR #1709, opened 2020-10-22
Bracketed escaping

jayvdb:
Adds `{\}`, `{{}` and `{}}` as an escaping mechanism in settings.

`{{}` and `{}}` are not implemented so they can be used safely together.
Additional coding is needed to handle values from the engine being pushed
back into the engine, so they are re-escaped.

Fixes https://github.com/tox-dev/tox/issues/1708

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/tests/unit/config/test_config.py
+++ b/tests/unit/config/test_config.py
@@ -2803,6 +2803,20 @@ class TestIndexServer:
 
 
 class TestConfigConstSubstitutions:
+    def test_replace_backslash(self, newconfig):
+        r"""Replace {\} with literal \."""
+        config = newconfig(
+            r"""
+            [testenv]
+            setenv =
+                FOO = a{\}b
+            commands = echo {env:FOO} c{\}d
+            """,
+        )
+        envconfig = config.envconfigs["python"]
+        assert envconfig.commands[0] == ["echo", "a\\b", "c\\d"]
+        assert envconfig.setenv["FOO"] == "a\\b"
+
     @pytest.mark.parametrize("pathsep", [":", ";"])
     def test_replace_pathsep(self, monkeypatch, newconfig, pathsep):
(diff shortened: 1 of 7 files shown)


tests/unit/config/test_config.py
```diff
@@ -2803,6 +2803,47 @@ def test_multiple_homedir_relative_local_indexservers(self, newconfig):
...
+        config = newconfig(
+            """
+        [testenv]
+        setenv =
+            FOO = a{{}b c{}}d
+        """,
+        )
+        envconfig = config.envconfigs["python"]
+        # This fails: ConfigError: substitution key 'b c' not found
+        # assert envconfig.setenv["FOO"] == "a{b c}d"
```
jayvdb, line 2845: highlighting this as what fails with the incomplete implementation.

gaborbernat:
> Adds `{\}`, `{{}` and `{}}` as an escaping mechanism in settings.
> 
> `{{}` and `{}}` are not implemented so they can be used safely together.
> Additional coding is needed to handle values from the engine being pushed
> back into the engine, so they are re-escape

I would prefer not to introduce new ways of doing things. As such only ``\`` character I'm interested in supporting, I don't agree with supporting ``{{`` or ``}}`` 🤔

jayvdb:
> I don't agree with supporting `{{` or `}}`

I am also not convinced about adding those; I've tried to lay out the problems at

https://github.com/tox-dev/tox/issues/1708

That commit in this PR is mostly to show where the problem exists, so anyone can play with it if they think it is easy to solve.

gaborbernat (changes requested):
This now has merge conflicts.

gaborbernat:

gaborbernat:

Issue #1571, opened 2020-04-28
Support expanding globs/path wildcards in configuration


webknjaz:
Currently, if you put a command like `delocate-wheel dist/*.whl` in `tox.ini`, that command (`delocate-wheel`) will get `"dist/*.whl"` string as its arg. But if you run it in shell, it'll get the list of matching paths.

Some programs are okay with that and apply `glob` internally while others don't know how to do this. Or, in case of `delocate`, it does it partially but has a bug where it outputs a file with an invalid filename called `*.whl`.

So in some cases, it's reasonable to want to expand a glob explicitly.

I suggest adding a new substitution syntax. This is what I have in mind:
```ini
commands =
  {envpython} -m some-command {glob:dist/*.whl} {glob:**/*.tar.gz:fallback/path/without/substitutions.txt} {env:VAR:{glob:fallback/*.zip}}
```

Refs:
* https://github.com/matthew-brett/delocate/issues/71
* https://github.com/pypa/twine/issues/612

webknjaz:
In my case, I implemented a workaround for my case but it's sort of ugly because it requires running a separate command for this env and doesn't allow taking advantage of `depends` and putting all related envs into one `TOXENV` to let tox figure out the sequence internally:
```ini
commands =
    {envpython} -m \
      delocate.cmd.delocate_wheel \
      -v \
      {posargs:"{env:PEP517_OUT_DIR}"/*.whl}
```

gaborbernat:
I think this is a good idea of a feature, note I personally don't plan to do work on this before tox 4... so if you want it before feel free to put in a PR.

jayvdb:
Checked, and the below does *not* work.  It also results in a literal `*`.  Globbing must be happening in `tox/venv.py`.

```ini
[tox]
distshare = {toxworkdir}/distshare

[testenv:py123]
deps=
  {distshare}/dep1-*

[testenv:deps_glob]
commands = ls {[testenv:py123]deps}
```

`pip` also isnt doing the globbing.
```
pip install '/path/to/.tox/distshare/dep1-*'
Defaulting to user installation because normal site-packages is not writeable
ERROR: Invalid requirement: '/path/to/.tox/distshare/dep1-*'
Hint: It looks like a path. File '/path/to/.tox/distshare/dep1-*' does not exist.
```

I do know of one way to achieve it, using my alpha grade plugin [`tox-backticks`](https://github.com/jayvdb/tox-backticks) which allows the following
```ini
[tox]
[...]

PR #1713, opened 2020-10-23
Replacer: Allow LocalPath to pass through

### LINKED ISSUE #763: tox fails when running in a path containing a hash

jayvdb:
Paths were already stored inside SectionReader._subs as
path objects, however they are turned into strings when they
go through Replacer, and various transforms occur on these
paths, for example breaking paths containing '#' and ' '.

This change defaults to path objects being retained through the
Replacer, and path segments being joined together using path object
joining semantics.

This mode can be disabled for settings of type `path` by wrapping
the value in quotes, and disabled for other values by disabling
new global setting literal_paths.

Fixes #763 and #924

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/docs/changelog/1713.feature.rst
+++ b/docs/changelog/1713.feature.rst
@@ -1,0 +1,1 @@
+Preserve paths through the substitution engine. Adds global setting ``literal_paths`` to disable new behaviour. - by :user:`jayvdb`
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -143,6 +143,19 @@ Global settings are defined under the ``tox`` section as:
     configure :conf:`basepython` in the global testenv without affecting environments
     that have implied base python versions.
 
+.. conf:: literal_paths ^ true|false ^ true
+
+    .. versionadded:: 3.21.0
+
+    tox defaults to interpretting values commencing with a path as a literal path, with
+    only segments inside ``{..}`` being substituted, without any need for quoting.
+
+    Disabling this setting to use shell-like syntax for all values, except settings of
+    type ``path``.
+
+    For settings of type ``path``, shell-like syntax can be activate by commencing the
+    value with a quotation mark.
+
 .. conf:: isolated_build ^ true|false ^ false
 
     .. versionadded:: 3.3.0
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -20,6 +20,7 @@ import py
 import toml
 from packaging import requirements
 from packaging.utils import canonicalize_name
+from py._path.common import PathBase
 
 import tox
 from tox.constants import INFO
@@ -163,6 +164,8 @@ class DepOption:
         deps = []
         config = testenv_config.config
         for depline in value:
+            if isinstance(depline, PathBase):
+                depline = str(depline)
             m = re.match(r":(\w+):\s*(\S+)", depline)
             if m:
                 iname, name = m.groups()
@@ -416,7 +419,7 @@ class SetenvDict(object):
         # such as {} being escaped using \{\}, suitable for use with
         # os.environ .
         return {
-            name: Replacer._unescape(value)
+            name: str(value) if isinstance(value, PathBase) else Replacer._unescape(value)
             for name, value in self.items()
             if value is not self._DUMMY
         }
@@ -1151,6 +1154,8 @@ class ParseIni(object):
             hash_seed = config.option.hashseed
         config.hashseed = hash_seed
 
+        config.literal_paths = reader.getbool("literal_paths", True)
+
         reader.addsubstitutions(toxinidir=config.toxinidir, homedir=config.homedir)
 
         if config.option.workdir is None:
@@ -1364,7 +1369,13 @@ class ParseIni(object):
 
     def make_envconfig(self, name, section, subs, config, replace=True):
         factors = set(name.split("-"))
-        reader = SectionReader(section, self._cfg, fallbacksections=["testenv"], factors=factors)
+        reader = SectionReader(
+            section,
+            self._cfg,
+            fallbacksections=["testenv"],
+            factors=factors,
+            literal_paths=config.literal_paths,
+        )
         tc = TestenvConfig(name, config, factors, reader)
         reader.addsubstitutions(
             envname=name,
@@ -1578,6 +1589,7 @@ class SectionReader:
         factors=(),
         prefix=None,
         posargs="",
+        literal_paths=True,
     ):
         if prefix is None:
             self.section_name = section_name
@@ -1590,6 +1602,7 @@ class SectionReader:
         self._subststack = []
         self._setenv = None
         self.posargs = posargs
+        self.literal_paths = literal_paths
 
     def get_environ_value(self, name):
         if self._setenv is None:
@@ -1602,7 +1615,7 @@ class SectionReader:
             self.posargs = _posargs
 
     def getpath(self, name, defaultpath, replace=True):
-        path = self.getstring(name, defaultpath, replace=replace)
+        path = self.getstring(name, defaultpath, replace=replace, is_path=True)
         if path is not None:
             toxinidir = self._subs["toxinidir"]
             return toxinidir.join(path, abs=True)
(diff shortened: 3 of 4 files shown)


jayvdb:
There are still more tests to write, to verify that path objects pass through each arg type sanely.
There will be some power users whose `tox.ini` will be broken, if they are doing complicated syntax in a single value that starts with a path, usually to workaround the problem that this solves by adding quotation marks around strings.  I'll construct an example of what can break.  And maybe detect any common cases and disable the new mode.

The most obvious way to reduce the surface area of the possible breakage is to not use literal path mode if the input value contains `'` anywhere (and maybe `"`).

gaborbernat (commented):
I wonder if we really need to handle this use case given no one complained about this yet. Makes the code much more complicated, and paths having ``#`` feels more an academic exercise than something likely to come up. Also now introduces a new flag, which I'm not that keen on migrating to version 4. tox already has too many options 🤔

jayvdb:
Just using this approach for type `path` solves the issues raised, and doesnt need the new flag, so I could split that bit off, and get that bit polished first.  The benefit isnt just `#` and ` ` - the code may look a bit more complicated, but having a shorter path through the engine for paths makes the parser conceptually easier, and makes it much easier for users to do medium complexity string operations using paths (which is a fairly common case), without fighting to get strings through `CommandLine` and `shlex` *twice* if they want the value in the subprocess environment.

The `literal_paths` flag is only part of the handling of types other than `path`, which is a segment I am much less attached to, and may also be unnecessary if I can find a nice way to avoid any reason people would want the old parser.

Another approach to allow to bypass the shlex'ing of paths for other types of values is to do allow something like `{path:{toxinidir}/foo#bar}` to inject literals, or maybe even `{glob:{toxinidir}/foo#bar}` (https://github.com/tox-dev/tox/issues/1571) if that provides most of the benefits.

jayvdb:
> Also now introduces a new flag, which I'm not that keen on migrating to version 4.

If this literal_paths flag is merged, I would expect that v4 does not have this flag, and only uses this new literal path mode.

There is nothing in the prior syntax which can not be achieved more cleanly with the new mode.
The messy case is where one key starts with a path and also contains additional strings which uses quotes that would be removed by shlex.  I'll provide some more concrete examples, but it will take a bit of effort to verify what exactly will break.

To convert old mode to new mode, those problematic value would need to be split into two keys, with one being a path only (and be literal, no shlex/CommandLine), and the other joining the first key and the other strings (and uses shlex and CommandLine as before).

The ability to disabling it in v3 would only be so that people dont need to re-write their ini, if they have complex syntax and it behaves slightly differently, and they dont want to learn the changes.

gaborbernat:
> If this literal_paths flag is merged, I would expect that v4 does not have this flag, and only uses this new literal path mode.

tox 4 aims (as much as possible) to be configuration file compatible, so any flag we introduce today will likely be grandfathered into tox 4... otherwise, people would have to rewrite their tox.ini just to start using tox 4, which IMHO is too big ask.

jayvdb:
> tox 4 aims (as much as possible) to be configuration file compatible

As you can see from this PR, the new mode passes all existing tests, and new tests.  It is compatible with an extremely high percentage of existing tox.ini uses.  I am being overly cautious and explaining that there are small breakages, so as to not mislead code reviewers.  The next step is to isolate them.  Also as above I have outlined quite a few ways to eliminate even the small breakages, and I am going to explore that also.

Also it would be nice if I could get your feedback on the proposal above at https://github.com/tox-dev/tox/pull/1713#issuecomment-715880396 to split off the `type=path` component which fixes the issues raised without any syntax breakage, and thus has no introduced flag.

gaborbernat:
> medium complexity string operations using paths (which is a fairly common case)

Can you detail on these a bit?

gaborbernat:

gaborbernat:

Issue #1716, opened 2020-10-25
TOX_MISSING_SUBSTITUTION


jayvdb:
While processing venv options defined by `parser.add_testenv_attribute`, if any of them raise `tox.exception.MissingSubstitution`, the exception is caught and the value "TOX_MISSING_SUBSTITUTION" will be the value of the setting.
i.e. there is no exception occurring for the config.  The `TestenvConfig` is created, silently placing invalid value "TOX_MISSING_SUBSTITUTION" into the config, and adding the setting name to `TestenvConfig._missing_subs`.

An easy way to trigger this is `commands = echo {env:{env:FOO}}` , but there are many other variations.

`VirtualEnv.setupenv` does detect the problem by reading `TestenvConfig._missing_subs`, and the tox cli tries to nicely show

> unresolvable substitution(s): 'FOO'. Environment variables are missing or defined recursively.

There is one cli test (test_missing_env_fails) for "unresolvable substitution".  There are no tests directly for `TOX_MISSING_SUBSTITUTION` or `MissingSubstitution`.

The invalid value "TOX_MISSING_SUBSTITUTION" causes a few strange results depending on which settings it is injected into, and how the settings are used.

The simplest example of undesirable behaviour is
```ini
[tox]
envlist = py27
[testenv:py27]
description = {env:{env:FOO}}
```
```
$ tox -l -v
default environments:
py27 -> TOX_MISSING_SUBSTITUTION
```

The next class of errors is plugin's with hook `tox_configure` are unaware that the config is invalid unless they check private variable `TestenvConfig._missing_subs`.
The invalid value "TOX_MISSING_SUBSTITUTION" could be in any other setting, then used in any other context by the configure hook, which can cause all sorts of strange exceptions because the other setting may be of type list, dict, etc, and the logic processing those settings are not expecting a string value.

```py
import tox

@tox.hookimpl
def tox_addoption(parser):
    parser.add_testenv_attribute(name="dummy", type="string", help="dummy")

@tox.hookimpl
def tox_configure(config):
    print('dummy', list(config.envconfigs.values())[0].dummy)
```
Output with key `dummy = {env:FOO}` :
```
dummy TOX_MISSING_SUBSTITUTION
...
ERROR:   py27: unresolvable substitution(s): 'FOO'. Environment variables are missing or defined recursively
```

Some example real cases

[...]

PR #1717, opened 2020-10-25
Raise MissingSubstitution on access

jayvdb:
Avoid unexpected invalid "TOX_MISSING_SUBSTITUTION" being inserted into settings.

Fixes https://github.com/tox-dev/tox/issues/1716

Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/docs/changelog/1716.bugfix.rst
+++ b/docs/changelog/1716.bugfix.rst
@@ -1,0 +1,1 @@
+Raise ``MissingSubstitution`` on access of broken ini setting. - by :user:`jayvdb`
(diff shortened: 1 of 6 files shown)


jayvdb:
With this, `tox -v -l` will show an `MissingSubstitution` exception if a section has an invalid `description`.  I can catch and print the exception instead, if that is preferable.  It was the only "tox core" problem (i.e. no plugins) I could quickly find, but any tox execution flow which does config and doesnt do venv `setupenv` is prone to have the type of problem explained in the issue.

The alternative would be to do check of `_missing_subs` for all venvs before invoking `tox_configure` hook, but I havent looked down that alley, and I guess that would break any plugin which using that hook to 'fix' the ini to be valid (never seen any plugin which does this, and would be badly designed anyway).

gaborbernat (approved):

Issue #1534, opened 2020-03-04
passenv=FOO passes on foo


blueyed:
When using `passenv = FOO` it willl pass-through `foo`.

Code ref: https://github.com/blueyed/tox/blob/2cfc0b218acec5cc8b2ff07e4351c72f4945c30e/src/tox/config/__init__.py#L720-L723

gaborbernat:
Feel free to open a PR against master and fix it. My available efforts at the moment are aimed at fixing this as part of https://github.com/tox-dev/tox/issues/1394, but that probably will take a while (ETA September).

gnikonorov:
Hey @gaborbernat, could I submit a pr for this?

gaborbernat:
Sure 👍

gnikonorov:
To elaborate on the description, it seems that the value passed to `envlist` in the `tox.ini` file will take all environment variables that match, regardless of the case of the letters.

Environment variables:
```
gnikonorov:~/temp/test_tox$ echo $FOO
big foo
gnikonorov:~/temp/test_tox$ echo $foo
little foo
gnikonorov:~/temp/test_tox$ echo $fOo
mixed case foo
gnikonorov:~/temp/test_tox$
```

`tox.ini`
```
gnikonorov:~/temp/test_tox$ cat tox.ini
[tox]
envlist = py36
isolated_build = True

[testenv]
passenv = foo
commands = pytest --capture=no test_this.py
gnikonorov:~/temp/test_tox$
```

sample test file:
```
gnikonorov:~/temp/test_tox$ cat test_this.py
import os

def test_get_foo():
    print(os.environ.get("FOO", "FOO variable not found"))
    print(os.environ.get("fOo", "fOo variable not found"))
    print(os.environ.get("foo", "foo variable not found"))
gnikonorov:~/temp/test_tox$
```

Test run
```
py37 run-test: commands[0] | pytest --capture=no test_this.py
...
=========================================================================================================== test session starts ============================================================================================================
platform linux -- Python 3.6.9, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
cachedir: .tox/py37/.pytest_cache
rootdir: /home/gnikonorov/temp/test_tox
plugins: dynamicrerun-1.0.4
collected 1 item

test_this.py big foo
mixed case foo
little foo
.

...
_________________________________________________________________________________________________________________ summary __________________________________________________________________________________________________________________
  py37: commands succeeded
  congratulations :)
[...]

PR #1718, opened 2020-10-25
Make the passenv environment setting case sensitive

gnikonorov:
## Description
Make the `passenv` environment setting case sensitive. Before this change it was including all environment variables that matched in a case insensitive fashion ( `FOO` included `FOO`, `fOo`, and `foo` for example )

Closes https://github.com/tox-dev/tox/issues/1534

## Contribution checklist:

- [x] added/updated test(s)
- [x] updated/extended the documentation - Not applicable in this case 
      in message body

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -36,6 +36,7 @@ Florian Bruhin
 Florian Preinstorfer
 Florian Schulze
 George Alton
+Gleb Nikonorov
 Gonéri Le Bouder
 Hazal Ozturk
 Henk-Jaap Wagenaar
--- a/docs/changelog/1534.bugfix.rst
+++ b/docs/changelog/1534.bugfix.rst
@@ -1,0 +1,1 @@
+Make the ``passenv`` environment setting case sensitive. - by :user:`gnikonorov`
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -793,7 +793,7 @@ def tox_addoption(parser):
             passenv.add("TMPDIR")
         for spec in value:
             for name in os.environ:
-                if fnmatchcase(name.upper(), spec.upper()):
+                if fnmatchcase(name, spec):
                     passenv.add(name)
         return passenv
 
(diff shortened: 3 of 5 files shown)


asottile:
at least on posix this is incorrect, `A` and `a` are two separate environment variables:

```console
$ a=1 A=2 env | grep -i '^a='
A=2
a=1
```

asottile:
on windows however, `a` and `A` are the same variable:

```
C:\Users\Anthony>set A=1

C:\Users\Anthony>set a=2

C:\Users\Anthony>echo %A% %a%
2 2
```

gnikonorov:
> on windows however, `a` and `A` are the same variable:
> 
> ```
> C:\Users\Anthony>set A=1
> 
> C:\Users\Anthony>set a=2
> 
> C:\Users\Anthony>echo %A% %a%
> 2 2
> ```

Thank you @asottile. I forgot about Windows. I'll add an exclusion for it.

asottile:
I'm 50/50 on whether we should change this, or whether we should just document it as case insensitive

gnikonorov:
Since this is my first time opening a PR in this repo, I'll defer to more experienced committers. Thoughts @gaborbernat ( since we spoke on the linked issue ) ?

gaborbernat (changes requested):
> FWIW I think it would be easy and clean enough to just uppercase `name` and `spec` if `platform.system() == "Windows":`

Yeah, on Windows we need to be the case insensitive.

asottile:
On second thought, changing this behaviour is likely to break a lot of people [even tox itself](https://github.com/tox-dev/tox/blob/c5a10d730e562164b03f2669596e0168ba9b3aac/tox.ini#L25-L27)

I'm leaning more towards not changing the behaviour and just documenting the case insensitivity

gnikonorov:
> On second thought, changing this behaviour is likely to break a lot of people [even tox itself](https://github.com/tox-dev/tox/blob/c5a10d730e562164b03f2669596e0168ba9b3aac/tox.ini#L25-L27)
> 
> 
> 
> I'm leaning more towards not changing the behaviour and just documenting the case insensitivity

At the very least, it sounds like a major version bump

gaborbernat:
Ok. Let's not address here in the code. But document the behavior.

PR #1727, opened 2020-10-31
Make parsing [testenv] sections in setup.cfg official

mauvilsa:
Even though #1724 was reverted, I suppose that to make testenv sections in setup.cfg an official feature, there should be unit tests for it and be mentioned in the documentation. So I open this new pull request. I have modified the code so that it is backward compatible and added a unit test so that now it is tested for both `[testenv]` and `[tox:testenv]` sections.

Self-referencing #1727 since there is no issue associated to this.

## Contribution checklist:

- [X] added/updated test(s)
- [X] added relevant [issue keyword](https://help.github.com/articles/closing-issues-using-keywords/)
      in message body
- [X] added news fragment in [changelog folder](../tree/master/docs/changelog)
- [X] added yourself to `CONTRIBUTORS` (preserving alphabetical order)


docs/changelog/1727.feature.rst
```diff
@@ -0,0 +1 @@
+Support parsing [tox:testenv*] sections in setup.cfg. - by :user:`mauvilsa`
```
gaborbernat, line 1: Perhaps the right wording here is officially/explicitly. Because the support is already there, just not spelled out.

gaborbernat (approved):

docs/changelog/1727.feature.rst
mauvilsa: Support with the `tox:` prefix is not there yet.

gaborbernat (changes requested):
Ah no, sorry, miss-understood then. We're not planning on adding new ways to do this. You can already achieve what you want via ``[tox]`` and ``[testenv]`` which should suffice. I see no reason to support yet another way.

mauvilsa:
@gaborbernat I guess you meant that we can already achieve this with `[tox:tox]` and `[testenv]`. Sure if this is what you want I can change the pull request only adding one unit test and the update of the documentation. Though I don't have time to do this right now.

gaborbernat (approved):

PR #1739, opened 2020-11-11
tests: do not depend on pathlib2 for modern python

### LINKED ISSUE #1549: [3.14.6] Build tests fails, because of pathlib2
pathlib2 is deprecated for python 3.
See:
https://github.com/mcmtroffaes/pathlib2

eli-schwartz:
Try pathlib and fall back to pathlib2.

The latter is technically installable on python 3.4+, but does nothing the stdlib doesn't do better. And in tightly constrained build environments, e.g. distro packaging, any dependency or test dependency must be a system package, but old, python2-specific backports of the stdlib might not actually be packaged...

Make the job of distro packagers easier by not requiring patches to use pathlib while testing python3.

Fixes #1549

## Contribution checklist:

- [ ] added/updated test(s)
      in message body

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,13 +61,13 @@ docs =
 testing =
     flaky>=3.4.0
     freezegun>=0.3.11
-    pathlib2>=2.3.3
     psutil>=5.6.1
     pytest>=4.0.0
     pytest-cov>=2.5.1
     pytest-mock>=1.10.0
     pytest-randomly>=1.0.0
     pytest-xdist>=1.22.2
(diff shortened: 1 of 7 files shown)


gaborbernat (changes requested):
Instead of try/fallback I prefer explicit if/else per version for these

eli-schwartz:
Done.

I notice one CI check is failing due to lack of a changelog. Do you consider this important enough to merit a changelog entry? Should I add one?

gaborbernat (approved):

Issue #1741, opened 2020-11-12
Allowing to package with formats different than zip


wherka:
Hello,
as it stands tox exposes the concept of **distshare** area which allows to get the packages as a side effect of the tests. Which in my opinion is quite nice as one may capture the packages which were actually used during the tests and do something with them - releasing or perhaps just publishing them as a mean of the pre-release validation.

However, there one aspect which bothers me a bit. The --formats is hardcoded to **zip**. Wouldn't it be nice to make it more flexible and pass the suitable value as per [official doc](https://docs.python.org/3/distutils/sourcedist.html)?

I appreciate tox was not meant for packaging,  but since it's there already, can we not make it a little bit more convenient to the users please?

Thanks in advance for consideration!

Cheers,
Waldek

gaborbernat:
As the name of the file indicates that part of the system is considered legacy, and no longer planned to be extended any further. Such functionality should be set via your build backends configuration files, per PEP-517.

PR #1742, opened 2020-11-12
PR#1741: Allowing to package with formats different than zip

wherka:
This PR is an example implementation/a supplement to: https://github.com/tox-dev/tox/issues/1741

It should not be considered as complete as there are no tests attached and the docs would need to be modified should the idea described in the original ticked be accepted. It should be considered merely as a guage to measure the complexity of the job at hand and potential side effects if any.

I'm happy to complete all the steps described in the checklist once we've passed the first discussion stage.

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -1187,6 +1187,7 @@ class ParseIni(object):
         config.temp_dir = reader.getpath("temp_dir", "{toxworkdir}/.tmp")
         reader.addsubstitutions(temp_dir=config.temp_dir)
         config.sdistsrc = reader.getpath("sdistsrc", None)
+        config.sdist_formats = reader.getstring("sdist_formats", "zip")
         config.setupdir = reader.getpath("setupdir", "{toxinidir}")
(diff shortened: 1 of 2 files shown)


asottile:
I'd rather we remove the `distshare` feature than give it more features personally.

wherka:
I honestly don't feel strongly about removing **distshare**. If you do it, that's okay too as long as the other end users don't care about that either. You could announce it and see what's the feedback.

However, I find it quite this feature quite useful if I'm being honest. Running all the tests with tox and having the packages which were used during the tests gives me the assurance that I ship something that was indeed tested and not just my assumption that my packaging mechanism(whatever it might be) is not breaking it. 
Of course  you may say, that this sort of level of testing is never sufficient and I should be running the integration, functional and other tests using the released packages and I would completely agree, but having the same bundles from start to finish is not a bad thing at all if you ask me ;-)

gaborbernat (commented):
As the name of the file indicates that part of the system is considered legacy, and no longer planned to be extended any further. Such functionality should be set via your build backends configuration files, per PEP-517.

Issue #1191, opened 2019-03-12
lazy configuration evaluation


gyermolenko:
Trying to run command for particular python3.7 testenv (by `tox -e env37`) I get error related to another testenv:
`"execution failed:  -- pyenv: python2.7: command not found`

Got same results for travis-ci and local env with `pyenv`.

Seems like using `{envsitepackagesdir}` in one testenv influences all other testenvs.

Files to reproduce:

```
from setuptools import setup
setup()
```
```
[tox]
envlist = env27,env37

[testenv:env27]
basepython = python2.7
commands =
    python -c 'print {envsitepackagesdir}'
    
[testenv:env37]
basepython = python3.7
commands =
    python -V
```

**Steps to reproduce:**  run `tox -e env37`

**Expected:** 
```
$ tox -e env37
GLOB sdist-make: /home/gy/code/temp/toxer2/setup.py
env37 inst-nodeps: /home/gy/code/temp/toxer2/.tox/.tmp/package/1/UNKNOWN-0.0.0.zip
env37 installed: UNKNOWN==0.0.0
env37 run-test-pre: PYTHONHASHSEED='1503595934'
env37 runtests: commands[0] | python -V
Python 3.7.2
_______________________________________________________ summary ________________________________________________________
  env37: commands succeeded
```

**Actual:**
```
$ tox -e env37
execution failed:  -- pyenv: python2.7: command not found

The `python2.7' command exists in these Python versions:
  2.7.15

GLOB sdist-make: /home/gy/code/temp/toxer2/setup.py
env37 inst-nodeps: /home/gy/code/temp/toxer2/.tox/.tmp/package/1/UNKNOWN-0.0.0.zip
env37 installed: UNKNOWN==0.0.0
env37 run-test-pre: PYTHONHASHSEED='3600038053'
env37 runtests: commands[0] | python -V
Python 3.7.2
_______________________________________________________ summary ________________________________________________________
  env37: commands succeeded
```

Not reproduced with `{envname}` or `{toxinidir}`

gaborbernat:
This is valid, we parse the entire configuration file at once, not just on demand the one you want to invoke. So if any data can't be deduced and configuration parsing (such as here the envsitepackagesdir) you'll see this error.

> Not reproduced with {envname} or {toxinidir}

Those are easily satisfied (deduced) hence why you can't see errors with that.

[...]

PR #1795, opened 2021-01-11
config: Lazy evaluation 

jayvdb:
Delay evaluation of config until accessed.

Fixes https://github.com/tox-dev/tox/issues/1191

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -99,7 +99,7 @@ class Parser:
             prog="tox",
             formatter_class=HelpFormatter,
         )
-        self._testenv_attr = []
+        self._testenv_attr = {}
 
     def add_argument(self, *args, **kwargs):
         """add argument to command line parser.  This takes the
@@ -122,7 +122,7 @@ class Parser:
         Any postprocess function must return a value which will then be set
         as the final value in the testenv section.
         """
-        self._testenv_attr.append(VenvAttribute(name, type, default, help, postprocess))
+        self._testenv_attr[name] = VenvAttribute(name, type, default, help, postprocess)
 
     def add_testenv_attribute_obj(self, obj):
         """add an ini-file variable as an object.
@@ -134,7 +134,7 @@ class Parser:
         assert hasattr(obj, "type")
         assert hasattr(obj, "help")
         assert hasattr(obj, "postprocess")
-        self._testenv_attr.append(obj)
+        self._testenv_attr[obj.name] = obj
 
     def parse_cli(self, args, strict=False):
         args, argv = self.argparser.parse_known_args(args)
@@ -1390,7 +1390,7 @@ class ParseIni(object):
             envpython=tc.get_envpython,
             **subs
         )
-        for env_attr in config._testenv_attr:
+        for env_attr in config._testenv_attr.values():
             atype = env_attr.type
             try:
                 if atype in (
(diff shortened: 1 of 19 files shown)


jayvdb:
There is one outstanding and rather obscure set of test failures in `test_test_usedevelop`, which I havent had time to hunt down.  Any clues would be appreciated.

I think this should be held back atm, and targetted at 3.23, so that 3.21 regressions have had time to be raised and fixed in 3.21.x or 3.22 if a minor release is needed to sort out any problems arising from 3.21.

tox.ini
```diff
@@ -33,10 +33,6 @@ deps =
     pip >= 19.3.1
 extras = testing
 commands = pytest \
-           --cov "{envsitepackagesdir}/tox" \
```
gaborbernat, line 36: Eh what?

gaborbernat (changes requested):
Is this a new feature, or a bugfix? If it's a new feature I'm going to close it as won't do. Lazy evaluation is done by design in tox 4, and I don't plan to introduce a such major feature (that's likely to break a lot of things) part of tox 3.

tox.ini
jayvdb: Just speeding up my testing whilst developing.  Needs removing before un-drafting this.

jayvdb:
> Is this a new feature, or a bugfix?

This was part of a cleaner way of solving https://github.com/tox-dev/tox/issues/1777 , but semi-abandoned because of the remaining problem, and the quick fix is better suited for a patch release.

IIRC there were some older bugs this would have fixed.  I'll see if I can find them, and them we can evaluate whether they are also best left unsolved until v4.

gaborbernat:
The non-lazy evaluation is a feature kind of tox 3. This forces all fields to be validated at startup time. And this kind of behavior change can have all kinds of side effects that are a much better-made part of a new major release than a patch release. I'm just warry of accepting a such major change in here.

src/tox/config/__init__.py
```diff
@@ -1389,48 +1452,16 @@ def make_envconfig(self, name, section, subs, config, replace=True):
...
-                setattr(tc, env_attr.name, res)
-            if atype in ("path", "string", "basepython"):
-                reader.addsubstitutions(**{env_attr.name: res})
+        reader._tc = tc
+        # Force postprocess as there is the expectation that attributes
+        # postprocess runs at startup.
+        # This is needed for setting up setenv, and also for tox plugins
+        # that rely on this assumption.
+        for env_attr in self.config._testenv_attr.values():
+            if env_attr.postprocess:
```
jayvdb, line 1461: Note that removing this condition makes it 100% non-lazy again, after which this has the primary benefit of `_replace_substitution` using the new `_get_value` which has access to the `VenvAttribute` metadata for each variable.  The reason this can be helpful is that e.g. `install_command` should declare that `{opts}` and `{packages}` are "fake" variables, and this is needed inside the replacement engine so that they are re-escaped variables for each pass through the replacement engine.

gaborbernat:
I'll close this. Doing such major change seems high-risk low reward within tox 3. Let me know if you're willing to push this through and then maintain any potential fallouts due to it. If so we can reopen, however on my side I'm not willing to invest energy in this one considering tox 4 is not long around the corner.

Issue #1684, opened 2020-10-06
Name ".tox" should not be allowed in "envlist"


sinoroc:
* _tox 3.20.0_

---

Name `.tox` should not be allowed in `envlist`.

Apparently there are safe guards in code, so that when `isolated_build` is set in `tox.ini` then adding `.package` to `envlist` and/or running `tox -e .package` throws an exception relatively early on:

```
tox.exception.ConfigError: ConfigError: isolated_build_env .package cannot be part of envlist
```

There doesn't seem to be such safe guards for the case where `requires` is set in `tox.ini` and `.tox` is added to `envlist`.

---

```
[tox]
envlist =
    .tox
requires =
    wheel
skipsdist = True

[testenv]
commands =
    python -V
```

```
$ rm -rf .tox && tox
.tox create: /home/sinoroc/workspace/dot-tox/.tox/.tox
.tox installdeps: wheel, tox >= 3.20.0
.tox recreate: /home/sinoroc/workspace/dot-tox/.tox/.tox
ERROR: invocation failed (exit code 2), logfile: /home/sinoroc/workspace/dot-tox/.tox/.tox/log/.tox-0.log
=============================================== log start ===============================================

================================================ log end ================================================
ERROR: InvocationError for command /home/sinoroc/workspace/dot-tox/.tox/.tox/bin/python -m virtualenv --no-download --python /home/sinoroc/workspace/dot-tox/.tox/.tox/bin/python .tox (exited with code 2)
________________________________________________ summary ________________________________________________
ERROR:   .tox: InvocationError for command /home/sinoroc/workspace/dot-tox/.tox/.tox/bin/python -m virtualenv --no-download --python /home/sinoroc/workspace/dot-tox/.tox/.tox/bin/python .tox (exited with code 2)

```

sinoroc:
I have not tested those and they might well be covered already, but since we are here: are other names such as `.tmp`, `dist`, `log`, `.package.lock`, etc. already safe guarded?

gaborbernat:
Those are within the tox environments, so they don't need safeguarding.

gaborbernat:
Released now via https://tox.readthedocs.io/en/latest/changelog.html#v3-21-1-2021-01-13

PR #1800, opened 2021-01-11
config: Prevent .tox in envlist

jayvdb:
Fixes https://github.com/tox-dev/tox/issues/1684


jayvdb:
IMO this should be left to after https://github.com/tox-dev/tox/pull/1801 and any other regressions have been fixed and a patch release is out the door, and then this included in the next minor release.

PR #1816, opened 2021-01-13
POC: tox4 catch plugin load failure

ssbarnea:
Not really working as the exception is opaque and does not give a hint on which plugin caused it.

Related: https://github.com/tox-dev/tox/discussions/1810


gaborbernat:
This is for tox 3, not tox 4 🤔

PR #1817, opened 2021-01-13
WIP: catch plugin loading failure

ssbarnea:


src/tox/plugin/manager.py
```diff
@@ -42,12 +43,16 @@ def __init__(self) -> None:
         )
 
         for plugin in internal_plugins:
+            logging.info("Registering %s", plugin)
```
gaborbernat, line 46: I'd prefer to be silent if nothing bad happens.

src/tox/plugin/manager.py
```diff
@@ -42,12 +43,16 @@ def __init__(self) -> None:
...
+        try:
+            self.manager.load_setuptools_entrypoints(NAME)
+            self.manager.register(state)
 
-        REGISTER.populate(self)
-        self.manager.check_pending()
+            REGISTER.populate(self)
+            self.manager.check_pending()
+        except ImportError as e:
+            logging.warning("Ignored plugin %s which failed to load. %s", NAME, e)
```
gaborbernat, line 55: Here you want `%r` to show the type too. And ideally, we'd also show the last source code line of the traceaback (you can get it via traceback.format_exc).

gaborbernat (changes requested):
Yeah, something similar I was thinking.

gaborbernat:
I'll close this now as it seems we'll not be able to merge it.

PR #1819, opened 2021-01-13
Simplify td_human_readable function

eumiro:
We're still miles away from `datetime.timedelta.__format__`, so this is to clean up and simplify the `td_human_readable` function. No external functionality would be affected, so no need to update anything else.

Looking forward to tox 4!

### DIFF
--- a/src/tox/util/spinner.py
+++ b/src/tox/util/spinner.py
@@ -5,7 +5,7 @@ import threading
 import time
 from collections import OrderedDict
 from types import TracebackType
-from typing import IO, Dict, Optional, Sequence, Type, TypeVar
+from typing import IO, Dict, List, Optional, Sequence, Type, TypeVar
 
 from colorama import Fore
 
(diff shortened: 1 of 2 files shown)


eumiro:
Since the coverage was not complete, I added some test cases and then had to fix the bugs in the function. Let's see what the CIs say.

gaborbernat (changes requested):
Can you add some test to fix the coverage:

```
src/tox/util/spinner.py     112      1     30      1    99%   174->178, 178
```

eumiro:
> Can you add some test to fix the coverage:
> 
> ```
> src/tox/util/spinner.py     112      1     30      1    99%   174->178, 178
> ```

Yes, I just do not understand which line of the function has not been tested.

eumiro:
```
 @codecov codecov / codecov/patch
src/tox/util/spinner.py#L178
Added line #L178 was not covered by tests
```

That's the line with a single `continue` that's processed each time a `period` does not appear in the result, which is actually every time the case. :shrug:

gaborbernat:
> That's the line with a single `continue` that's processed each time a `period` does not appear in the result, which is actually every time the case. 🤷

Be that so would imply a coveragepy bug, can we track that down? I'll hold off merging it until it gets to 100%, I spent quite a bit of time to get there so I want to avoid not having it.

eumiro:
> Be that so would imply a coveragepy bug, can we track that down? I'll hold off merging it until it gets to 100%, I spent quite a bit of time to get there so I want to avoid not having it.

Refactored the code a little bit, `codecov-patch` has now 100%. `codecov-project` has a little bit less only because `hits` went down (because the number of LOC went down), but `misses` and `partials` remained the same. So now it should be okay, right?

gaborbernat (approved):
It's alright, no changelog needed for this because it's an internal refactor and has no external effect.

PR #1874, opened 2021-01-26
Add PIP_TRUSTED_HOST to default passenv variables

d-ryzhikov:
Add `PIP_TRUSTED_HOST` to default passenv variables.

Closes https://github.com/tox-dev/tox/issues/1875

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -28,6 +28,7 @@ Cyril Roelandt
 Dane Hillard
 David Staheli
 David Diaz
+Dmitry Ryzhikov
 Ederag
 Eli Collins
 Eugene Yunak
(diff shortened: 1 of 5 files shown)


asottile (changes requested):
I don't think this is universal enough to warrant a special case in tox -- there are many-many pip options and allowlisting each one doesn't seem correct.  you can allow this in your own configuration using `passenv = PIP_TRUSTED_HOST` (or `passenv = PIP_*` if you want to allow all pip options)

I'll let the others chime in, but I don't think we should merge this and the associated issue should be closed

d-ryzhikov:
@asottile , in my experience it was often that private package index has either misconfigured or missing HTTPS, which is why I decided to propose to add this variable to the list of defaults.
What do you think about adding all `PIP_*` variables to the default passenv list?

gaborbernat:
In tox 4 we do accept (for now) all PIP_*. I'd defer making any decision in tox 3 for now.

Issue #1921, opened 2021-02-16
Ability to disable provisioning


hroncok:
I was looking into redefining how provisioning works via a plugin. I was not successful, it seems that provisioning happens before any plugin code even runs. I've seen the code in tox and realized that changing it in 3.x would be very tedious.

However, I've also realized, that I don't need to redefine the behavior, all I need is to stop it from happening. Apparently, all I needed to do was to set the `TOX_PROVISION=1` environment variable to get this exception:

https://github.com/tox-dev/tox/blob/d5957b82de26c5785fd88eb29106e478b622ac8b/src/tox/config/__init__.py#L1349-L1351

Before I start using this as a hack:

 - is this even an API?
 - what are the side effects? I see that `WITHIN_PROVISION` is used at more places.
 
Would it be possible to have a documented way to error here, without side effects? I was thinking `TOX_NO_PROVISION` or even `TOX_PROVISION=error`. Ideally, the error would print the list of missing dependencies in a way that is easily parseable (with literal_eval is fine).

I can work  on this, if this would be accepted. I can also go into details about why I'd like to have this if you want.

gaborbernat:
Provisioning is meant to pull in the plugins so having a plugin that alters how you pull in plugins seems very meta. There has been no thought put into this behavior and overriding ``TOX_PROVISION`` leaves you in undefined land. Please do tell why though 👍🏻

hroncok:
In Fedora, we use tox to run the upstream tests when we build the RPM packages. Unlike regular tox users, we don't want to install packages from PyPI to a virtual environment, because we need to test integration with other Fedora packages. We have created a plugin for our needs. More details about the use case are in the README. https://github.com/fedora-python/tox-current-env

 1. We use the plugin to generate a list of requirements.
 2. We install those requirements via Fedora RPM packages.
 3. We use the plugin to run the tests in "current" Python environment.

This all works unless tox determines it needs provision. With provisioning, the situation differs depending on internet availability.

When online:

 1. We use the plugin to generate a list of requirements.
    1. tox creates a provision environment
    2. tox fails, because our plugin is not installed into the provision environment

When offline:

 1. We use the plugin to generate a list of requirements.
    1. tox attempts to creates a provision environment, it fails

What I'd like to do instead:

 1. We use the plugin to generate a list of requirements with `TOX_NO_PROVISION` (or similar) set
    1. tox figures requires are missing, it errors with the list
    2. we parse the list
[...]

PR #1922, opened 2021-02-17
Add --no-provision flag

hroncok:
Add --no-provision flag

tox can now be invoked with a new --no-provision flag that prevents provision,
if requires or minversion are not satisfied, tox will fail;
if a path is specified as an argument to the flag
(e.g. as `tox --no-provision missing.json`) and provision is prevented,
provision metadata are written as JSON to that path.

Fixes https://github.com/tox-dev/tox/issues/1921

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -2,6 +2,7 @@ from __future__ import print_function
 
 import argparse
 import itertools
+import json
 import os
 import random
 import re
@@ -571,6 +572,16 @@ def tox_addoption(parser):
         "--alwayscopy",
         action="store_true",
         help="override alwayscopy setting to True in all envs",
+    )
+    parser.add_argument(
+        "--no-provision",
+        action="store",
+        nargs="?",
+        default=False,
+        const=True,
+        metavar="REQUIRES_JSON",
+        help="do not perform provision, but fail and if a path was provided "
+        "write provision metadata as JSON to it",
     )
 
     cli_skip_missing_interpreter(parser)
@@ -1318,8 +1329,8 @@ class ParseIni(object):
         # raise on unknown args
         self.config._parser.parse_cli(args=self.config.args, strict=True)
 
-    @staticmethod
-    def ensure_requires_satisfied(config, requires, min_version):
+    @classmethod
+    def ensure_requires_satisfied(cls, config, requires, min_version):
         missing_requirements = []
         failed_to_parse = False
         deps = []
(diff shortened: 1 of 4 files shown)


src/tox/config/__init__.py
```diff
@@ -1346,12 +1357,31 @@ def ensure_requires_satisfied(config, requires, min_version):
...
+            )
         if WITHIN_PROVISION and missing_requirements:
             msg = "break infinite loop provisioning within {} missing {}"
             raise tox.exception.Error(msg.format(sys.executable, missing_requirements))
         config.run_provision = bool(len(missing_requirements))
         return deps
 
+    @staticmethod
+    def write_requires_to_json_file(config):
+        import json
```
gaborbernat, line 1376: Can you make this a top-level import, please?

src/tox/config/__init__.py
```diff
@@ -1346,12 +1357,31 @@ def ensure_requires_satisfied(config, requires, min_version):
                 missing_requirements.append(str(requirements.Requirement(require)))
         if failed_to_parse:
             raise tox.exception.BadRequirement()
+        if config.option.no_provision and missing_requirements:
+            msg = "Provisioning explicitly disabled within {}, but missing {}"
```
gaborbernat, line 1361: lower case messages please 👍🏻

src/tox/config/__init__.py
```diff
@@ -1346,12 +1357,31 @@ def ensure_requires_satisfied(config, requires, min_version):
...
+            if config.option.no_provision is not True:  # it's a path
+                msg += ". Requires written to {}."
```
gaborbernat, line 1363: I don't think we need a new sentence we can just happen " and wrote to {}"

src/tox/config/__init__.py
```diff
@@ -572,6 +572,17 @@ def tox_addoption(parser):
...
         help="override alwayscopy setting to True in all envs",
     )
+    parser.add_argument(
+        "--no-provision",
+        action="store",
+        nargs="?",
+        default=False,
+        const=True,
+        metavar="REQUIRES_JSON",
+        help="Prevent tox provision, fail when requires are missing. "
```
gaborbernat, line 582: Can we start this with a lower case?

```
do not perform provision, but fail and if a path provided write provision metadata as JSON to it
```

src/tox/config/__init__.py
hroncok: I wanted to avoid it, since in 99% cases it won't be needed, but I can.

hroncok:
Docs added. I've discovered there is no way to reference an command line option: https://github.com/sphinx-contrib/autoprogram/issues/8

gaborbernat:
> Docs added. I've discovered there is no way to reference an command line option: [sphinx-contrib/autoprogram#8](https://github.com/sphinx-contrib/autoprogram/issues/8)

Hence why I've created https://pypi.org/project/sphinx-argparse-cli/ already used by the rewrite branch.  We could migrate this too.

hroncok:
> We could migrate this too.

I would not block this on it.

Working on tests.

gaborbernat (approved):

hroncok:
Thanks! When writing the docs, I've assumed this will be released as 3.23.0, correct?

gaborbernat:
https://github.com/tox-dev/tox/pull/1952/files looks like it

PR #1934, opened 2021-03-01
Add a test for #1928

hexagonrecursion:
See #1928

This PR does not fix the bug. It only adds a test for it. Note: this PR currently includes https://github.com/tox-dev/tox/pull/1930 because otherwise CI fails

## Contribution checklist:

- [x] added/updated test(s)
- [ ] updated/extended the documentation - N/A
- [ ] added relevant [issue keyword](https://help.github.com/articles/closing-issues-using-keywords/) - N/A
      in message body
- [ ] added news fragment in [changelog folder](../tree/master/docs/changelog) - will add after I open the PR
- [ ] added yourself to `CONTRIBUTORS` (preserving alphabetical order) - CONTRIBUTORS: No such file or directory

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -65,6 +65,7 @@ testing =
     flaky>=3.7
     freezegun>=1
     psutil>=5.7
+    pyramid<2  # workaround for https://github.com/devpi/devpi/issues/835
     pytest>=5.4.1
     pytest-cov>=2
     pytest-mock>=3
--- a/tests/bug/test_gh_1928.py
+++ b/tests/bug/test_gh_1928.py
@@ -1,0 +1,46 @@
+import json
+from pathlib import Path
+from pprint import pprint
+from typing import Any, List
+
+import pytest
+
+from tox.pytest import ToxProjectCreator
+
+
+@pytest.mark.xfail(raises=AssertionError)  # noqa: SC200
+@pytest.mark.parametrize(("posargs", "expected"), [([], ["probe.py"]), (["foo"], ["probe.py", "foo"])])
+def test_gh1928(  # noqa: SC200
+    tox_project: ToxProjectCreator,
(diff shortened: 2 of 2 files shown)


hexagonrecursion:
Since this PR does not fix the bug there is no user-visible change. Does it need a changelog entry?

setup.cfg
```diff
@@ -65,6 +65,7 @@ testing =
     flaky>=3.7
     freezegun>=1
     psutil>=5.7
+    pyramid<2  # workaround for https://github.com/devpi/devpi/issues/835
```
gaborbernat, line 68: Backdoor PR?

gaborbernat (changes requested):

setup.cfg
hexagonrecursion: > Note: this PR currently includes #1930 because otherwise CI fails

gaborbernat: Let's wait until upstream releases a fix.

tests/bug/test_gh_1928.py
```diff
@@ -0,0 +1,41 @@
+import json
```
gaborbernat, line 1: Please don't introduce a new layout pattern that goes against what the project already has in place. Also, I don't believe in the test only PRs, so would prefer to feature with test together.

gaborbernat: This should go and follow the existing ``{posargs}`` syntax from https://github.com/tox-dev/tox/blob/rewrite/tests/config/loader/ini/replace/test_replace_posargs.py

hexagonrecursion:
> Please don't introduce a new layout pattern that goes against what the project already has in place. Also, I don't believe in the test only PRs, so would prefer to feature with test together.

OK. back to the drawing board

PR #1954, opened 2021-03-03
Support for bracket styled posargs

hexagonrecursion:
Fixes #1928
In #1950 me and @gaborbernat have discussed how the `[]` substitution should behave in tox4. In this PR I'll implement the changes.

- [x] added relevant [issue keyword](https://help.github.com/articles/closing-issues-using-keywords/)
      in message body
- [X] added yourself to `CONTRIBUTORS` (preserving alphabetical order)
- [x] added news fragment in [changelog folder](../tree/master/docs/changelog)
- [x] added the tests
- [x] implemented the syntax

I assume adding myself to `CONTRIBUTORS` on the master branch counts.

PR #1955, opened 2021-03-03
Add xfail to the spelling whitelist

hexagonrecursion:
Without this pytest.mark.xfail is flagged as a misspelling.

https://github.com/tox-dev/tox/pull/1954#discussion_r586387012
> don't silence error instead add to whitelist https://github.com/tox-dev/tox/blob/rewrite/whitelist.txt

## Contribution checklist:

I assume adding myself to CONTRIBUTORS on the master branch counts.

### DIFF
--- a/whitelist.txt
+++ b/whitelist.txt
@@ -216,3 +216,4 @@ whl
 win32
 writestr
 xdist
+xfail


gaborbernat:
Let's add this where it's needed (aka where xfail is used) and not a separate PR.

hexagonrecursion:
IMHO it is best to change files like this on the trunk instead of changing them on a feature branch that may take a while to merge - fewer git conflicts.

gaborbernat:
I only want words in the whitelist that's actually used. But if the PR landing will not use xfail then we don't need it.

PR #1958, opened 2021-03-04
Use utf-8 for redirected stdio

### LINKED ISSUE #1908: Possible UnicodeError caused by missing encoding="utf-8"
https://github.com/tox-dev/tox/blob/555f3f13b18da1470b6518be654e3e4e2fdec654/src/tox/config/__init__.py#L304-L306

toml must be encoded by UTF-8. Please add `encoding="utf-8"` here.
It may cause UnicodeDecodeError when pyproject.toml contains non-ASCII character and locale encoding is not UTF-8 (e.g. Windows).

https://github.com/tox-dev/tox/blob/555f3f13b18da1470b6518be654e3e4e2fdec654/src/tox/util/stdlib.py#L52-L55

On Windows, stdout is UTF-8 encoded when it is console (e.g. _WinConsoleIO), but the default encoding of TemporaryFile() is legacy encoding. So UTF-8 is better encoding for this purpose.

This is not a real bug because this function is used only here:
https://github.com/tox-dev/tox/blob/cfe38bb853ff8573817c640a158b87eaad5b6e01/src/tox/session/__init__.py#L54

methane:
The default encoding for TemporaryFile is locale encoding. It is not UTF-8 on Windows.
So printing non-ASCII characters (e.g. Emoji) will cause UnicodeEncodeError.

Fixes #1908.

## Contribution checklist:

- [ ] added/updated test(s)
      in message body

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,7 +36,7 @@ repos:
           - black==20.8b1
         language_version: python3.8
   - repo: https://github.com/pre-commit/pygrep-hooks
-    rev: v1.7.1
+    rev: v1.8.0
     hooks:
       - id: rst-backticks
   - repo: https://github.com/asottile/setup-cfg-fmt
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,11 +17,11 @@ repos:
     rev: v2.7.4
     hooks:
       - id: pyupgrade
-  - repo: https://github.com/pre-commit/mirrors-isort
-    rev: v5.6.4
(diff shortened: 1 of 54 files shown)


gaborbernat (changes requested):
Needs a test and changelog.

methane:
This is not a real bug that end users see. suppress_output() is used only for suppress error message from argparse so it is ASCII only.

I don't want to make code uglier for Python 2 support, so I will create PR to rewrite branch. (I expect rewrite branch is Python 3 only).

methane:
I find suppress_output has been removed in rewrite branch. No need to fix it.

PR #1970, opened 2021-03-11
Show skipped envs in reports on platform mismatch

### LINKED ISSUE #1926: tox4: Not reporting skipped testenvs
So I trial-ran `tox==4.0.0a6` on the [latest master branch of aiosmtpd](https://github.com/aio-libs/aiosmtpd).

With tox3, I got the following summary conclusion:

```
  qa: commands succeeded
SKIPPED:  static: platform mismatch ('win32' does not match '^(?!win32)(?!cygwin)')
  docs: commands succeeded
```

On tox4, the "static" testenv is just missing:

```
  qa: OK (45.77=setup[2.86]+cmd[0.16,3.84,37.09,1.81] seconds)
  docs: OK (9.81=setup[2.89]+cmd[0.17,3.73,1.25,1.77] seconds)
```

Scrolling up in Terminal, tox4 never mentioned anything about the missing "static" testenv:

```
qa: OK ✔ in 45.77 seconds
docs: install_package> python -I -m pip install --no-deps --force-reinstall --no-build-isolation -e C:\Repos\aiosmtpd
docs: commands[0]> python housekeep.py prep
```

It would be nice if this tox3 behavior (i.e., reporting on skipped testenvs, and why) is kept in tox4.

jugmac00:
## Contribution checklist:

- [x] added/updated test(s)
      in message body

Resolves #1926.


jugmac00:
on my local machine I get a traceback from the coverage env oO

```
cov: commands[0]> coverage combine
cov: commands[1]> coverage report -m
Name    Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------
---------------------------------------------------
TOTAL    7797      0   1542      0   100%

178 files skipped due to complete coverage.
cov: commands[2]> coverage xml -o /home/jugmac00/Projects/tox/.tox/4/coverage.xml
cov: commands[3]> coverage html -d /home/jugmac00/Projects/tox/.tox/4/htmlcov
cov: commands[4]> diff-cover --compare-branch upstream/rewrite /home/jugmac00/Projects/tox/.tox/4/coverage.xml
  File "/home/jugmac00/Projects/tox/.tox/4/cov/bin/diff-cover", line 8, in <module>
    sys.exit(main())
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/diff_cover_tool.py", line 230, in main
    percent_covered = generate_coverage_report(
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/diff_cover_tool.py", line 212, in generate_coverage_report
    reporter.generate_report(output_file)
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/report_generator.py", line 251, in generate_report
    report = template.render(self._context())
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/report_generator.py", line 300, in _context
    context = super().report_dict()
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/report_generator.py", line 183, in report_dict
    src_stats = {src: self._src_path_stats(src) for src in self.src_paths()}
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/report_generator.py", line 85, in src_paths
    for src, summary in self._diff_violations().items()
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/report_generator.py", line 178, in _diff_violations
    for src_path in self._diff.src_paths_changed()
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/diff_reporter.py", line 151, in src_paths_changed
    diff_dict = self._git_diff()
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/diff_reporter.py", line 200, in _git_diff
    for diff_str in self._get_included_diff_results():
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/diff_reporter.py", line 173, in _get_included_diff_results
    included = [self._git_diff_tool.diff_committed(self._compare_branch)]
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/git_diff.py", line 66, in diff_committed
    return execute(self._default_git_args + self._default_diff_args + [diff_range])[
  File "/home/jugmac00/Projects/tox/.tox/4/cov/lib/python3.8/site-packages/diff_cover/command_runner.py", line 45, in execute
    raise CommandError(stderr)
diff_cover.command_runner.CommandError: fatal: ambiguous argument 'upstream/rewrite...HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

cov: exit 1 (0.29 seconds) /home/jugmac00/Projects/tox> diff-cover --compare-branch upstream/rewrite /home/jugmac00/Projects/tox/.tox/4/coverage.xml pid=8407
```

It's late, so I had no chance to dig deeper why coverage is failing...

gaborbernat:
> diff_cover.command_runner.CommandError: fatal: ambiguous argument 'upstream/rewrite...HEAD': unknown revision or path not in the working tree.

https://github.com/tox-dev/tox/blob/rewrite/tox.ini#L65 If you don't have an upstream remote (and work of origin), you should change it ``origin/rewrite``. Essentially you want to point that env-var to the base of the merge to get locally a coverage diff.

src/tox/tox_env/api.py
```diff
@@ -169,6 +169,11 @@ def setup(self) -> None:
             self._handle_env_tmp_dir()
 
     def ensure_setup(self, recreate: bool = False) -> None:
+        platform_str: str = self.conf["platform"]
```
gaborbernat, line 172: I'd prefer to keep this check behind the active property. So please restore the property; and ideally, I think we'll use the property to not schedule the tox environments at all here https://github.com/tox-dev/tox/blob/rewrite/src/tox/session/cmd/run/common.py#L267, but instead set skip status for it.

src/tox/session/cmd/run/common.py
```diff
@@ -179,9 +179,9 @@ def execute(state: State, max_workers: Optional[int], has_spinner: bool, live: b
     results: List[ToxEnvRunResult] = []
     future_to_env: Dict["Future[ToxEnvRunResult]", ToxEnv] = {}
     to_run_list: List[str] = []
-    for env in state.env_list():  # ensure envs can be constructed and are active
-        if state.tox_env(env).active:
-            to_run_list.append(env)
+    for env in state.env_list():  # ensure envs can be constructed
+        state.tox_env(env)  # this causes some side effects
```
gaborbernat, line 183: ``ensures envs can be constructed`` is all the comment you need here

tests/session/cmd/test_single.py
```diff
@@ -0,0 +1,20 @@
+import sys
```
gaborbernat, line 1: I believe this belongs into https://github.com/tox-dev/tox/blob/rewrite/tests/session/cmd/test_sequential.py

gaborbernat (changes requested):

jugmac00:
> > diff_cover.command_runner.CommandError: fatal: ambiguous argument 'upstream/rewrite...HEAD': unknown revision or path not in the working tree.
> 
> https://github.com/tox-dev/tox/blob/rewrite/tox.ini#L65 If you don't have an upstream remote (and work of origin), you should change it `origin/rewrite`. Essentially you want to point that env-var to the base of the merge to get locally a coverage diff.

Thanks for the clarification!

Indeed, I always keep `origin` for `upstream` and use `fork` for ... my fork :-) Made the most sense to me, but I guess most go with `upstream`.

As we are at 100% coverage, does this make sense any more? Or would it be sufficient to run `coverage` with `--fail-under 100`?

gaborbernat:
> As we are at 100% coverage, does this make sense anymore? Or would it be sufficient to run `coverage` with `--fail-under 100`?

Not yet on all platforms/python versions. Once we get there we can drop it 👍🏻

src/tox/session/cmd/run/common.py
jugmac00: Thanks - changed with the next push.

tests/session/cmd/test_single.py
jugmac00: Thanks - moved it with the next commit/push.

src/tox/tox_env/api.py
jugmac00: Hm, I probably can prevent to schedule the tox environment when I do the regex check inside `ToxEnv.active`.

But `active` returns a boolean, so I need to apply the same regex check later again, when I need to assemble a proper message a la "plat: skipped environment because platform linux does not match win32".

Hm, would I then trigger the message in `runcommands`?

https://github.com/tox-dev/tox/blob/41fa387a8c3e139ae7f953227dcbf57bede703a9/src/tox/session/cmd/run/single.py#L36-L45

I guess I need to spend more time with this :-)

Any additional tips appreciated.

gaborbernat: I think the best option is to make ToxEnvRunResult skipped an optional string, it's `None` if it's not skipped, but if it's skipped can be an arbitrary message. It also implies that active probably should be renamed to skip_reason that also returns an optional string.

gaborbernat: On second thought you don't need the property, can keep it here, just move it out to a method.

jugmac00: Thanks! I hope to find some time on the weekend. Really enjoying this!

jugmac00: I moved the regex check (platform) into its own method, but kept it in the `ensure_setup` method.

I then had a look around where `ensure_setup` is called - and there are 4 callers, but only the `_evaluate` method catches the `Skip` exception, so I think it would be better to call the newly created `check_platform` method just above the `tox_env.ensure_setup(recreate=recreate)` call...

https://github.com/tox-dev/tox/blob/34ef0e58d044782274a8a1fac6c43ad35994bb39/src/tox/session/cmd/run/single.py#L34-L45

ie...

```
    try:
        try:
            tox_env.check_platform()
            tox_env.ensure_setup(recreate=recreate)
            code, outcomes = run_commands(tox_env, no_test)
```

What do you think?

Thanks for taking the time to guide me through his PR!

jugmac00:
The coverage fail is probably because I have not yet rebased. My local cov is 100%

src/tox/tox_env/api.py
gaborbernat: Why do we need to catch it? Can't it go under the same logic that already catches missing interpreter logic?

jugmac00: Currently, the platform check is done via `check_platform` inside the `ensure_setup` method...

https://github.com/tox-dev/tox/blob/5bedb68c29b8ae54262c3770dab65bd68eb38fdb/src/tox/tox_env/api.py#L171-L185

In single.py -> _evaluate (which calls `ensure_setup`) this works already perfectly... if the the platform does not match, the `Skip` exception gets raised, and then caught and a message is printed... (analguous to not found interpreters) - all good!

https://github.com/tox-dev/tox/blob/34ef0e58d044782274a8a1fac6c43ad35994bb39/src/tox/session/cmd/run/single.py#L34-L45

But the `ensure_setup` method has also a couple of other callers.... like

https://github.com/tox-dev/tox/blob/fd915cc8481265da415b039f8464a2437737060c/src/tox/tox_env/python/virtual_env/package/api.py#L103-L108

and

https://github.com/tox-dev/tox/blob/fd915cc8481265da415b039f8464a2437737060c/src/tox/tox_env/python/virtual_env/package/api.py#L114-L120

and

https://github.com/tox-dev/tox/blob/ec50a6f2ccdf83a070d0037d7e685aa3c843821e/src/tox/provision.py#L96-L115

If there the `raise Skip` from within `check_platform` within `ensure_setup` gets triggered, nobody's there to handle it.

That's why I'd prefer to call `check_platform` directly in the try block of `_evaluate` - as there is the only place the `Skip` exception gets properly handled.

like this...
```
def _evaluate(tox_env: RunToxEnv, recreate: bool, no_test: bool) -> Tuple[bool, int, List[Outcome]]:
    skipped = False
    code: int = 0
    outcomes: List[Outcome] = []
    try:
        try:
            tox_env.check_platform()  # may raise Skip exception, but that's ok as it gets handled right below
            tox_env.ensure_setup(recreate=recreate)
            code, outcomes = run_commands(tox_env, no_test)
        except Skip as exception:
            LOGGER.warning("skipped environment because %s", exception)
            skipped = True
```

gaborbernat: Platform match is only enabled for the run environment. Neither packaging nor provisioning environments can't set that, so we're safe.

jugmac00: Very well.

So, what's missing?

Rebase to please codecov?

Squash locally and forcepush?

gaborbernat: We need to ensure (add tests for it) that my statement above holds 😊

jugmac00: So, I rebased.

As I am not yet familiar with the other packages, I then had a look at the coverage report, which tests call the packaging and provisioning parts, took two as a template, tried to strip them down, so I can just prove your statement from above... worked out fine on my Ubuntu laptop, but now it fails in CI, but only on Ubuntu / Python 3.8 and Windows / Python 3.7

The failures on above platform combinations seem odd.

The one error comes from `test_provision_requires_ok` which I took as a template. I can't see what's the problem. Maybe an isolation issue? Is it safe to use `pypi_index_self.use(monkeypatch)` in two tests?  I am not familiar with the monkeypatch fixture - do we need to clean up later manually?

Now, this is interesting... ` ERROR at setup of test_provisioning_platform_check_does_not_raise_unhandled_skip_exception` - so it is not the test, it is already failing in the setup :-/

I could def. need a helping hand.

jugmac00: Maybe it is the session scope of the fixture which is causing the problem.

```
@pytest.fixture(scope="session")
def pypi_index_self(pypi_server: IndexServer, tox_wheels: List[Path], demo_pkg_inline_wheel: Path) -> Index:
```

jugmac00:
Oh my... After "Ubuntu / Python 3.8 and Windows / Python 3.7" failed in the last run, now Mac 3.7 is failing... 

I am running out of ideas. :white_flag:

gaborbernat:
> Oh my... After "Ubuntu / Python 3.8 and Windows / Python 3.7" failed in the last run, now Mac 3.7 is failing...

I Will take a look at this weekend but I've been overrun with work stuff this week, so had not availability on looking at this.

gaborbernat:
So the issue here seems there are two things building the host wheel at the same time and they are stepping on each others toe, one is the `tox_wheel`, not sure yet about the other...

gaborbernat:
Ah wait I think I know, tox_wheel is conflicting with itself due to having to two tests using it, https://github.com/pytest-dev/pytest-xdist#making-session-scoped-fixtures-execute-only-once

gaborbernat (approved):
Thanks for your contribution, much appreciated.

jugmac00:
@gaborbernat Thanks so much for the guidance and especially for the "finishing touch" 😬😂

PR #1971, opened 2021-03-11
Fix typos, modernize and simplify some code

eumiro:
A few independent commits for the tox4 rewrite in five commits:

1. fix some typos
2. push minimal required Python version from 3.5 to 3.6
3. simplify code to parse string into smaller parts (here the flake8 CI checker dislikes `str.rpartition` and thinks it's a typo)
4. simplify a nested condition
5. remove one unused variable

### DIFF
--- a/src/tox/config/loader/api.py
+++ b/src/tox/config/loader/api.py
@@ -16,14 +16,10 @@ if TYPE_CHECKING:
 
 class Override:
     def __init__(self, value: str) -> None:
-        split_at = value.find("=")
-        if split_at == -1:
+        key, equal, self.value = value.partition("=")
+        if not equal:
             raise ArgumentTypeError(f"override {value} has no = sign in it")
-        key = value[:split_at]
-        ns_at = key.find(".")
-        self.namespace = "" if ns_at == -1 else key[:ns_at]
-        self.key = key[ns_at + 1 :]
-        self.value = value[split_at + 1 :]
+        self.namespace, _, self.key = key.rpartition(".")
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}('{self}')"
--- a/src/tox/config/loader/ini/factor.py
+++ b/src/tox/config/loader/ini/factor.py
@@ -19,11 +19,7 @@ def filter_for_env(value: str, name: Optional[str]) -> str:
                 overall.append(content)
         else:
             for group in factors:
-                for a_name, negate in group:
-                    contains = a_name in current
-                    if not ((contains is True and negate is False) or (contains is False and negate is True)):
-                        break
-                else:
+                if all((a_name in current) ^ negate for a_name, negate in group):
                     overall.append(content)
     result = "\n".join(overall)
     return result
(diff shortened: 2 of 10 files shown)


src/tox/config/loader/api.py
```diff
@@ -16,17 +16,13 @@
...
-        if split_at == -1:
+        key, equal, self.value = value.partition("=")
+        if not equal:
             raise ArgumentTypeError(f"override {value} has no = sign in it")
-        key = value[:split_at]
-        ns_at = key.find(".")
-        self.namespace = "" if ns_at == -1 else key[:ns_at]
-        self.key = key[ns_at + 1 :]
-        self.value = value[split_at + 1 :]
+        self.namespace, _, self.key = key.rpartition(".")
```
gaborbernat, line 22: you need to add this to whitelist.txt at root

src/tox/config/loader/api.py
```diff
@@ -16,17 +16,13 @@
...
 
     def __repr__(self) -> str:
-        return f"{self.__class__.__name__}('{self}')"
+        return f"{self.__class__.__name__}({self!r})"
```
gaborbernat, line 25: revert this, this generates an infinite stack, you're essentially calling yourself

src/tox/config/loader/ini/factor.py
```diff
@@ -19,11 +19,7 @@ def filter_for_env(value: str, name: Optional[str]) -> str:
                 overall.append(content)
         else:
             for group in factors:
-                for a_name, negate in group:
-                    contains = a_name in current
-                    if not ((contains is True and negate is False) or (contains is False and negate is True)):
-                        break
-                else:
+                if all((a_name in current) ^ negate for a_name, negate in group):
```
gaborbernat, line 22: This is one of those now it's shorter but harder to read. I mean I'm not against keeping this, but just looking at it I'm not 100% it's correct, though I trust the test suite.

src/tox/util/pep517/backend.py
```diff
@@ -23,11 +23,11 @@ def __init__(self, backend_module, backend_obj):
     def __call__(self, name, *args, **kwargs):
```
gaborbernat, line 23: Revert changes on this file. This needs to be python 2 compatible to allow python 2 backends. See https://github.com/eumiro/tox/blob/fix-typos/.pre-commit-config.yaml#L20-L21 pinning it so.

gaborbernat (changes requested):
Most are good, some changes need to be reverted.

src/tox/config/loader/api.py
eumiro: Okay, thanks.

src/tox/config/loader/api.py
eumiro: Oh, that was a bad idea. Reverted, thank you.

src/tox/config/loader/ini/factor.py
eumiro: Please decide what fits you. From my point of view, my proposal is slightly slower to read per word but more compact and therefore needs less time and mental energy to understand. But that's my point, so please feel free to decide.

src/tox/util/pep517/backend.py
eumiro: I was actually wondering why there was so much `.format` code left. I'll drop this commit.

gaborbernat (approved):

PR #1976, opened 2021-03-15
Document how to replace deprecated indexserver configuration directive.

zejn:
The `install_command` directive lists it can configure index server, but the `indexserver` directive does not point to that. This PR adds this link.

I have verified the documentation builds ok with `tox -e docs`.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -541,7 +541,7 @@ Complete list of settings that you can put into ``testenv*`` sections:
     to configure PIP_INDEX_URL environment variable, see below.
 
     Multi-line ``name = URL`` definitions of python package servers.
-    Dependencies can specify using a specified index server through the
+    You can specify an alternative index server for dependencies by applying the
     ``:indexservername:depname`` pattern.  The ``default`` indexserver
     definition determines where unscoped dependencies and the sdist install
     installs from.  Example:
@@ -555,7 +555,7 @@ Complete list of settings that you can put into ``testenv*`` sections:
     will make tox install all dependencies from this PyPI index server
     (including when installing the project sdist package).
 
-    The recommended way to set custom index server URL is to use :conf:`setenv`:
+    The recommended way to set a custom index server URL is to use :conf:`setenv`:
 
     .. code-block:: ini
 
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -537,9 +537,11 @@ Complete list of settings that you can put into ``testenv*`` sections:
 
     .. versionadded:: 0.9
 
-    (DEPRECATED, will be removed in a future version) Multi-line ``name =
-    URL`` definitions of python package servers.  Dependencies can
-    specify using a specified index server through the
+    (DEPRECATED, will be removed in a future version) Use
```
gaborbernat, line 540: IMHO a better way to specify the index server is to do:

```
setenv =
    PIP_INDEX_URL = {env:PIP_INDEX_URL:https://magic.com/simple}
```

gaborbernat (changes requested):

gaborbernat (approved):

docs/config.rst
zejn: I agree.

I've changed the commit so there's only a note at the top, following is the description of indexserver configuration option and last is the new recommendation.

I've changed magic.com/simple to pypi.org/simple to avoid someone getting magic.com. (example.com could also be used).

docs/config.rst
```diff
@@ -553,6 +555,17 @@ Complete list of settings that you can put into ``testenv*`` sections:
     will make tox install all dependencies from this PyPI index server
     (including when installing the project sdist package).
 
+    The recommended way to set custom index server URL is to use :conf:`setenv`:
```
jugmac00, line 558: ... to set a custom index server ...

docs/config.rst
```diff
@@ -553,6 +555,17 @@ Complete list of settings that you can put into ``testenv*`` sections:
...
+
+    .. code-block:: ini
+
+        [testenv]
+        setenv =
+            PIP_INDEX_URL = {env:PIP_INDEX_URL:https://pypi.org/simple/}
+
+    This will ensure the desired index server gets used for virtual environment
+    creation and allow overriding index server URL with environment variable.
```
jugmac00, line 567: creation and allow overriding the index server URL with an environment variable.

docs/config.rst
```diff
@@ -537,9 +537,11 @@ Complete list of settings that you can put into ``testenv*`` sections:
...
     .. versionadded:: 0.9
 
-    (DEPRECATED, will be removed in a future version) Multi-line ``name =
-    URL`` definitions of python package servers.  Dependencies can
-    specify using a specified index server through the
+    (DEPRECATED, will be removed in a future version) Use :conf:`setenv`
+    to configure PIP_INDEX_URL environment variable, see below.
+
+    Multi-line ``name = URL`` definitions of python package servers.
+    Dependencies can specify using a specified index server through the
```
jugmac00, line 544:
```suggestion
    You can specify an alternative index server for dependencies by applying the
```

PR #1984, opened 2021-03-30
Allow building with setuptools_scm 6+

hroncok:
Once again a new setuptools_scm version is available, 6.
Lets' not change this to `<7` again but allow any future version.
If it breaks in the future, we can adapt.

### DIFF
--- a/docs/changelog/1984.misc.rst
+++ b/docs/changelog/1984.misc.rst
@@ -1,0 +1,1 @@
+Enable building tox with ``setuptools_scm`` 6+ by :user:hroncok
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "setuptools >= 40.0.4",
-    "setuptools_scm >= 2.0.0, <6",
+    "setuptools_scm >= 2.0.0",
     "wheel >= 0.29.0",
 ]
 build-backend = 'setuptools.build_meta'


gaborbernat (approved):

docs/changelog/1984.misc.rst
```diff
@@ -0,0 +1 @@
+Enable building tox with ``setuptools_scm`` 6+ by :user:hroncok
```
webknjaz, line 1: @hroncok you forgot the backticks after `:user:` :point_up:

hroncok: Oh, not again. If only a machine could tell me this before this was merged :thinking:

hroncok: https://github.com/tox-dev/tox/pull/1989

webknjaz: */me needs to teach chronographer parse RST maybe*

gaborbernat: Even simpler, https://github.com/tox-dev/tox/blob/master/.pre-commit-config.yaml#L56-L66 can have a simple regex rule in there 🤷🏻

webknjaz: @gaborbernat I also thought so and I'm trying to explore this. But then I started to think about cases when the person doesn't want to mark themselves as an author and it got rather complicated...

gaborbernat: It doesn't have to be a must-have field, but just check for typos with it 🤔

webknjaz: Yeah.. You're right. I overengineered it, I guess: https://github.com/tox-dev/tox/pull/1990.

PR #1988, opened 2021-03-31
Fix legacy list env command for empty/missing envlist

### LINKED ISSUE #1987: tox4: cannot handle empty envlist for `tox -l`
```
❯ tox4 -l
  File "/home/jugmac00/.local/bin/tox4", line 8, in <module>
    sys.exit(run())
  File "/home/jugmac00/.local/lib/python3.6/site-packages/tox/run.py", line 22, in run
  File "/home/jugmac00/.local/lib/python3.6/site-packages/tox/run.py", line 47, in main
    result = handler(state)
  File "/home/jugmac00/.local/lib/python3.6/site-packages/tox/session/cmd/legacy.py", line 98, in legacy
    return list_env(state)
  File "/home/jugmac00/.local/lib/python3.6/site-packages/tox/session/cmd/list_env.py", line 27, in list_env
    max_length = max(len(env) for env in (default.envs + extra))
ValueError: max() arg is an empty sequence
```

## reproduction

Just delete the complete content of `tox.ini` on the rewrite branch.

(I noticed this problem when trying to reproduce #1985)

Interestingly (for me), `tox4 list` does not crash... 

TIL: `tox -l` and `tox list` are not aliases => `-l` and `--listenvs` are legacy options, and not the same as `list`

```
❯ tox4 list
additional environments:
py -> [no description]
```
(btw what is `py`??? and why is there an entry, when there is no additional environment)

jugmac00:
Calling `tox -l` for a tox.ini without an envlist or with an empty
envlist resulted in an error.

Now, it prints nothing - just as tox version 3 did.

This fixes #1987

## Contribution checklist:

- [x] added/updated test(s)
      in message body


tests/session/cmd/test_legacy.py
```diff
@@ -28,6 +28,21 @@ def test_legacy_list_default(tox_project: ToxProjectCreator, mocker: MockerFixtu
     assert outcome.state.options.show_core is False
 
 
+@pytest.mark.parametrize(
+    "configuration",
+    [
+        pytest.param("", id="missing envlist"),
+        pytest.param("[toxenv]", id="empty envlist"),
```
gaborbernat, line 35: Both of these are missing. An empty env list would look like this:
`[toxenv]\nenv_list=`

gaborbernat (changes requested):
Add an actually empty env list test case too 👍🏻 Also worthwhile adding another test when invoking this command without an existing `tox.ini` available. Otherwise looks good.

src/tox/session/cmd/list_env.py
```diff
@@ -24,7 +24,7 @@ def list_env(state: State) -> int:
 
     if not option.list_no_description and default:
         print("default environments:")
-    max_length = max(len(env) for env in (default.envs + extra))
+    max_length = max(len(env) for env in (default.envs + extra)) if default.envs + extra else 0
```
gaborbernat, line 27: better to use
```python
max_length = max((len(env) for env in (default.envs + extra)), default=0)
```

jugmac00: Awesome! I learned something new!

jugmac00:
Thanks for the review - I will push the amendments tonight!

tests/session/cmd/test_legacy.py
```diff
@@ -28,6 +28,32 @@ def test_legacy_list_default(tox_project: ToxProjectCreator, mocker: MockerFixtu
...
+def test_legacy_list_env_with_empty_or_missing_env_list(tox_project: ToxProjectCreator, configuration: str) -> None:
+    """we want to stay backwards compatible with tox 3 and show no output"""
+    outcome = tox_project({"tox.ini": configuration}).run("le", "-l")
+
+    outcome.assert_success()
+    outcome.assert_out_err("", "")
+
+
+def test_legacy_list_env_with_no_tox_file(tox_project: ToxProjectCreator) -> None:
+    project = tox_project({"pytest.ini": ""})
```
gaborbernat, line 48: You can leave the dictionary here empty.

tests/session/cmd/test_legacy.py
```diff
@@ -28,6 +28,32 @@ def test_legacy_list_default(tox_project: ToxProjectCreator, mocker: MockerFixtu
...
+    outcome = project.run("le", "-l")
+
+    outcome.assert_success()
+    outcome.assert_out_err(
+        f"ROOT: No tox.ini or setup.cfg or pyproject.toml found, assuming empty tox.ini at {project.path}\n", ""
```
gaborbernat, line 53: save the message into a variable to avoid this splitting over 3 lines

gaborbernat (changes requested):

tests/session/cmd/test_legacy.py
jugmac00: Ah, thanks! Thanks, I first got an "positional argument files ..." error, but I must have missed the empty dict before.

gaborbernat (approved):

PR #1990, opened 2021-04-01
Add a check for broken use of :user: in RST

webknjaz:
This change attempts to improve the linting of the `:user:` role in RST files.

Ref https://github.com/tox-dev/tox/pull/1984#discussion_r605646689

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,7 +67,6 @@ repos:
       - id: changelogs-user-role
         name: changelog files have a non-broken :user:`name` role
         language: pygrep
-        entry: >-
-          :user:([^`]+`?|`[^`]+[\s,])
+        entry: :user:([^`]+`?|`[^`]+[\s,])
         pass_filenames: true
         types: [file, rst]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -64,3 +64,28 @@ repos:
         exclude: >-
           ^docs/changelog/(\d+\.(bugfix|feature|deprecation|breaking|doc|misc).rst|README.rst|template.jinja2)
         files: ^docs/changelog/
+      - id: changelogs-user-role
+        name: changelog files have a non-broken :user:`name` role
+        language: system
+        entry: >-
+          sh -c '
+          rc=0;
+          for f in "$@";
+          do
+          if ! >/dev/null grep ":user:" -- "${f}";
+          then
+          continue;
+          fi;
+          grep ":user:" -- "${f}"
+          | sed "s/.*\(:user:[^ .,]\+\).*$/\1/"
+          | >/dev/null grep -v "^:user:\`[^ \`]\+\`";
+          if [ "$?" -eq 0 ];
+          then
+          rc=1;
+          2>&1 echo ":user: RST role syntax is broken in '${f}'...";
+          fi;
+          done;
+          exit $rc
+          '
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -64,3 +64,28 @@ repos:
         exclude: >-
           ^docs/changelog/(\d+\.(bugfix|feature|deprecation|breaking|doc|misc).rst|README.rst|template.jinja2)
         files: ^docs/changelog/
+      - id: changelogs-user-role
+        name: changelog files have a non-broken :user:`name` role
+        language: system
+        entry: >-
+          sh -c '
```
gaborbernat, line 71: Oh, we can't accept any Linux checks. Use python to make it cross platform.

gaborbernat (changes requested):

.pre-commit-config.yaml
webknjaz: I'm trying to figure out if I can come up with a pygrep check but can't think of a solution. `--negate` isn't helpful. Any ideas?

asottile: maybe this?

```
args: [--multiline, --negate]
entry: :user:`[^`]+`
```

webknjaz: Alright... I think I came up with something working.

webknjaz: > maybe this?

No, this matches files not having `:user:` and we agreed not to enforce this. FWIW I've committed the solution that works with different tests I came up with.

.pre-commit-config.yaml
```diff
@@ -64,3 +64,10 @@ repos:
         exclude: >-
           ^docs/changelog/(\d+\.(bugfix|feature|deprecation|breaking|doc|misc).rst|README.rst|template.jinja2)
         files: ^docs/changelog/
+      - id: changelogs-user-role
+        name: changelog files have a non-broken :user:`name` role
+        language: pygrep
+        entry: >-
+          :user:([^`]+`?|`[^`]+[\s,])
```
asottile, line 71: there's no reason to bring `>-` into this imo -- just do it inline

also this is probably just as good as the more complicate regex:

```
entry: ':user:[^`]'
```

webknjaz: @asottile my regex
```
docs/changelog.rst:25:  provision metadata are written as JSON to that path - by :user:
docs/changelog.rst:26:  provision metadata are written as JSON to that path - by :user:hroncok`
docs/changelog.rst:27:  provision metadata are written as JSON to that path - by :user:hroncok
docs/changelog.rst:28:  provision metadata are written as JSON to that path - by :user:`hroncok
```
your regex
```
docs/changelog.rst:25:  provision metadata are written as JSON to that path - by :user:
docs/changelog.rst:26:  provision metadata are written as JSON to that path - by :user:hroncok`
docs/changelog.rst:27:  provision metadata are written as JSON to that path - by :user:hroncok
```

webknjaz: > there's no reason to bring `>-` into this imo -- just do it inline

we clearly prefer different styles :man_shrugging:

gaborbernat: I'd also say make it inline but use the more advanced regex 🙏

webknjaz: Done.

gaborbernat (approved):

PR #1997, opened 2021-04-05
Refactor: Simplify and clean up the code

eumiro:
Several independent refactoring improvements simplifying and cleaning up the code.

### DIFF
--- a/src/tox/config/of_type.py
+++ b/src/tox/config/of_type.py
@@ -2,6 +2,7 @@
 Group together configuration values that belong together (such as base tox configuration, tox environment configs)
 """
 from abc import ABC, abstractmethod
+from itertools import product
 from typing import TYPE_CHECKING, Any, Callable, Generic, Iterable, List, Mapping, Optional, Type, TypeVar, Union, cast
 
 from tox.config.loader.api import Loader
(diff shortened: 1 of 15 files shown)


src/tox/config/of_type.py
```diff
@@ -1,6 +1,7 @@
 """
 Group together configuration values that belong together (such as base tox configuration, tox environment configs)
 """
+import itertools as it
```
gaborbernat, line 4: `from itertools import product`

tests/session/cmd/test_sequential.py
```diff
@@ -88,15 +88,15 @@ def test_result_json_sequential(
     packaging_test = get_cmd_exit_run_id(log_report, ".pkg", "test")
     assert packaging_test == [(None, "build_wheel")]
     packaging_installed = log_report["testenvs"][".pkg"].pop("installed_packages")
-    assert {i[: i.find("==")] for i in packaging_installed} == {"pip", "setuptools", "wheel"}
+    assert {i.partition("==")[0] for i in packaging_installed} == {"pip", "setuptools", "wheel"}
```
gaborbernat, line 91: In these two cases here I find `find` easier to read and understand than partition 🤔

gaborbernat (changes requested):

src/tox/config/of_type.py
eumiro: Changed. How about importing from `itertools` in `src/tox/config/loader/ini/factor.py`?

tests/session/cmd/test_sequential.py
eumiro: First, I experimented with `assert all(i.startswith(("pip==", "setuptools==", "wheel==")) for i in packaging_installed)` but `pytest` would probably won't be very helpful if the test failed. I'll revert my change.

src/tox/config/of_type.py
gaborbernat: What do you mean?

gaborbernat (approved):
Given all CI will pass.

src/tox/config/of_type.py
eumiro: Everywhere else there's `from itertools import …`, only in that `src/tox/config/loader/ini/factor.py` there's the plain `import itertools`. I'm asking whether you'd like to have it repo-wide consistent.

gaborbernat: Oh must be an oversight, definitely let's keep it consistent.👌

eumiro: Okay, will add an extra commit to this PR.

PR #2011, opened 2021-04-10
Replace .format with f-strings

### LINKED ISSUE #2012: tox4: Prefer f-strings instead of the str.format method.
In https://tox.readthedocs.io/en/rewrite/development.html#code-style-guide it says:

- Prefer f-strings instead of the `str.format` method.

It would be nice to have more f-strings in the code.

eumiro:
Replace most of the `.format` calls with proper f-strings. Left `.format` where backslashes or complex escapes were used.

The second commit shortens the line using explicit `textwrap.shorten` instead of trivial string slicing. This may deliver different results because it cuts off whole words and not individual characters.  

Fixes #2012

### DIFF
--- a/docs/changelog/2012.bugfix.rst
+++ b/docs/changelog/2012.bugfix.rst
@@ -1,0 +1,1 @@
+Prefer f-strings instead of the str.format method - by :user:`eumiro`.
--- a/src/tox/config/cli/parser.py
+++ b/src/tox/config/cli/parser.py
@@ -236,7 +236,7 @@ class ToxParser(ArgumentParserWithEnvAndConfig):
 def add_verbosity_flags(parser: ArgumentParser) -> None:
     from tox.report import LEVELS
 
-    level_map = "|".join("{} - {}".format(c, logging.getLevelName(l)) for c, l in sorted(LEVELS.items()))
+    level_map = "|".join(f"{c} - {logging.getLevelName(l)}" for c, l in sorted(LEVELS.items()))
     verbosity_group = parser.add_argument_group("verbosity")
     verbosity_group.description = f"verbosity=verbose-quiet, default {logging.getLevelName(LEVELS[3])}, map {level_map}"
     verbosity = verbosity_group.add_mutually_exclusive_group()
--- a/src/tox/config/of_type.py
+++ b/src/tox/config/of_type.py
@@ -109,7 +109,7 @@ class ConfigDynamicDefinition(ConfigDefinition[T]):
 
     def __repr__(self) -> str:
         values = ((k, v) for k, v in vars(self).items() if k != "post_process" and v is not None)
-        return f"{type(self).__name__}({', '.join('{}={}'.format(k, v) for k,v in values)})"
+        return f"{type(self).__name__}({', '.join(f'{k}={v}' for k, v in values)})"
 
     def __eq__(self, o: Any) -> bool:
         return (
--- a/src/tox/pytest.py
+++ b/src/tox/pytest.py
@@ -151,7 +151,7 @@ class ToxProject:
             elif isinstance(value, str):
                 at_path.write_text(textwrap.dedent(value))
             else:
-                msg = "could not handle {} with content {!r}".format(at_path / key, value)  # pragma: no cover
+                msg = f"could not handle {at_path / key} with content {value!r}"  # pragma: no cover
                 raise TypeError(msg)  # pragma: no cover
 
     def patch_execute(self, handle: Callable[[ExecuteRequest], Optional[int]]) -> MagicMock:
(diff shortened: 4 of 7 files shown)


jugmac00 (changes requested):
@eumiro Thank you for the PR.

Please have a look at our developer documentation. You need to create a changelog entry.

https://tox.readthedocs.io/en/rewrite/development.html#changelog-entries

src/tox/util/pep517/backend.py
```diff
@@ -23,11 +23,11 @@ def __init__(self, backend_module, backend_obj):
     def __call__(self, name, *args, **kwargs):
         on_object = self if name.startswith("_") else self.backend
         if not hasattr(on_object, name):
-            raise MissingCommand("{!r} has no attribute {!r}".format(on_object, name))
+            raise MissingCommand(f"{on_object!r} has no attribute {name!r}")
```
gaborbernat, line 26: The backend needs to run under python 2 too.

tests/demo_pkg_inline/build.py
```diff
@@ -8,17 +8,17 @@
 pkg_name = name.replace("_", "-")
 
 version = "1.0.0"
-dist_info = "{}-{}.dist-info".format(name, version)
-logic = "{}/__init__.py".format(name)
-metadata = "{}/METADATA".format(dist_info)
-wheel = "{}/WHEEL".format(dist_info)
-record = "{}/RECORD".format(dist_info)
+dist_info = f"{name}-{version}.dist-info"
+logic = f"{name}/__init__.py"
```
gaborbernat, line 12: This file needs to run under python 2 too.

tests/execute/local_subprocess/test_local_subprocess.py
```diff
@@ -105,14 +105,12 @@ def test_local_execute_write_a_lot(os_env: Dict[str, str]) -> None:
...
             "-c",
-            (
-                "import sys; import time; from datetime import datetime; import os;"
-                "print('e' * {0}, file=sys.stderr);"
-                "print('o' * {0}, file=sys.stdout);"
-                "time.sleep(0.5);"
-                "print('a' * {0}, file=sys.stderr);"
-                "print('b' * {0}, file=sys.stdout);"
-            ).format(count),
+            f"import sys; import time; from datetime import datetime; import os;"
```
gaborbernat, line 108: This too I think 🤔

eumiro:
> Some files still need to be python 2 compatible, to allow and test for running python 2 envs from tox.

What would be the best way to tag these files compatible with both 2/3? Something in the header?

gaborbernat:
That can be, currently they're tagged in my head and hopefully by the test suite.

eumiro:
Reverted changes in Py2/3 files, added changelog entry.

And, @gaborbernat, I'd like to find a standard way of marking files to be Py2/3 compatible to avoid contributors from keeping fixing the Py2 relics from them.

gaborbernat (approved):

gaborbernat:
> Reverted changes in Py2/3 files, added changelog entry.
> 
> And, @gaborbernat, I'd like to find a standard way of marking files to be Py2/3 compatible to avoid contributors from keeping fixing the Py2 relics from them.

Can't think of any better than mention this in the module docstring...

eumiro:
There's one formatting test failing, can have a look at it later.

And created #2013 to discuss the Py2/3 problem. Module docstring is a good idea.

eumiro:
The failed tests come from the usage of `textwrap.shorten`. Now the question is, how do you want to handle the words in the output? Should it cut after X characters like before or should it show whole words and possibly cut earlier?

gaborbernat:
> The failed tests come from the usage of `textwrap.shorten`. Now the question is, how do you want to handle the words in the output? Should it cut after X characters like before or should it show whole words and possibly cut earlier?

As long as the ... are added I'm happy to cut it at the word boundary.

tests/execute/local_subprocess/test_local_subprocess.py
eumiro: Looks like this file could be modernized to Python 3 as it was not mentioned in #2015 ? :thinking:

gaborbernat: It does look like 🤔 unless some tests fail in which case we need to exclude it from pyupgrade 👍🏻 But I think it's alright.

gaborbernat:
@eumiro in light of your recent PRs I've invited you to the maintainers' group 👏🏻 No pressure though, you can keep contributing as much or as little you want, this just gives you some more permissions to review/approve/triage. Thanks for all your very helpful feedback and effort to the project 🙇🏻.

jugmac00 (approved):

eumiro:
Thank you for your active feedback, @gaborbernat and @jugmac00, and also for the opportunity to take part in the future of tox4!

PR #2014, opened 2021-04-10
Add Python3.10 to the test suite

gaborbernat:


eumiro:
Should this list of supported Python versions be updated too or would you add a reference to `setup.cfg` as well?

https://github.com/tox-dev/tox/blob/rewrite/docs/installation.rst#python-and-os-compatibility

gaborbernat:
It should.

eumiro:
Okay, added in #2017

PR #2015, opened 2021-04-10
Docs: define which files to keep py2 compatible

### LINKED ISSUE #2013: tox4: Mark files that have to be compatible with Python 2
Some files have to be kept compatible with Python 2.

To prevent premature modernization of code to Python 3.6+, mark these files as Python 2 compatible. Non-marked files are supposed to follow the current compatibility listed in `setup.cfg` (currently Python 3.6–3.9).

eumiro:
Code style tells which Python versions code syntax should be used in which files, referring to the relevant configuration files (single point of truth docs).

Fixes #2013

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -96,6 +96,11 @@ Code style guide
 
 - First and foremost, the linters configured for the project must pass; this generally means following PEP-8 rules,
   as codified by: ``flake8``, ``black``, ``isort``, ``pyupgrade``.
+- The supported Python versions (and the code syntax to use) are listed in the ``setup.cfg`` file
+  in the ``options/python_requires`` entry. However, there are some files that have to be kept compatible
+  with Python 2.7 to allow and test for running Python 2 envs from tox. They are listed in ``.pre-commit-config.yaml``
+  under ``repo: https://github.com/asottile/pyupgrade`` under ``hooks/exclude``.
+  Please do not attempt to modernize them to Python 3.x.
 - Packaging options should be specified within ``setup.cfg``; ``setup.py`` is only kept for editable installs.
 - All code (tests too) must be type annotated as much as required by ``mypy``.
 - We use a line length of 120.
--- a/src/tox/util/pep517/backend.py
+++ b/src/tox/util/pep517/backend.py
@@ -1,4 +1,8 @@
-"""Handles communication on the backend side between frontend and backend"""
+"""Handles communication on the backend side between frontend and backend
+
+Please keep this file Python 2.7 compatible.
+See https://tox.readthedocs.io/en/rewrite/development.html#code-style-guide
+"""
 from __future__ import print_function, unicode_literals
 
 import json
--- a/tests/demo_pkg_inline/build.py
+++ b/tests/demo_pkg_inline/build.py
@@ -1,3 +1,8 @@
+"""
+Please keep this file Python 2.7 compatible.
+See https://tox.readthedocs.io/en/rewrite/development.html#code-style-guide
+"""
+
 import os
 import sys
 import tarfile


jugmac00:
Why don't you just mention this in the module docstrings?

https://github.com/tox-dev/tox/pull/2011#issuecomment-817106808

I also think this would be the best place. If somebody would try to update the code to Python 3 anyway, the test suite will fail (at least I hope so).

eumiro:
Added the information to the docstring of the two respective files.

docs/development.rst
```diff
@@ -96,6 +96,10 @@ Code style guide
 
 - First and foremost, the linters configured for the project must pass; this generally means following PEP-8 rules,
   as codified by: ``flake8``, ``black``, ``isort``, ``pyupgrade``.
+- The supported Python versions (and the code syntax to use) are listed in the ``tox.ini`` file in the ``tox/envlist``
+  section as ``py3*``. However, there are some files that have to be kept compatible with Python 2.7. They are
+  listed in ``.pre-commit-config.yaml`` under ``repo: https://github.com/asottile/pyupgrade`` under ``hooks/exclude``.
+  Please do not attempt to modernize them to Python 3.x.
```
jugmac00, line 102: Please add the reason why we need to keep the files Python 2 compatible:
https://github.com/tox-dev/tox/pull/2011#pullrequestreview-632904062

src/tox/util/pep517/backend.py
```diff
@@ -1,4 +1,7 @@
-"""Handles communication on the backend side between frontend and backend"""
+"""Handles communication on the backend side between frontend and backend
+
+PLEASE KEEP THIS FILE PYTHON 2.7 COMPATIBLE
```
jugmac00, line 3: There is no need to use all caps. Please add a reference to the development documentation so you can read about the reasons.

tests/demo_pkg_inline/build.py
```diff
@@ -1,3 +1,5 @@
+"""PLEASE KEEP THIS FILE PYTHON 2.7 COMPATIBLE"""
```
jugmac00, line 1: There is no need to use all caps. Please add a reference to the development documentation so you can read about the reasons.

docs/changelog/2013.doc.rst
```diff
@@ -0,0 +1 @@
+Update code style to mention which Python versions code syntax should be used in which files - by :user:`eumiro`.
```
jugmac00, line 1:
```suggestion
Add information why certain files need to stay Python 2 compatible - by :user:`eumiro`.
```

jugmac00 (changes requested):
I have some suggestions, then we are good to go.

docs/changelog/2013.doc.rst
gaborbernat: I don't think this needs a changelog, considering its not user-facing change 🤔

gaborbernat (changes requested):

eumiro:
Okay, I think now all the suggested changes are in.

docs/development.rst
```diff
@@ -96,6 +96,12 @@ Code style guide
 
 - First and foremost, the linters configured for the project must pass; this generally means following PEP-8 rules,
   as codified by: ``flake8``, ``black``, ``isort``, ``pyupgrade``.
+- The supported Python versions (and the code syntax to use) are listed in the ``tox.ini`` file in the ``tox/envlist``
```
gaborbernat, line 99: This is not true. The only authoritative source of truth for this is the setup.cfg python_requires entry.

gaborbernat (changes requested):

gaborbernat (approved):

jugmac00 (approved):

PR #2017, opened 2021-04-11
Minor code improvements

eumiro:
Some tiny independent refactorings probably without need to be mentioned in the changelog.

### DIFF
--- a/src/tox/config/cli/env_var.py
+++ b/src/tox/config/cli/env_var.py
@@ -18,8 +18,7 @@ def get_env_var(key: str, of_type: Type[Any]) -> Optional[Tuple[Any, str]]:
     :return:
     """
     key_upper = key.upper()
-    for fmt in ("TOX_{}", "TOX{}"):
-        environ_key = fmt.format(key_upper)
+    for environ_key in (f"TOX_{key_upper}", f"TOX{key_upper}"):
         if environ_key in os.environ:
             value = os.environ[environ_key]
             try:
--- a/src/tox/session/cmd/run/common.py
+++ b/src/tox/session/cmd/run/common.py
@@ -33,7 +33,7 @@ class SkipMissingInterpreterAction(Action):
     ) -> None:
         value = "true" if values is None else values
         if value not in ("config", "true", "false"):
-            raise ArgumentError(self, f"value must be 'config', 'true', or 'false' (got {repr(value)})")
+            raise ArgumentError(self, f"value must be 'config', 'true', or 'false' (got {value!r})")
         setattr(namespace, self.dest, value)
 
 
--- a/src/tox/session/cmd/run/common.py
+++ b/src/tox/session/cmd/run/common.py
@@ -144,7 +144,7 @@ def report(start: float, runs: List[ToxEnvRunResult], is_colored: bool) -> int:
     for run in runs:
         all_good &= run.code == Outcome.OK or run.ignore_outcome
         duration_individual = [o.elapsed for o in run.outcomes]
-        extra = f"+cmd[{','.join(f'{i:.2f}' for i in duration_individual)}]" if len(duration_individual) else ""
+        extra = f"+cmd[{','.join(f'{i:.2f}' for i in duration_individual)}]" if duration_individual else ""
         setup = run.duration - sum(duration_individual)
         msg, color = _get_outcome_message(run)
         out = f"  {run.name}: {msg} ({run.duration:.2f}{f'=setup[{setup:.2f}]{extra}' if extra else ''} seconds)"
@@ -227,10 +227,7 @@ class ToxSpinner(Spinner):
 
     def update_spinner(self, result: ToxEnvRunResult, success: bool) -> None:
         if success:
-            if result.skipped:
-                done = self.skip
-            else:
-                done = self.succeed
+            done = self.skip if result.skipped else self.succeed
         else:
             done = self.fail
         done(result.name)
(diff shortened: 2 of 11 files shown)


src/tox/config/cli/parser.py
```diff
@@ -217,12 +217,7 @@ def parse_known_args(  # type: ignore[override]
             args = sys.argv[1:]
         cmd_at: Optional[int] = None
         if self._cmd is not None and args:
-            for at, arg in enumerate(args):
-                if arg in self._cmd.choices:
-                    cmd_at = at
-                    break
-            else:
-                cmd_at = None
+            cmd_at = next((at for at, arg in enumerate(args) if arg in self._cmd.choices), None)
```
jugmac00, line 220: I think the old version is easier to read, and if necessary, easier to debug.

src/tox/config/cli/env_var.py
```diff
@@ -17,9 +17,7 @@ def get_env_var(key: str, of_type: Type[Any]) -> Optional[Tuple[Any, str]]:
     :param of_type: the type we would like to convert it to
     :return:
     """
-    key_upper = key.upper()
-    for fmt in ("TOX_{}", "TOX{}"):
-        environ_key = fmt.format(key_upper)
+    for environ_key in (f"TOX_{key.upper()}", f"TOX{key.upper()}"):
```
gaborbernat, line 20: technically here you're calling upper twice, you could use a variable as was done before to do it only once

src/tox/config/loader/str_convert.py
```diff
@@ -31,8 +31,7 @@ def to_list(value: str, of_type: Type[Any]) -> Iterator[str]:
 
     @staticmethod
     def to_set(value: str, of_type: Type[Any]) -> Iterator[str]:
-        for value in StrConvert.to_list(value, of_type):
-            yield value
+        yield from StrConvert.to_list(value, of_type)
```
gaborbernat, line 34: nitpick: While this technically is a shorter version of it, I find the left side is easier to debug because you actually can put a breakpoint there, while in this case, you're out of luck

src/tox/tox_env/python/pip/pip_install.py
```diff
@@ -168,10 +168,7 @@ def build_install_cmd(self, args: Sequence[str]) -> List[str]:
             opts_at = install_command.index("{packages}")
         except ValueError:
             opts_at = len(install_command)
-        result = install_command[:opts_at]
-        result.extend(args)
-        result.extend(install_command[opts_at + 1 :])
-        return result
+        return install_command[:opts_at] + list(args) + install_command[opts_at + 1 :]
```
gaborbernat, line 171: again if you don't save the return value into a variable is a PITA to debug when something here is wrong

src/tox/session/cmd/run/common.py
```diff
@@ -33,7 +33,7 @@ def __call__(
     ) -> None:
         value = "true" if values is None else values
         if value not in ("config", "true", "false"):
-            raise ArgumentError(self, f"value must be 'config', 'true', or 'false' (got {repr(value)})")
+            raise ArgumentError(self, f"value must be 'config', 'true', or 'false' (got {value!r})")
```
jugmac00, line 36: Why is the new version better than the old one?

src/tox/session/cmd/run/common.py
```diff
@@ -227,10 +227,7 @@ def __init__(self, enabled: bool, state: State, total: int) -> None:
 
     def update_spinner(self, result: ToxEnvRunResult, success: bool) -> None:
         if success:
-            if result.skipped:
-                done = self.skip
-            else:
-                done = self.succeed
+            done = self.skip if result.skipped else self.succeed
         else:
```
jugmac00, line 231: I like the old one better.

src/tox/tox_env/python/pip/req/file.py
```diff
@@ -333,7 +333,7 @@ def _break_args_options(line: str) -> Tuple[str, str]:
         args = []
         options = tokens[:]
         for token in tokens:
-            if token.startswith("-") or token.startswith("--"):
+            if token.startswith("-"):
```
jugmac00, line 336: While this won't make a difference, maybe the old version showed the intention better? hm... not sure.

jugmac00 (commented):
see comments; shorter is not always better

src/tox/tox_env/python/pip/req/file.py
eumiro: First I rewrote this to equivalent `token.startswith(("-", "--"))` but then saw its redundancy. The original intention could be written in a comment WHY we expect both `-` and `--`. Otherwise it's to be re-read to search for a typo.

src/tox/session/cmd/run/common.py
eumiro: It's more about formatting a string than about calling some external function.

gaborbernat: I'm ok with the new version.

src/tox/session/cmd/run/common.py
gaborbernat: I'm ok with the new version.

eumiro:
Reverted/adjusted all your comments, thank you!

(although some reverted steps will itch me next time again :-) )

gaborbernat (approved):

src/tox/tox_env/python/pip/pip_install.py
jugmac00: This was not reverted, but set resolved? 🤔

gaborbernat (approved):

PR #2025, opened 2021-04-12
Replace `chain.from_iterable` with double comprehensions

eumiro:
This is just a follow-up of an off-record discussion with @jugmac00 about `itertools.chain.from_iterable`.

Maybe you'll like the double comprehensions syntax more.

### DIFF
--- a/src/tox/config/loader/ini/factor.py
+++ b/src/tox/config/loader/ini/factor.py
@@ -2,14 +2,12 @@
 Expand tox factor expressions to tox environment list.
 """
 import re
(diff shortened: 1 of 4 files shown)


gaborbernat (commented):
I personally like better the chain from iterable, but 🤷🏻 not too picky about it.

jugmac00:
Well, I asked you not to create a PR for this on Signal. `chain.from_iterable` was pretty new to me as I spent way to many years in Zope 2 and Python 2 land, but certainly a double list comprehension does not make it any better / or worse.

PR #2027, opened 2021-04-13
Evaluate factor conditions for `command` keys

### LINKED ISSUE #2002: tox4: Factor conditions are not considered for `commands`
Running tox v4 against the following tox.ini currently fails: https://github.com/pypa/virtualenv/blob/0262fa6d9a0f8b45c80e6b2f82109ab1250ec336/tox.ini#L45

It fails with the following error:

```
...
py39: commands[4]> python -m coverage xml -o /home/ashley/workspace/virtualenv/.tox/4/coverage.py39.xml
py39: commands[5]> python -m coverage html -d /home/ashley/workspace/virtualenv/.tox/4/py39/tmp/htmlcov '!py34:' --show-contexts --title virtualenv-py39-coverage
No source for code: '/home/ashley/workspace/virtualenv/!py34:'.
Aborting report output, consider using -i.
py39: exit 1 (0.12 seconds) /home/ashley/workspace/virtualenv> python -m coverage html -d /home/ashley/workspace/virtualenv/.tox/4/py39/tmp/htmlcov '!py34:' --show-contexts --title virtualenv-py39-coverage pid=471636
.pkg: _exit> python /home/ashley/envs/tox4/lib/python3.9/site-packages/tox/util/pep517/backend.py True setuptools.build_meta
  py39: FAIL code 1 (153.51=setup[11.91]+cmd[0.14,135.50,3.64,0.97,1.24,0.12] seconds)
  evaluation failed :( (153.54 seconds)
```

Running the same tests with tox v3 works fine.

jugmac00:
Due to the order how the configuration got loaded, and due to a bug in
the factor parser, factors were not identified and thus not evaluated
for `command` keys.

This has been fixed now.

Closes #2002

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/src/tox/config/loader/ini/factor.py
+++ b/src/tox/config/loader/ini/factor.py
@@ -45,7 +45,7 @@ def explode_factor(group: List[Tuple[str, bool]]) -> str:
 
 def expand_factors(value: str) -> Iterator[Tuple[Optional[Iterator[List[Tuple[str, bool]]]], str]]:
     for line in value.split("\n"):
-        match = re.match(r"^((?P<factor_expr>[\w{}.!,-]+):\s+)?(?P<content>.*?)$", line)
+        match = re.match(r"^(\s*(?P<factor_expr>[\w{}.!,-]+):\s+)?(?P<content>.*?)$", line)
         if match is None:  # pragma: no cover
             raise RuntimeError("for a valid factor regex this cannot happen")
         groups = match.groupdict()
--- a/src/tox/config/loader/ini/factor.py
+++ b/src/tox/config/loader/ini/factor.py
@@ -45,7 +45,7 @@ def explode_factor(group: List[Tuple[str, bool]]) -> str:
 
 def expand_factors(value: str) -> Iterator[Tuple[Optional[Iterator[List[Tuple[str, bool]]]], str]]:
     for line in value.split("\n"):
-        match = re.match(r"^(\s*(?P<factor_expr>[\w{}.!,-]+):\s+)?(?P<content>.*?)$", line)
+        match = re.match(r"^((?P<factor_expr>[\w{}.!,-]+):\s+)?(?P<content>.*?)$", line)
         if match is None:  # pragma: no cover
             raise RuntimeError("for a valid factor regex this cannot happen")
         groups = match.groupdict()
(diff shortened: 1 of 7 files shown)


tests/config/loader/ini/test_ini_loader.py
```diff
@@ -56,3 +57,24 @@ def test_ini_loader_strip_comments(mk_ini_conf: Callable[[str], ConfigParser], c
...
+
+
+@pytest.mark.parametrize(
+    ("env", "result"),
+    [
+        ("py35", "python -m coverage html -d cov"),
+        ("py36", "python -m coverage html -d cov\n--show-contexts"),
+    ],
+)
+def test_ini_loader_raw_with_factors(
```
gaborbernat, line 69: Feels to me this is a factor specific test, so should be https://github.com/tox-dev/tox/blob/rewrite/tests/config/loader/ini/test_factor.py

src/tox/config/loader/ini/factor.py
```diff
@@ -45,7 +45,7 @@ def explode_factor(group: List[Tuple[str, bool]]) -> str:
 
 def expand_factors(value: str) -> Iterator[Tuple[Optional[Iterator[List[Tuple[str, bool]]]], str]]:
     for line in value.split("\n"):
-        match = re.match(r"^((?P<factor_expr>[\w{}.!,-]+):\s+)?(?P<content>.*?)$", line)
+        match = re.match(r"^(\s*(?P<factor_expr>[\w{}.!,-]+):\s+)?(?P<content>.*?)$", line)
```
gaborbernat, line 48: Why do we need this part?

jugmac00: Yesterday, before I wrote tests, I setup a minimal `tox.ini`  to reproduce the problem, and I could swear I head a leading whitespace or even a linebreak for `line` (which does not seem feasible in retro as it gets split at `\n`), and the fix to get it parsed correctly was to update the regex.

As I deleted the example and both the tests and the reported problem with the `tox.ini` in `virtualenv` are already fixed without the regex update, I will revert it (later that night). :+1:

jugmac00:
I need to update the changelog, too...

gaborbernat (approved):

jugmac00:
Thanks for the review (and the solution to the fix :-D )!

Issue #2030, opened 2021-04-15
tox 3.23.0 produces logging on stdout breakind use of --showconfig, -l, -a


ssbarnea:
I realised that `tox --showconfig` "API" is broken by the fact that tox does output logging messages to stdout instead of stderr when it recreates the .tox envs.

Here you can see capture of stdout from `tox --showconfig >stdout.out` when run on a fresh clone, clearly invalid ini file. If you run it again it will produce valid content because it will no longer produce stdout output.
```
[1m.tox create: /Users/ssbarnea/c/opendev/tripleo-ci-health-queries/.tox/.tox
[0m[1m.tox installdeps: setuptools >= 41.4.0, pip >= 19.3.0, tox-bindep, tox >= 3.16.1
[0m[tox]
args = ['--showconfig']
distdir = /Users/ssbarnea/c/opendev/tripleo-ci-health-queries/.tox/dist
distshare = /Users/ssbarnea/.tox/distshare
envlist_default = ['py', 'lint', 'error-testing']
hashseed = 2551992764
homedir = /Users/ssbarnea
ignore_basepython_conflict = False
```

Ideally logging should happen only on stderr, at least for commands that are expected to be machine parseable like `-l`, `-a` and `--showconfig`. I am not sure if the other two are affected but this one is for sure.

This is a serious bug for integration with other tools as --showconfig is currently the only reliable way to query tox about its config (looking inside config fix not being one).

When we fix this we should also add a simple test to avoid regressions, still with v4 around the corner I am not sure how easy it would be to port the test for v4.

The irony is that running `tox -q --showconfig` does produce zero stdout output. That , means that printing config uses the logging level when it should not and instead use effective output to stdout, aka print as this is the effective "output of the program" not some logging information.

gaborbernat:
I don't think using the reporter is the troublesome part here. the troublesome part is that the provisioning pollutes the output. I don't see an easy solution to this though; so I can only suggest a workaround: call first `tox --help` to perform the provisioning, and then call `tox --showconfig`  this is what I've also been doing in my systems. changing to print doesn't help you because `tox -v --showconfig` might still pollute the output or similarly anything might trigger a warning message to polluting the output. The way things are I'm inclined to mark this won't fix because there's no easy solution here.

ssbarnea:
[...]

PR #2031, opened 2021-04-15
Assure listing commands can produce machine readable output

ssbarnea:
* Assures we use print() when displaying output using `-a`, `-l` or `--showconfig`, which are all commands that should not be affected by the currently configured logging level. 
* Documents the fact that `-qq` is needed if user expects to be able to parse output of these commands.
* Not updating the tests because v3 is near its EOL and testing for side-effects that happen when .tox folder does not
  exist would require significant effort (linked bug happened only when tox own env needed to be recreated).

Fixes: #2030

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -446,19 +446,21 @@ def tox_addoption(parser):
         "--showconfig",
         action="store_true",
         help="show live configuration (by default all env, with -l only default targets,"
-        " specific via TOXENV/-e)",
+        " specific via TOXENV/-e). Add -qq to ensure is machine-readable",
     )
     parser.add_argument(
         "-l",
         "--listenvs",
         action="store_true",
-        help="show list of test environments (with description if verbose)",
+        help="show list of test environments (with description if verbose),"
+        " combine with -qq to ensure output is machine-readable",
     )
     parser.add_argument(
         "-a",
         "--listenvs-all",
         action="store_true",
-        help="show list of all defined environments (with description if verbose)",
+        help="show list of all defined environments (with description if verbose),"
+        " combine with -qq to ensure output is machine-readable",
     )
     parser.add_argument(
         "-c",
--- a/src/tox/session/commands/show_config.py
+++ b/src/tox/session/commands/show_config.py
@@ -29,7 +29,10 @@ def show_config(config):
     content = StringIO()
     parser.write(content)
     value = content.getvalue().rstrip()
-    reporter.verbosity0(value)
+    # We are not using the reporter because showconfig result should not be
+    # affected by logging level as its stdout output *is* expected to
+    # be machine parseable.
+    print(value)
 
 
 def tox_envs_info(config, parser):
--- a/src/tox/session/commands/show_env.py
+++ b/src/tox/session/commands/show_env.py
@@ -10,7 +10,7 @@ def show_envs(config, all_envs=False, description=False):
     extra = [e for e in env_conf if e not in ignore] if all_envs else []
 
     if description and default:
-        report.line("default environments:")
+        report.verbosity0("default environments:")
     max_length = max(len(env) for env in (default + extra) or [""])
 
     def report_env(e):
@@ -19,14 +19,17 @@ def show_envs(config, all_envs=False, description=False):
             msg = "{} -> {}".format(e.ljust(max_length), text).strip()
         else:
             msg = e
-        report.line(msg)
+        # Avoiding reporter because that is the result of program execution,
+        # and we do not want to be affected by logging level.
+        print(msg)
 
     for e in default:
         report_env(e)
     if all_envs and extra:
         if description:
+            # breakpoint()
             if default:
-                report.line("")
-            report.line("additional environments:")
+                report.verbosity0("")
+            report.verbosity0("additional environments:")
         for e in extra:
             report_env(e)
--- a/docs/changelog/2030.bugfix.rst
+++ b/docs/changelog/2030.bugfix.rst
@@ -1,0 +1,3 @@
+``tox --showconfig``, ``-a`` and ``-l`` primary output is no longer affected
+by the logging level. Allows users to add ``-qq`` to make output machine
+parseable output (no pollution from the logging). - by :user:`ssbarnea`
--- a/src/tox/session/__init__.py
+++ b/src/tox/session/__init__.py
@@ -181,7 +181,9 @@ class Session(object):
         reporter.using(
             "tox-{} from {} (pid {})".format(tox.__version__, tox.__file__, os.getpid()),
         )
-        show_description = reporter.has_level(reporter.Verbosity.DEFAULT)
+        # We display description if verbosity was increase but ignoring value
+        # of quiet_level. Needed in order to produce machine readable output.
+        show_description = self.config.option.verbose_level > 0
         if self.config.run_provision:
             provision_tox_venv = self.getvenv(self.config.provision_tox_env)
             return provision_tox(provision_tox_venv, self.config.args)


gaborbernat (changes requested):
I'm strong -1 on this as expanded on the issue. Unless you manage to convince the other maintainers we'll close this.

src/tox/config/__init__.py
```diff
@@ -446,7 +446,7 @@ def tox_addoption(parser):
         "--showconfig",
         action="store_true",
         help="show live configuration (by default all env, with -l only default targets,"
-        " specific via TOXENV/-e)",
+        " specific via TOXENV/-e). Add -q if you plan to parse it.",
```
obestwalter, line 449: I wouldn't add that here, as it is also not mentioned in help of -a and -l. It would be nice though, if we had a dedicated little bit in the docs, which explains where tox actually makes a certain promise to produce machine readable output that doesn't change in format. AFAICT this is the case for -l and --showconfig (and maybe for -a for advanced tools?). So this could go into the docs for ... I don't know ... tool builders? 

Do you think you could find a suitable place in the docs and write a little bit about that "promise"? And there it would then need to be explained that `-q` always has to be used in to be sure that provisioning output is not polluting the parseable output.

It definitely feels a bit dirty even writing that, as the question pops up again, why any additional output is produced on stdout for commands that produce machine readable output, but here we are and tox3 is a classic case of "grown organically" and everything will be super consistent and correct in tox4 :)

src/tox/session/commands/show_config.py
```diff
@@ -29,7 +29,10 @@ def show_config(config):
     content = StringIO()
     parser.write(content)
     value = content.getvalue().rstrip()
-    reporter.verbosity0(value)
+    # We are not using the reporter because showconfig result should not be
+    # affected by logging level as its stdout output *is* expected to
+    # be machine parseable.
```
obestwalter, line 34: That explanation should then also be in the other places where this promise is made. If we have a little dedicated section in the docs about the commands that are machine readable they could all have a comment like `# log level independent output - intended for machines to read (see docs/...)`

obestwalter (changes requested):

obestwalter:
I had a closer look now and reproduced the problem. When using `requires = attrs` in the tox section, then provisioning is also triggered, when calling --showconfig, but -q will just swallow all the output, which is inconsistent with -l. I wonder, why the provisioning even happens for these activities, but this is outside of the scope of this PR and maybe things like this shouldn't even worry us anymore with tox4 on the horizon.

If we agree on -l and --showconfig "promising" to be machine readable (and maybe -a), then the behaviour is inconsistent and this PR would fix that inconsistency by also using `print`.

I would also like to see some documentation though, that explains this and documents the necessity for `-q` due to potential provisioning output, if that not already exists.

Whether the -q hint for the machine parseable should be in the help, I am not sure. I think the help is already quite verbose and most people don't care. But if it should go there, then this also should be consistent and added for all three commands.

Another approach that I would like maybe even more than this quick fix, would be to suppress any output on stdout that is not the wanted output for -l, -a and --showconfig.

The approach I would like the most would be to not do provisioning at all for those commands as I am pretty sure that it is not needed for them. But I guess that is hard or impossible in tox3, so maybe the quick fix is all we need to make everybody happy enough.

So unless I am overlooking something obvious because I just started working on tox again after a long time out, I would say let's merge this (with some extra docs though).

gaborbernat:
> . I wonder, why the provisioning even happens for these activities
> The approach I would like the most would be to not do provisioning at all for those commands as I am pretty sure that it is not needed for them. But I guess that is hard or impossible in tox3, so maybe the quick fix is all we need to make everybody happy enough. 

Here you're totally wrong. It must happen because plugins will impact both the environment list and the configuration values.

> e things like this shouldn't even worry us anymore with tox4 on the horizon

Nothing goes away with tox 4. The same issue/problem is happening there too.

> If we agree on-a, -l and --showconfig "promising" to be machine readable, then the behaviour is inconsistent and this PR would fix that inconsistency by also using `print`.

The problem is that until now the verbosity of these flags were controlled by -v/-q... changing that might have an unexpected outcome. For example, makes sense to do ``tox -l -vv`` to debug the initialization of tox.


> Another approach that I would like maybe even more than this quick fix, would be to suppress any output on stdout that is not the wanted output for -l, -a and --showconfig.

I don't think in general the extra output is not wanted, but perhaps by default is unwanted, but we don't have a good way of saying this command is expected to be machine readable. We could also print machine readble on stdout, and other on stderr 🤔 

> So unless I am overlooking something obvious because I just started working on tox again after a long time out, I would say let's merge this (with some extra docs though).

I'm tempted to leave it won't fix for tox 3 🤷🏻 tox 3 was working like this for the last two years, and only now we got just one user complaining. For those few edge cases where this matters calling ``tox -l`` twice works around the problem for now.

ssbarnea:
Please wait for me to update it to move the docs. “ We could also print machine readble on stdout, and other on stderr 🤔” is the “correct” thing to do but a pragmatic approach like this is less likely to produce undesired side effects for v3.

src/tox/session/commands/show_config.py
ssbarnea: I was not able to identify any suitable place inside the docs for mentioning this. If you really want one new section please mention exactly where it should be added inside the docs.

I am afraid that the effort to add this to the docs would be wasted because it would not be disoverable by the users that will want to use this. All of these will use the --help instead, and that is why I mentioned it there, so
you do not endup using a feature without knowing how to use it in a reliable way.

obestwalter: I also had a quick look in the docs now. I also wouldn't really know where to put it as it is. A new section for tool builders that build on top of tox, seems overkill also, so I guess having it in the command line help help, is the best place currently.

gaborbernat:
> Please wait for me to update it to move the docs. “ We could also print machine readble on stdout, and other on stderr 🤔” is the “correct” thing to do but a pragmatic approach like this is less likely to produce undesired side effects for v3.

Hence why I'd prefer to leave this as is :-D

gaborbernat (commented):
I remain -1 on this.

ssbarnea:
@gaborbernat I was hoping to get some more constructive feedback from you, like providing a path out of this situation.

I want to fix a major design flaw but with tox which prevents it from creating machine parseable output. I did spend some effort working on this, please give me a way out this situation, one that would work for both v3 and v4.

I want to help tox and I do not have the expectation for you or other maintainer to fix it. I can do it myself, once I get the support.

I know you want to focus on bringing v4 to release but the reality is that even if you release it tomorrow it will likely take more than 6 months to address most of the broken plugins due to the api changes. This creates a serious gap and refusing fixes for v3 would only deter others from contributing to the project in general.

**What would be the kind of fix that you would support?**

gaborbernat:
> @gaborbernat I was hoping to get some more constructive feedback from you, like providing a path out of this situation.

I did. It's called call twice the command. The second run is guaranteed to be clean.

> I want to fix a major design flaw

Exactly the point. This is a major design part of tox 3, and considering it's coming EOL I'd rather not touch it. It's a niche use-case at the end of the day (we've been living with it for over 2 years now, so having to wait another 6 months is not that bad).

obestwalter:
I don't quite get, why there needs to be such a big discussion for a low risk fix in a sub command that is not even part of the main program flow, which effectively mitigates an agreed upon inconsistency in tox3. 

Independent of how close the release of tox4 is, I think @ssbarnea is very likely correct in his assessment that tox3 will still be around for a while. So what is keeping us from simply merging this and moving on?

gaborbernat:
I don't consider this low risk change.

obestwalter:
> I don't consider this low risk change. All flows are main flows, I'm using these flows often myself in many downstream places.

I am still struggling to understand, what the risk is to make the third (potentially) machine readable command consistent with the other two. What am I missing?

gaborbernat:
> e third (potentially) machine-readable command consistent with the other two

Can you clarify which commands are you referring to here?

obestwalter:
`--showconfig`

@ssbarnea wrote:

> Assures we use print() when displaying output using -a, -l or --showconfig, which are all commands that should not be affected by the currently configured logging level.

So this PR is about making it behave the same as -l and -a

obestwalter (approved):

src/tox/session/commands/show_config.py
```diff
@@ -29,7 +29,10 @@ def show_config(config):
...
+    print(value)
```
gaborbernat, line 35: > So this PR is about making it behave the same as -l and -a

This is not true. https://github.com/tox-dev/tox/blob/master/src/tox/session/commands/show_env.py#L22 is not using print.

gaborbernat (changes requested):

src/tox/session/commands/show_env.py
```diff
@@ -19,14 +19,17 @@ def report_env(e):
             msg = "{} -> {}".format(e.ljust(max_length), text).strip()
         else:
             msg = e
-        report.line(msg)
+        # Avoiding reporter because that is the result of program execution,
+        # and we do not want to be affected by logging level.
+        print(msg)
```
gaborbernat, line 24: Why does need to change from line to print?

src/tox/session/commands/show_config.py
obestwalter: What is not true? The behaviour will not be consistent, after the PR is merged? Ok then, how can we make it consistent? 

You asked for the opinion of other maintainers. I was out of the loop for a while and I am trying to help to solve a dispute without having any skin in the game myself. I reviewed the changes and gave my opinion and it is: I think it would be good if the 3 machine readable commands would be behave consistently and with my limited understanding that I have about the code base in its current state, the PR solves that problem. If it doesn't do it in the correct way according to your (undoubtably better) understanding, then you are the one who is in the best position to know how to do advise how to do it.

So, my question is: if this is **not** the correct way to make the behaviour of the three commands consistent (and I think we agree that this would be a GoodThing(TM)), then what **is** the correct way?

I also still don't see the danger in these changes and it would help if you could explain for the less experienced developers in the project, what the danger exactly is that you perceive.

gaborbernat: > What is not true?

Linked you the line showing that's not true. What's happening on master is not what's being changed to.

> , the PR solves that problem

It does so by changing the behaviour of all three commands, not by bringing show config in line with the other two. Also, the new commands side-steps our timestamp printing logic built-in into the reporter. These commands are only meant to be machine-readable when the verbosity is <0. The change here breaks that by trying to make it always machine-readable.

The danger is breaking workflows that actually use the current behaviour.

src/tox/session/commands/show_env.py
ssbarnea: If `default environments:` and `additional environments:` use verbosity0, it will be impossible to display environments with descriptions in parsable format.

If you want to list all environments with their descriptions, you must add `-v`, but that also enables those two extra lines above, which render the output as non-parsable. That is caused by another historical mistake of altering the output format based on the verbosity. 

If you want to avoid touching this area we may only fix `--showconfig` and avoid touching the other two. AFAIK, `--showconfig` is a clear case of output that is expected to be machine parseable. The other two, are less important as their main focus was human users. I can live without being able to extra description from `tox -l` or `tox -a` as I am already o parsing --showconfig which already contains it.

Ok to touch only showconfig?

gaborbernat: My question was more along the lines why are you using print over `report.line` (wasn't talking about verbosity0).

gaborbernat: > If you want to list all environments with their descriptions, you must add `-v`, but that also enables those two extra lines above, which render the output as non-parsable. That is caused by another historical mistake of altering the output format based on the verbosity.

I don't think this is a mistake. To get all environments you should use `tox -a`. To get just default environments you should use `tox -l`. To get environments with the description you should use `tox --showconfig`.

> AFAIK, `--showconfig` is a clear case of output that is expected to be machine parseable.

Not true. It's actually primarily human-readable, so users can see how their config would be expanded. Secondary to be machine-readable, many of the more complex settings don't have a good machine representation (e.g. enums, etc). Primarly all of these are aimed at human users. Without verbosity `tox -l` and `tox -v` are machine parse-able. tox --showconfig (without verbosity) is intended to be mostly machine parse-able.

gaborbernat:
I'm closing this now as the PR became stale, we can reopen if we pick it up again. Thanks!

Issue #1484, opened 2019-12-14
tox 3.14.2 uses wrong Python interpreters


hynek:
When trying to run a tox env using Python 3.8, it actually uses 3.7 (concretely the system interpreter).

I only noticed it because I used hashes in my lockfiles and my builds started failing due to a lack of a hash to dependency that is need on 3.7 but not on 3.8.

Downgrading to homebrew's 3.14.1 fixes it.

``tox -rvv``:

<details>

```
using tox.ini: /Users/hynek/Work/rdap-svc/tox.ini (pid 49542)
  removing /Users/hynek/Work/rdap-svc/.tox/log
using tox-3.14.2 from /Users/hynek/.local/pipx/venvs/tox/lib/python3.7/site-packages/tox/__init__.py (pid 49542)
GLOB start: packaging
GLOB sdist-make: /Users/hynek/Work/rdap-svc/setup.py
  removing /Users/hynek/Work/rdap-svc/.tox/dist
[49543] /Users/hynek/Work/rdap-svc$ /Users/hynek/.local/pipx/venvs/tox/bin/python setup.py sdist --formats=zip --dist-dir .tox/dist >.tox/log/GLOB-0.log
running sdist
running egg_info
writing src/rdap_svc.egg-info/PKG-INFO
writing dependency_links to src/rdap_svc.egg-info/dependency_links.txt
writing top-level names to src/rdap_svc.egg-info/top_level.txt
reading manifest file 'src/rdap_svc.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no previously-included files found matching '*.txt'
warning: no previously-included files found matching '*.cfg'
warning: no previously-included files found matching '*.rst'
warning: no previously-included files found matching '*.ini'
warning: no previously-included files found matching '*.yml'
warning: no previously-included files found matching '*.yaml'
warning: no previously-included files found matching '.coveragerc'
warning: no previously-included files found matching 'Makefile'
warning: no previously-included files found matching 'conftest.py'
warning: no previously-included files found matching '.envrc'
warning: no previously-included files found matching 'Dockerfile'
warning: no previously-included files found matching 'test-queries.http'
warning: no previously-included files found matching '.dockerignore'
warning: no previously-included files found matching 'Pipfile*'
warning: no previously-included files found matching '*.toml'
warning: no previously-included files found matching '*.paw'
warning: no previously-included files found matching '*.pot'
no previously-included directories found matching 'tests'
[...]

Issue #882, opened 2018-07-08
Tox envlist selects incorrect python versions


labrys:
# Minimal reproducible example and detailed description
If a Python version is not available for a defined environment in the `envlist` tox selects the default Python for some environments and fails with `InterpreterNotFound` for others.  This should be consistent and fail for all versions of python that aren't available.

### Prerequisites:
- Must have at least one version of Python installed (Python 3.7.0 in this example)
- Must __NOT__ have at least one of the *affected* Python environments installed
- Must __NOT__ have at least one of the *unaffected* Python environments installed
- Must have `tox` installed in the current environment

#### Create the following `tox.ini` in the current directory:
```ini
[tox]
envlist =
;    Affected Python Environments (not comprehensive)
;    The following will not fail if they are unavailable
    py25
    py33
    py40
;    Unaffected Python environments (not comprehensive)
;    The expected "InterpreterNotFound" error will be displayed if they are unavailable
    py34
    py35

skipsdist = true

[testenv]
commands =
    python --version
```
#### Check current Python version
In this example Python 3.7.0 is the system default, your version may differ
```
>python --version
Python 3.7.0
```
#### Run test and review output
Only one success and one failure shown for brevity:
```shell
>python -m tox
python -m tox
py25 create: E:\src\toxin\.tox\py25
py25 runtests: PYTHONHASHSEED='50'
py25 runtests: commands[0] | python --version
Python 3.7.0
...
py34 create: E:\src\toxin\.tox\py34
ERROR: InterpreterNotFound: python3.4
```
Python 2.5 is not installed but an environment for Python 2.5 was created (`py25 create: E:\src\toxin\.tox\py25`) but is using the current system default Python 3.7.0

Python 3.4 is not installed and the expected error is displayed (`ERROR: InterpreterNotFound: python3.4`)

# OS and `pip list` output
```shell
>python -c "import platform; print(platform.platform())"
Windows-10-10.0.17134-SP0

>py -0
Installed Pythons found by py Launcher for Windows
 -3.7-32 *
 -3.6-32
 -2.7-32

>pip list
packaging (17.1)
pip (9.0.1)
pluggy (0.6.0)
py (1.5.4)
pyparsing (2.2.0)
setuptools (28.8.0)
six (1.11.0)
tox (3.1.0)
virtualenv (16.0.0)
```

obestwalter:
Hi @labrys, thanks for the detailed report. I can reproduce this with 3.0 on Linux.

[...]

PR #2047, opened 2021-04-26
Fail if Specified Interpreter Not Found

jackdesert:
Fail if specified interpreter not found.
(Do NOT use basepython as a fallback)

This is intended to remedy these two github issues:

  https://github.com/tox-dev/tox/issues/882
  https://github.com/tox-dev/tox/issues/1484

In particular, when python3.1 is NOT installed on my machine, using this tox.ini:

```
[tox]
envlist = py3.1

[testenv]
deps = pytest
commands = pytest
```

this pull request makes is so that tox fails.

(Without this commit, tox runs pytest using the same python interpreter that was used to invoke tox.
In my opinion it is more helpful to fail than to run tests with a different-than-intended interpreter.)

If this pull request seems generally useful, let me know what it would take to get it accepted. (It broke several tests, for example.)

### DIFF
--- a/src/tox/interpreters/unix.py
+++ b/src/tox/interpreters/unix.py
@@ -1,5 +1,7 @@
 from __future__ import unicode_literals
+import sys
 
+from tox import reporter
 import tox
 
 from .common import base_discover
@@ -8,12 +10,15 @@ from .via_path import check_with_path
 
 @tox.hookimpl
 def tox_get_python_executable(envconfig):
+    """
+    Return path to specified interpreter.
+    If not available, exit(1).
+    """
     spec, path = base_discover(envconfig)
-    if path is not None:
+
+    if path is None:
+        reporter.error('envname {} not found'.format(envconfig.envname))
+        sys.exit(1)
+    else:
         return path
-    # 3. check if the literal base python
-    candidates = [envconfig.basepython]
-    # 4. check if the un-versioned name is good
-    if spec.name is not None and spec.name != envconfig.basepython:
-        candidates.append(spec.name)
-    return check_with_path(candidates, spec)
+~
--- a/src/tox/interpreters/unix.py
+++ b/src/tox/interpreters/unix.py
@@ -1,8 +1,9 @@
 from __future__ import unicode_literals
+
 import sys
 
-from tox import reporter
 import tox
+from tox import reporter
 
 from .common import base_discover
 from .via_path import check_with_path
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -46,6 +46,7 @@ Igor Duarte Cardoso
 Ilya Kulakov
 Ionel Maries Cristian
 Itxaka Serrano
+Jack Desert
 Jake Windle
 Jannis Leidel
 Jesse Schwartzentruber


gaborbernat (changes requested):
How is this taking into account https://tox.readthedocs.io/en/latest/config.html#conf-skip_missing_interpreters?

src/tox/interpreters/unix.py
```diff
@@ -8,12 +10,15 @@
...
 def tox_get_python_executable(envconfig):
+    """
+    Return path to specified interpreter.
+    If not available, exit(1).
+    """
     spec, path = base_discover(envconfig)
-    if path is not None:
+
+    if path is None:
+        print('envname {} not found'.format(envconfig.envname))
```
gaborbernat, line 20: Please don't use print, the reporter module exists instead.

src/tox/interpreters/unix.py
```diff
@@ -8,12 +10,15 @@
...
+        sys.exit(1)
```
gaborbernat, line 21: You're not supposed to use sys.exit here, instead raise an exception.

gaborbernat (changes requested):

src/tox/interpreters/unix.py
jackdesert: Can you say which exception could be raised to give a concise error message back to the user?

I have tried raising tox.exception.UnsupportedInterpreter but that presented a lot of error text to the user.

gaborbernat: Would need to investigate, but that does seem the right exception type. Also, I don't think this should be handled here, but higher up in the call chain.

asottile:
@jackdesert I think this is a misunderstanding `py3.1` is not the proper env name for python 3.1 -- that would be `py31`, which fails:

```console
$ tox
py31 create: /tmp/y/.tox/py31
ERROR: InterpreterNotFound: python3.1
ERROR:  py31: InterpreterNotFound: python3.1
```

src/tox/interpreters/unix.py
jackdesert: Changed to reporter.error()

jackdesert:
> @jackdesert I think this is a misunderstanding `py3.1` is not the proper env name for python 3.1 -- that would be `py31`, which fails:
> 
> ```
> $ tox
> py31 create: /tmp/y/.tox/py31
> ERROR: InterpreterNotFound: python3.1
> ___________________________________ summary ____________________________________
> ERROR:  py31: InterpreterNotFound: python3.1
> ```

Would it be straightforward to create a pull request that raises an error if an improper env name is used?

asottile:
> 
> 
> > @jackdesert I think this is a misunderstanding `py3.1` is not the proper env name for python 3.1 -- that would be `py31`, which fails:
> > ```
> > $ tox
> > py31 create: /tmp/y/.tox/py31
> > ERROR: InterpreterNotFound: python3.1
> > ___________________________________ summary ____________________________________
> > ERROR:  py31: InterpreterNotFound: python3.1
> > ```
> 
> Would it be straightforward to create a pull request that raises an error if an improper env name is used?

no, envnames can be literally anything -- only a handful have special meaning (`py`, `pyX`, `pyXX`, `pypy`, `pypyX`, `pypyXX`)

jackdesert:
> no, envnames can be literally anything -- only a handful have special meaning (`py`, `pyX`, `pyXX`, `pypy`, `pypyX`, `pypyXX`)

How can this be paved to be clear to future users? It was frustrating to be attempting to specify python 3.1 and having it choose the system python instead.

asottile:
> 
> 
> > no, envnames can be literally anything -- only a handful have special meaning (`py`, `pyX`, `pyXX`, `pypy`, `pypyX`, `pypyXX`)
> 
> How can this be paved to be clear to future users? It was frustrating to be attempting to specify python 3.1 and having it choose the system python instead.

I don't know how this could be more clear!  no part of our documentation nor the quick start nor any of the thousands of examples in open source use names like what you've used -- I guess the better question is how did you arrive to your naming?

jackdesert:
> I don't know how this could be more clear! no part of our documentation nor the quick start nor any of the thousands of examples in open source use names like what you've used -- I guess the better question is how did you arrive to your naming?

Not sure. I am using pyenv, which uses periods.

Issue #2065, opened 2021-05-20
--devenv fails because root .tox environment is provisioned in identical envdir


isaac-ped:
Attempting to create a development environment with `--devenv` causes  fatal errors if the root tox environment needs to be provisioned

This minimal setup reproduces the issue (as long as that version of setuptools is not already installed on the host)
```ini
[tox]
requires =
    setuptools==56.2.0
```
with a very minimal setup.py:
```python
from setuptools import setup   
                               
setup(                         
    name="noop",               
    packages=[]                
)
```

Running:
`tox -rvv --devenv venv`

Tox provisions its root environemnt in `venv`:
```
.tox start: getenv /sandbox/tox-debug/venv
.tox cannot reuse: -r flag
.tox create: /sandbox/tox-debug/venv
```
and then later tries to remove it to create the devenv:
```
py recreate: /sandbox/tox-debug/venv
  removing /sandbox/tox-debug/venv
```

Seems related to https://github.com/tox-dev/tox/issues/1501, though that issue seems to be less general 

```shell
$ tox --version
3.23.1 imported from /u1/ipedisich/.local/lib/python3.6/site-packages/tox/__init__.py
```
(also reproduced with python3.7)
[Full output of `tox -rvvv`](https://github.com/tox-dev/tox/files/6517600/tox-error.txt)

gaborbernat:
Released via https://pypi.org/project/tox/3.24.0/

PR #2066, opened 2021-05-20
Do not modify provisioned envdir when --devenv is specified

isaac-ped:
This change prevents `--devenv` from changing the envdir for the `.tox` environment

Fixes tox-dev/tox#2065

- [x] added/updated test(s)
      in message body

### DIFF
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -595,7 +595,10 @@ def tox_addoption(parser):
     )
 
     def _set_envdir_from_devenv(testenv_config, value):
-        if testenv_config.config.option.devenv is not None and testenv_config.envname != ".tox":
+        if (
+            testenv_config.config.option.devenv is not None
+            and testenv_config.envname != testenv_config.config.provision_tox_env
+        ):
             return py.path.local(testenv_config.config.option.devenv)
         else:
             return value
--- a/src/tox/config/__init__.py
+++ b/src/tox/config/__init__.py
@@ -595,7 +595,7 @@ def tox_addoption(parser):
     )
 
     def _set_envdir_from_devenv(testenv_config, value):
-        if testenv_config.config.option.devenv is not None:
+        if testenv_config.config.option.devenv is not None and testenv_config.envname != ".tox":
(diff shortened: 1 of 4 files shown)


jugmac00:
I had no chance to look closer, but anyway, could you please add a changelog entry? Thank you.

src/tox/config/__init__.py
```diff
@@ -595,7 +595,7 @@ def tox_addoption(parser):
     )
 
     def _set_envdir_from_devenv(testenv_config, value):
-        if testenv_config.config.option.devenv is not None:
+        if testenv_config.config.option.devenv is not None and testenv_config.envname != ".tox":
```
gaborbernat, line 598: This cannot be a static value, you need to read it from https://tox.readthedocs.io/en/latest/config.html#conf-provision_tox_env

gaborbernat (changes requested):

src/tox/config/__init__.py
isaac-ped: Thank you for the pointer! That has been fixed

isaac-ped:
> I had no chance to look closer, but anyway, could you please add a changelog entry? Thank you.

I was unsure if it was intensional for my changelog file to be the only one in that directory, and didn't realize that was the cause of the `Timeline protection` failure. Given that it fixed the failing CI, I'm guessing that was indeed the correct location for it.

jugmac00:
> > I had no chance to look closer, but anyway, could you please add a changelog entry? Thank you.
> 
> I was unsure if it was intensional for my changelog file to be the only one in that directory, and didn't realize that was the cause of the `Timeline protection` failure. Given that it fixed the failing CI, I'm guessing that was indeed the correct location for it.

Yes, you are absolutely right. We use https://pypi.org/project/towncrier/ for changelogs. One of the great advantages is, that there are no merge conflicts because of the change log, as the changelog is only built when there is a new release. Compare this to the situtation, when e.g. 10 pull requests are open - as soon as the first one gets merged, all the others would have to rebase over and over again.

gaborbernat (approved):

PR #2086, opened 2021-06-11
remove wrong example from documentation

jugmac00:
It does not work to only provide constraints, as they only apply to
specified dependencies.


jugmac00:
Thanks to @altendky for pointing out at https://discord.com/channels/802911963368783933/802911964053504032/852176487124697088

gaborbernat (approved):

altendky:
Thanks for turning my random comment into something actually useful.  `:]`

Issue #2089, opened 2021-06-17
changedir breaks script discovery


felix-hilden:
Hi, firstly thanks for providing Tox. It's awesome! I also saw the PyCon talk about Tox 4 and I'm really looking forward to it 😄

I've created an env for building Sphinx documentation with its make scripts. I know the documentation suggests using `sphinx-build` directly but anyways: the script cannot be found under the directory.

> InvocationError for command could not find executable my_script

<details>
<summary>

Output of `tox -e script -rvv`

</summary>

```sh
(venv) C:\Felix\Git\ztst>tox -e script -rvv
using tox.ini: C:\Felix\Git\ztst\tox.ini (pid 14348)
  removing C:\Felix\Git\ztst\.tox\log
using tox-3.23.1 from c:\felix\git\ztst\venv\lib\site-packages\tox\__init__.py (pid 14348)
GLOB start: packaging 
GLOB sdist-make: C:\Felix\Git\ztst\setup.py
  removing C:\Felix\Git\ztst\.tox\dist
[6856] C:\Felix\Git\ztst$ 'c:\felix\git\ztst\venv\scripts\python.exe' setup.py sdist --formats=zip --dist-dir '.tox\dist' >.tox\log
\GLOB-0.log
running sdist
running egg_info
writing package.egg-info\PKG-INFO
writing dependency_links to package.egg-info\dependency_links.txt
writing top-level names to package.egg-info\top_level.txt
reading manifest file 'package.egg-info\SOURCES.txt'
writing manifest file 'package.egg-info\SOURCES.txt'
warning: sdist: standard file not found: should have one of README, README.rst, README.txt, README.md

running check
warning: check: missing required meta-data: url

creating package-0.0.1
creating package-0.0.1\package
creating package-0.0.1\package.egg-info
copying files to package-0.0.1...
copying setup.py -> package-0.0.1
copying package\__init__.py -> package-0.0.1\package
copying package.egg-info\PKG-INFO -> package-0.0.1\package.egg-info
copying package.egg-info\SOURCES.txt -> package-0.0.1\package.egg-info
copying package.egg-info\dependency_links.txt -> package-0.0.1\package.egg-info
copying package.egg-info\top_level.txt -> package-0.0.1\package.egg-info
Writing package-0.0.1\setup.cfg
creating '.tox\dist\package-0.0.1.zip' and adding 'package-0.0.1' to it
adding 'package-0.0.1'
adding 'package-0.0.1\package'
[...]

PR #2096, opened 2021-06-22
Document script discovery wrt changedir (#2089)

gaborbernat:
👍


docs/config.rst
```diff
@@ -360,7 +367,7 @@ Complete list of settings that you can put into ``testenv*`` sections:
 
 .. conf:: changedir ^ PATH ^ {toxinidir}
 
-    Change to this working directory when executing the test command.
+    Change to the current working directory when executing the test command.
```
felix-hilden, line 370:
```suggestion
    Change the current working directory when executing the test command.
```

The original change could be confused for "this command changes the workind directory be the current working directory".

felix-hilden (approved):
I agree, it's better in this concise form. Thanks for the reviews!

PR #2103, opened 2021-07-13
Fix show config when package names are not canonical

gaborbernat:
After landing https://github.com/pypa/virtualenv/pull/2126 I'm getting:

```bash
[2021-07-13T12:54:28.017Z]     current_dist = importlib_metadata.distribution(current)
[2021-07-13T12:54:28.017Z]   File "/opt/bb/lib/python3.9/importlib/metadata.py", line 524, in distribution
[2021-07-13T12:54:28.017Z]     return Distribution.from_name(distribution_name)
[2021-07-13T12:54:28.017Z]   File "/opt/bb/lib/python3.9/importlib/metadata.py", line 187, in from_name
[2021-07-13T12:54:28.017Z]     raise PackageNotFoundError(name)
[2021-07-13T12:54:28.017Z] importlib.metadata.PackageNotFoundError: backports-entry-points-selectable
```

### DIFF
--- a/src/tox/session/commands/show_config.py
+++ b/src/tox/session/commands/show_config.py
@@ -2,7 +2,6 @@ import sys
 from collections import OrderedDict
 
 from packaging.requirements import Requirement
-from packaging.utils import canonicalize_name
 from six import StringIO
 from six.moves import configparser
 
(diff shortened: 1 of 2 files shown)


src/tox/session/commands/show_config.py
```diff
@@ -65,16 +64,15 @@ def version_info(parser):
     while to_visit:
         current = to_visit.pop()
         current_dist = importlib_metadata.distribution(current)
-        current_name = canonicalize_name(current_dist.metadata["name"])
```
asottile, line 68: hmmm, I was pretty sure I added the canonicalize name because there were duplicates in the output

gaborbernat: Yeah, not sure what happened here but canonicalize name now makes it break... do you still have your use case to validate?

asottile:
I don't unfortunately :S -- I think it was within the tree of tox
dependencies but if that doesn't reproduce anymore than it's probably fine

PR #2106, opened 2021-07-15
Fix misformed title in changelog

AntoineD:
The title of the last release in the changelog is not well rendered because it is not underlined .


jugmac00 (approved):
Thank you!

PR #2114, opened 2021-07-24
`tox -e ALL` now runs all environments as expected

### LINKED ISSUE #2112: tox4: specify behavior for `tox -e ALL`
In tox3 this runs really all environments, in tox4 this runs all envs in the envlist.

@asottile @gaborbernat @obestwalter What's your take on this?

I am +1 on running really all environments with `tox -e ALL`.

Related to #2045

Interestingly, `tox -e ALL` also behaves differently between tox3 and tox4.

```
❯ tox -e ALL
hello run-test-pre: PYTHONHASHSEED='1741336755'
hello run-test: commands[0] | python3 -c 'print("Hello")'
Hello
goodbye run-test-pre: PYTHONHASHSEED='1741336755'
goodbye run-test: commands[0] | python3 -c 'print("Goodbye")'
Goodbye
_______________________________________________________ summary ________________________________________________________
  hello: commands succeeded
  goodbye: commands succeeded
  congratulations :)


❯ tox4dev -e ALL
hello: commands[0]> python3 -c 'print("Hello")'
Hello
  hello: OK (0.13=setup[0.10]+cmd[0.03] seconds)
  congratulations :) (0.16 seconds)
```

I suppose `ALL` should really run all envs, not only those in `envlist`? @gaborbernat

_Originally posted by @jugmac00 in https://github.com/tox-dev/tox/issues/2045#issuecomment-826657324_

jugmac00:
This fixes #2112.

- [x] added/updated test(s)
      in message body

### DIFF
--- a/tests/session/cmd/test_legacy.py
+++ b/tests/session/cmd/test_legacy.py
@@ -89,3 +89,29 @@ def test_legacy_run_sequential(tox_project: ToxProjectCreator, mocker: MockerFix
     tox_project({"tox.ini": ""}).run("le", "-e", "py")
 
     assert run_sequential.call_count == 1
+
+
+def test_ensure_all_environments_are_executed(tox_project: ToxProjectCreator) -> None:
+    ini = """
+    [tox]
+    env_list = aaa
+    [testenv]
+    [testenv:bbb]
+    description = not in env_list
+    """
+    setup_py = """
+    from setuptools import setup
+    setup(
+        name='a',
+        version='1.0',
+        install_requires=['setuptools>44'],
+    )
+    """
+    outcome = tox_project(
+        {
+            "tox.ini": ini,
+            "setup.py": setup_py,
+        }
+    ).run("le", "-e", "ALL")
+    assert "aaa: OK" in outcome.out
+    assert "bbb: OK" in outcome.out
--- a/tests/session/cmd/test_legacy.py
+++ b/tests/session/cmd/test_legacy.py
@@ -89,29 +89,3 @@ def test_legacy_run_sequential(tox_project: ToxProjectCreator, mocker: MockerFix
     tox_project({"tox.ini": ""}).run("le", "-e", "py")
 
     assert run_sequential.call_count == 1
-
-
-def test_ensure_all_environments_are_executed(tox_project: ToxProjectCreator) -> None:
-    ini = """
-    [tox]
-    env_list = aaa
-    [testenv]
-    [testenv:bbb]
-    description = not in env_list
-    """
-    setup_py = """
-    from setuptools import setup
-    setup(
-        name='a',
-        version='1.0',
-        install_requires=['setuptools>44'],
-    )
-    """
-    outcome = tox_project(
-        {
-            "tox.ini": ini,
-            "setup.py": setup_py,
-        }
-    ).run("le", "-e", "ALL")
-    assert "aaa: OK" in outcome.out
-    assert "bbb: OK" in outcome.out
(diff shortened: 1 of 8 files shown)


jugmac00:
Hm, just noticed the flaw in the PR - will have another look tomorrow.

jugmac00:
So, the problem here is that `CliEnv.all` only reflects two states, but there are three:

https://github.com/tox-dev/tox/blob/ae4fd379ebecc9a07c719a5951c6eb12c092a834/src/tox/session/common.py#L7-L12

There are the following states:
- no value -> choose complete `envlist` (e.g. `tox`)
- "ALL" value -> choose all defined envs (also include those not in `envlist`, `tox -e ALL`)
- some value -> choose specified envs (e.g. `tox -e py39`)

I am thinking of creating an enum like ...

```python
class ChosenEnvs(Enum):
    SPECIFIED_ENVS = 1
    ENV_LIST = 2
    ALL_ENVS = 3


class CliEnv:
    def __init__(self, value: Union[None, List[str], str] = None):
        if isinstance(value, str):
            value = StrConvert().to(value, of_type=List[str], kwargs={})
        if value is None:
            self.chosen_envs = ChosenEnvs.ENV_LIST
        elif "ALL" in value:
            self.chosen_envs = ChosenEnvs.ALL_ENVS
        else:
            self.chosen_envs = ChosenEnvs.SPECIFIED_ENVS
        self._names = value
```

This would lead to quite some changes, as `self.all` was used a lot in the `CliEnv` class itself, and also at least one time outside 
https://github.com/tox-dev/tox/blob/1f565b17f446ba349604681bef2a3635828250a6/src/tox/session/cmd/show_config.py#L64-L68

@gaborbernat Any opinions on this?

tests/session/cmd/test_legacy.py
```diff
@@ -89,3 +89,29 @@ def test_legacy_run_sequential(tox_project: ToxProjectCreator, mocker: MockerFix
...
+
+def test_ensure_all_environments_are_executed(tox_project: ToxProjectCreator) -> None:
+    ini = """
+    [tox]
+    env_list = aaa
+    [testenv]
+    [testenv:bbb]
+    description = not in env_list
+    """
+    setup_py = """
```
gaborbernat, line 102: You don't neeed the package part, the problem is independent weather there's a packaging env or not. And when that's a true is better to not have it to make the test faster.

tests/session/cmd/test_legacy.py
```diff
@@ -89,3 +89,29 @@ def test_legacy_run_sequential(tox_project: ToxProjectCreator, mocker: MockerFix
...
+        version='1.0',
+        install_requires=['setuptools>44'],
+    )
+    """
+    outcome = tox_project(
+        {
+            "tox.ini": ini,
+            "setup.py": setup_py,
+        }
+    ).run("le", "-e", "ALL")
```
gaborbernat, line 115: This test should be better served in the non legacy scope as it's present in the modern CLI too. Generally in the legacy suite we only test that the legacy shim transforms correctly the config to our normal case.

gaborbernat (commented):
I don't think you need the enum, but otherwise you were on the right path. In trying to understand why this is like this I've actually done most of the lifting left.

This exposed that our config showing sub-command was incosistent with how we expected the -e flag to work in other sub-commands. Fixing that required quite a signficiant rework of show_config, see https://github.com/gaborbernat/tox/commit/375151495d80445eb4fdc64284efc858ad39275f

Will put a commit later today on top of your PR with it.

tests/session/cmd/test_legacy.py
jugmac00: Nice catch and thanks for the explanation!

gaborbernat (approved):

Issue #1127, opened 2019-01-08
tox4: turn non whitelisted warning into error


obestwalter:
At the moment we show this warning if we call to a tool from the parent env:

```
WARNING: test command found but not installed in testenv
  cmd: /home/ob/.virtualenvs/play/bin/pytest
  env: /home/ob/do/play/.tox/python
Maybe you forgot to specify a dependency? See also the whitelist_externals envconfig setting.
```

I am working on documentation improvements for #461 and my current feeling is that we should start throwing an error if somebody does not whitelist an external command from tox 4 on. From 3.7 on we should show an enhanced warning that will announced that the warning will be an error from tox 4 on - like:

```
WARNING: test command found but not installed in testenv
  cmd: /home/ob/.virtualenvs/play/bin/pytest
  env: /home/ob/do/play/.tox/python
Maybe you forgot to specify a dependency? See also the whitelist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above
```

gaborbernat:
Sure 👍

obestwalter:
One thing to keep in mind is that this is true for all commands that are not known inside the env (e.g. system `echo` and `ls`) so we should definitely have a proper deprecation period but I think tox 4 is still a while away :)

gaborbernat:
Yeah, I'm fine with starting to emit warnings and remove with tox 4 (maybe next year at earliest) if we have enough breaking changes.

tucked:
With 3.7.0, I get a warning for `commands = bash` but not for `commands = /bin/bash`. Is this expected?
(FWIW, I'm actually using `commands =  dev: {posargs:python}` and `tox -e py35-dev -- bash` vs. `tox -e py35-dev -- "$SHELL"`.)

hexagonrecursion:
Note that tox4 currently does not even issue a warning in this case. See https://github.com/tox-dev/tox/issues/1946

jugmac00:
In the current tox4 rewrite documenation, you can read:
> In general, all executables in the path are available in ``commands``, but
tox will emit a warning if it was not explicitly allowed via ``whitelist_externals``.

So we need to update the `user_guide.rst` once this issue gets resolved.

gaborbernat:
I haven't started yet to dive deep into the documentation, so likely a lot of things will be out of date for now.

jugmac00:
> I haven't started yet to dive deep into the documentation, so likely a lot of things will be out of date for now.

[...]

PR #2134, opened 2021-08-02
Add support for allowlist_externals

gaborbernat:
Reolves https://github.com/tox-dev/tox/issues/1127

### DIFF
--- a/src/tox/execute/local_sub_process/__init__.py
+++ b/src/tox/execute/local_sub_process/__init__.py
@@ -192,7 +192,7 @@ class LocalSubProcessExecuteInstance(ExecuteInstance):
                             break
                     else:
                         msg = f"{base} (resolves to {executable})" if base == executable else base
-                        raise Fail(f"{msg} is not allowed, use allowlist_external to allow it")
+                        raise Fail(f"{msg} is not allowed, use allowlist_externals to allow it")
                 # else use expanded format
                 cmd = [executable, *self.request.cmd[1:]]
             self._cmd = cmd
--- a/src/tox/execute/local_sub_process/__init__.py
+++ b/src/tox/execute/local_sub_process/__init__.py
@@ -1,4 +1,5 @@
 """Execute that runs on local file system via subprocess-es"""
+import fnmatch
 import logging
 import os
 import shutil
@@ -7,6 +8,8 @@ import time
 from subprocess import DEVNULL, PIPE, TimeoutExpired
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Sequence, Tuple, Type
+
+from tox.tox_env.errors import Fail
 
 from ..api import Execute, ExecuteInstance, ExecuteStatus
 from ..request import ExecuteRequest, StdinSource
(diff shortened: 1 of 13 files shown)


jugmac00 (commented):
I'll have a closer look this afternoon!

`whitelist_externals` was announced to be removed with the next major release - did you keep it intentionally?
https://tox.readthedocs.io/en/latest/changelog.html?highlight=whitelist#v3-18-0-2020-07-23

On the other hand - we do not emit a deprecation warning with `tox3.24` so removing it in `tox4` would be pretty surprising to many, as not everybody reads the docs.

gaborbernat:
> On the other hand - we do not emit a deprecation warning with `tox3.24` so removing it in `tox4` would be pretty surprising to many, as not everybody reads the docs.

Yeah, pretty much this. I guess we could extend it to emmit a warning now... But tox 4 will already be breaking enough that did not want to add this extra on top.

src/tox/execute/local_sub_process/__init__.py
```diff
@@ -176,10 +179,20 @@ def __init__(
...
             else:
+                if self.request.allow is not None:
+                    for allow in self.request.allow:
+                        # 1. allow matches just the original name of the executable
+                        # 2. allow matches the entire resolved path
+                        if fnmatch.fnmatch(self.request.cmd[0], allow) or fnmatch.fnmatch(executable, allow):
+                            break
+                    else:
+                        msg = f"{base} (resolves to {executable})" if base == executable else base
+                        raise Fail(f"{msg} is not allowed, use allowlist_external to allow it")
```
jugmac00, line 195:
```suggestion
                        raise Fail(f"{msg} is not allowed, use allowlist_externals to allow it")
```

src/tox/execute/request.py
```diff
@@ -38,6 +44,9 @@ def __init__(
         self.env = env  #: the environment variables to use
         self.stdin = stdin  #: the type of standard input interaction allowed
         self.run_id = run_id  #: an id to identify this run
+        if allow is not None and "*" in allow:
+            allow = None  # if we allow everything we can just disable the check
+        self.allow = allow
```
jugmac00, line 49: So, if I get it right, `allow` is usually an empty list, or when you use e.g. make, it would be `["make"]`.

In which case would `allow` directly be `None`?

tests/execute/local_subprocess/test_local_subprocess.py
```diff
@@ -275,3 +276,23 @@ def test_local_subprocess_tty(monkeypatch: MonkeyPatch, mocker: MockerFixture, t
...
         "terminal": [100, 100],
     }
+
+
+@pytest.mark.parametrize("mode", ["stem", "full", "stem-pattern", "full-pattern", "all"])
+def test_allow_list_external_ok(fake_exe_on_path: Path, mode: str) -> None:
+    exe = f"{fake_exe_on_path}{'.EXE' if sys.platform == 'win32' else ''}"
+    allow = exe if "full" in mode else fake_exe_on_path.stem
+    allow = f"{allow[:-2]}*" if "pattern" in mode else allow
+    allow = "*" if mode == "all" else allow
```
jugmac00, line 286: I would prefer to use `allow` as parameter for this test, but you possible had something in mind :-)

jugmac00 (approved):
Some mind-blowing stuff going on, but looks good with a couple of inline comments.

I confirmed it works as intended also manually.

I noticed a minor difference between tox3 and tox4 - I used the `ls` command to test the behavior and while tox3 puts the output of `ls` on one line, tox4 puts each file/directory on a separate line, ie there is some end of line splitting going on - this is probably not related to this PR.

src/tox/execute/request.py
gaborbernat: An empty allow list would mean nothing is allowed. Which I'd imagine is always a bad value. So that should never happen.

An allow containing ``*`` means allow everything, for which we use ``allow=None``. 

When invoking process calls where we don't want to check allow list (e.g. for the PEP-517 packaging calls) the caller does not set allow, and such remains `None` meaning don't do allow validation.

When executing commands from within tox run we inject the environments path arguments into the allow list, plus whatever the user sets. See from the file below:

```
    def _allow_externals(self) -> List[str]:
        result: List[str] = [f"{i}{os.sep}*" for i in self._paths]
        result.extend(i.strip() for i in self.conf["allowlist_externals"])
        return result
```

This means that only within tox commands, commands_pre and commands_post the allow will be:

```
.tox\py39\bin\*  # set by the framework
ls               # set by the user
```

while in other scopes remains disabled. If the user passes ``*`` then we disable allow validation.

tests/execute/local_subprocess/test_local_subprocess.py
gaborbernat: You don't know the value of fake_exe_on_path upfront to be able to do that 😊

Issue #2125, opened 2021-07-28
tox4: {:} support


gaborbernat:
https://tox.readthedocs.io/en/latest/config.html#globally-available-substitutions

PR #2138, opened 2021-08-04
Add support for os path separator substitution in ini files

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/2125


jugmac00 (approved):
LGTM

I'd strongly suggest that we add documentation to all new/old features - as far as I see the substitutes for path separators are not yet documented, neither `{:}` nor `{/}`.

gaborbernat:
Yeah, we have a ticket for it. Though this is not new feature per se 👍

PR #2139, opened 2021-08-05
Add envlogdir support

### LINKED ISSUE #2123: tox4: envlogdir support
https://tox.readthedocs.io/en/latest/config.html#conf-envlogdir

gaborbernat:
Resolves #2123.

### DIFF
--- a/docs/changelog/2123.feature.rst
+++ b/docs/changelog/2123.feature.rst
@@ -1,0 +1,2 @@
+Add support for :ref:`env_log_dir` (compared to tox 3 extend content and keep only last run entries) -
+by :user:`gaborbernat`.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -222,6 +222,15 @@ Base
    A folder that is always reset at the start of the run.
 
 .. conf::
+   :keys: env_log_dir, envlogdir
+   :default: {work_dir}/{env_name}/log
+
+   A folder containing log files about tox runs. It's always reset at the start of the run. Currently contains every
+   process invocation in the format of ``<index>-<run name>.log``, and details the execution request (command,
+   environment variables, current working directory, etc.) and its outcome (exit code and standard output/error
+   content).
+
+.. conf::
    :keys: platform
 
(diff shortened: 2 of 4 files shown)


jugmac00 (approved):
I just scanned the code, but tried both `env_log_dir` and `envlogdir` and it works.

I noticed a `strange` behavior when I cd'ed into the configured `env_log` directory and re-ran `tox` - the result was that the directory was emptied, but no new logs were created.

Unfortunately I have zero time before the weekend - not sure whether this is something which needs to be looked into or not.

gaborbernat:
Did you have any commands configured? It only generates logs if it runs some processes 😊

jugmac00:
> Did you have any commands configured? It only generates logs if it runs some processes

Yes, I used my https://github.com/jugmac00/flask-reuploaded to test your changes.

The behavior is consistent:
- when I cd into the configured directory, no new files get created (see below)
- when I am outside the directory, the log files get written correctly


A funky...

```
/tmp 
❯ mkdir tests

/tmp 
❯ cd tests/

/tmp/tests 
❯ ls -la
total 6612
drwxrwxr-x  2 jugmac00 jugmac00    4096 Aug  5 13:58 .
drwxrwxrwt 36 root     root     6758400 Aug  5 13:58 ..

/tmp/tests 

# I now delete /tmp/tests from a second terminal
# back in terminal 1 and inside tests

❯ ls -la
total 0

/tmp/tests 
❯ ls /tmp/t*
ls: cannot access '/tmp/t*': No such file or directory
```

So, the `env_dir` gets deleted, but it cannot get created again, as maybe the terminal still holds a handle to the deleted one .. or so :)

jugmac00:
Update - while I am still in the old, and deleted directory in terminal 1, I can create a folder with the same name in terminal 2, and also create files. I do not see the files in terminal one, as there I am not in the new directory (with the same name)... 

I should have cd'ed out of the (deleted) directory, and then in the directory with the same name, and then I would have seen the files.... funky, funky.

PR #2140, opened 2021-08-06
When cleaning dirs clean only their content

gaborbernat:
Instead, only delete their content. This allows the user to cd into these
and them being still valid after a new run.

Resolves https://github.com/tox-dev/tox/pull/2139#issuecomment-893406979


jugmac00 (approved):
LGTM! Thanks for taking care of it!

Issue #1790, opened 2021-01-10
tox4: run arbitrary command in a tox environment


uranusjr:
Moving the discussion from Discourse: https://discuss.python.org/t/tool-tox-4-on-the-horizon/6537/11?u=uranusjr

Currently tox offers ways to populate an environment without executing `command`. But to make the environment actually useful, the user needs to manually dig into the `.tox` directory themselves. It would be useful to have a tox command like

```
tox exec <env-name> -- command [args ...]
```

to populate an environment and run an arbitrary command on it. It’d be best if there’s a flag to skip the `install_reqs` step as well (i.e. just run the command I provide, don’t do anything else).

gaborbernat:
Seems tox 3 has this as a plugin https://github.com/dstanek/tox-run-command

PR #2142, opened 2021-08-07
Add support for exec subcommand

gaborbernat:
Run arbitrary commands within your tox environment. Don't abuse it.

Resolves https://github.com/tox-dev/tox/issues/1790

### DIFF
--- a/docs/changelog/1790.feature.rst
+++ b/docs/changelog/1790.feature.rst
@@ -1,2 +1,2 @@
-Add ``exec`` subcommand that allows users to run arbitrary command within the tox environment (without needing to modify
+Add ``exec`` subcommand that allows users to run an arbitrary command within the tox environment (without needing to modify
 their configuration) - by :user:`gaborbernat`.
--- a/docs/changelog/1790.feature.rst
+++ b/docs/changelog/1790.feature.rst
@@ -1,0 +1,2 @@
+Add ``exec`` subcommand that allows users to run arbitrary command within the tox environment (without needing to modify
+their configuration) - by :user:`gaborbernat`.
--- a/docs/changelog/1790.feature.rst
+++ b/docs/changelog/1790.feature.rst
@@ -1,2 +1,2 @@
-Add ``exec`` subcommand that allows users to run an arbitrary command within the tox environment (without needing to modify
-their configuration) - by :user:`gaborbernat`.
+Add ``exec`` subcommand that allows users to run an arbitrary command within the tox environment (without needing to
+modify their configuration) - by :user:`gaborbernat`.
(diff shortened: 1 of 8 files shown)


docs/changelog/1790.feature.rst
```diff
@@ -0,0 +1,2 @@
+Add ``exec`` subcommand that allows users to run arbitrary command within the tox environment (without needing to modify
```
jugmac00, line 1:
```suggestion
Add ``exec`` subcommand that allows users to run an arbitrary command within the tox environment (without needing to modify
```

jugmac00: Is it supposed to run exactly one command?

I tried some variations of quoting to run two commands, but failed :-)

jugmac00 (commented):
While there are several potentially dangerous configuration options (e.g. delete directories) already present in tox, this one is special - let's hope for the best :+1: 

@asottile Any opinion on this one?

I am not super happy with the autogenerated documentation, as at least I could not see that the command needs to be passed in after a double dash `--`.
Screenshot from 2021-08-07 11-19-43

Is there anything we could do with the autogenerated docs or maybe should we add a FAQ entry? Although - technically it is not a frequently asked question :-)

docs/changelog/1790.feature.rst
gaborbernat: Yeah, single run only. Hence in the description ``execute an arbitrary command``.

gaborbernat:
> I am not super happy with the autogenerated documentation, as at least I could not see that the command needs to be passed in after a double dash `--`.

It's documented in the epilogue of the sub-command (see tox4 `e --help`), if it's not rendering means we need to fix that in our documentation generation.

jugmac00:
> > I am not super happy with the autogenerated documentation, as at least I could not see that the command needs to be passed in after a double dash `--`.
> 
> It's documented in the epilogue of the sub-command (see tox4 `e --help`), if it's not rendering means we need to fix that in our documentation generation.

Thanks - `tox4 exec --help` indeed renders the example - the documentation does not - yet.

gaborbernat:
Created https://github.com/gaborbernat/sphinx-argparse-cli/issues/20 to address that.

gaborbernat:
> While there are several potentially dangerous configuration options (e.g. delete directories) already present in tox, this one is special - let's hope for the best

I don't think it's overly dangerous. Users can already invoke commands within the tox environments (as in `.tox/py39/bin/pip install x`). The only thing this adds is convenience to easier invocation for that, and allowing you to take advantage of tox environment isolation and altering. And we're all consenting adults here at the end of the day, not?

Issue #2041, opened 2021-04-20
[tox4] first run succeeds, second run fails


jugmac00:
```
git clone https://github.com/zopefoundation/Products.PythonScripts
```

```
tox -e py38
```
=> works
```
tox -e py38
```
=> works

```
tox4dev 
```
=> works

```
tox4dev
```
=> fails

```
❯ tox4dev -e py38
py38: commands_pre[0]> .tox/4/py38/bin/buildout -nc /home/jugmac00/All/output_zope/zopefoundation/Products.PythonScripts/buildout.cfg buildout:directory=/home/jugmac00/All/output_zope/zopefoundation/Products.PythonScripts/.tox/4/py38 buildout:develop=/home/jugmac00/All/output_zope/zopefoundation/Products.PythonScripts install test
Version and requirements information containing setuptools:
  Base installation request: 'zc.buildout', 'setuptools'
  Requirement of zc.buildout>=2.13.4: setuptools<52,>=8.0
While:
  Installing.
  Checking for upgrades.
Error: There is a version conflict.
We already have: setuptools 54.2.0
but zc.buildout 2.13.4 requires 'setuptools<52,>=8.0'.
py38: exit 1 (0.69 seconds) /home/jugmac00/All/output_zope/zopefoundation/Products.PythonScripts> .tox/4/py38/bin/buildout -nc /home/jugmac00/All/output_zope/zopefoundation/Products.PythonScripts/buildout.cfg buildout:directory=/home/jugmac00/All/output_zope/zopefoundation/Products.PythonScripts/.tox/4/py38 buildout:develop=/home/jugmac00/All/output_zope/zopefoundation/Products.PythonScripts install test pid=29999
  py38: FAIL code 1 (0.79=setup[0.10]+cmd[0.69] seconds)
  evaluation failed :( (0.82 seconds)
```

I can't do a `-rvv` as then the problem would go away...

```
❯ tox4dev -e py38 -vv
ROOT: 99 D setup logging to DEBUG on pid 31389 [tox/report.py:222]
py38: 133 I find interpreter for spec PythonSpec(implementation=CPython, major=3, minor=8) [virtualenv/discovery/builtin.py:62]
[...]

PR #2147, opened 2021-08-08
Fix virtual environment creation not cached

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/2041


jugmac00 (approved):
Awesome!

Issue #2037, opened 2021-04-19
[tox4] combination of `config` and `-r` throws KeyError/AttributeError


jugmac00:
`tox4dev` is a symlink to an editable installation of the git repo

used repository: tox

```
❯ tox4dev config -r
[testenv:fix]
type = VirtualEnvRunner
set_env =
  COVERAGE_FILE=/home/jugmac00/Projects/tox/.tox/4/.coverage.fix
  COVERAGE_PROCESS_START=/home/jugmac00/Projects/tox/setup.cfg
  PIP_DISABLE_PIP_VERSION_CHECK=1
base = testenv
runner = virtualenv
description = format the code base to adhere to our styles, and complain about what we cannot do automatically
depends = 
env_name = fix
env_dir = /home/jugmac00/Projects/tox/.tox/4/fix
env_tmp_dir = /home/jugmac00/Projects/tox/.tox/4/fix/tmp
platform = 
pass_env =
  CURL_CA_BUNDLE
  LANG
  LANGUAGE
  LD_LIBRARY_PATH
  PIP_*
  PROGRAMDATA
  PYTEST_*
  REQUESTS_CA_BUNDLE
  SSL_CERT_FILE
  TERM
  TMPDIR
  VIRTUALENV_*
  http_proxy
  https_proxy
  no_proxy
parallel_show_output = False
recreate = False
commands_pre = 
commands =
  pre-commit run --all-files --show-diff-on-failure --color=always
  python -c 'print(r"hint: run /home/jugmac00/Projects/tox/.tox/4/fix/bin/pre-commit install to add checks as pre-commit hook")'
commands_post = 
change_dir = /home/jugmac00/Projects/tox
ignore_errors = False
ignore_outcome = False
skip_install = True
package = skip
base_python = python3.9
env_site_packages_dir = /home/jugmac00/Projects/tox/.tox/4/fix/lib/python3.9/site-packages
env_bin_dir = /home/jugmac00/Projects/tox/.tox/4/fix/bin
env_python = /home/jugmac00/Projects/tox/.tox/4/fix/bin/python
deps = pre-commit>=2.2
system_site_packages = False
always_copy = False
download = False
Traceback (most recent call last):
  File "/home/jugmac00/Projects/tox/src/tox/config/main.py", line 106, in get_env
    return self._env_to_set[item]
KeyError: 'py310'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #2150, opened 2021-08-09
Support CLI run flags for tox config

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/2037

### DIFF
--- a/docs/changelog/2037.bugfix.rst
+++ b/docs/changelog/2037.bugfix.rst
@@ -1,0 +1,2 @@
+tox config fails on `--recreate <tox-config---recreate>`_ flag, and once specified the output does not reflect the
+impact of the CLI flags - by :user:`gaborbernat`.
--- a/src/tox/tox_env/runner.py
+++ b/src/tox/tox_env/runner.py
@@ -121,9 +121,10 @@ class RunToxEnv(ToxEnv, ABC):
             with package_env.display_context(suspend=self._has_display_suspended):
                 getattr(package_env, method_name)(*args)
 
-    def _clean(self, force: bool = False) -> None:
-        super()._clean(force)
-        self._call_pkg_envs("_clean")  # do not pass force along, allow package env to ignore if requested
+    def _clean(self, transitive: bool = False) -> None:
+        super()._clean(transitive)
+        if transitive:
+            self._call_pkg_envs("_clean")
 
     @property
     def _default_package_env(self) -> str:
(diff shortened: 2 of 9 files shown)


docs/changelog/2037.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+tox config fails on `--recreate <tox-config---recreate>`_ flag, and once specified the output does not reflect the
+impact of the CLI flags - by :user:`gaborbernat`.
```
jugmac00, line 2: Please rephrase this in what is fixed.

Either "tox no longer fails on..." or "fixed xyz so tox no longer..."

src/tox/tox_env/runner.py
```diff
@@ -121,9 +121,10 @@ def _call_pkg_envs(self, method_name: str, *args: Any) -> None:
             with package_env.display_context(suspend=self._has_display_suspended):
                 getattr(package_env, method_name)(*args)
 
-    def _clean(self, force: bool = False) -> None:
-        super()._clean(force)
-        self._call_pkg_envs("_clean")  # do not pass force along, allow package env to ignore if requested
+    def _clean(self, transitive: bool = False) -> None:
+        super()._clean(transitive)
```
jugmac00, line 125: The parent class's `_clean` does not evaluate the `transitive` argument, but maybe it is better to pass it anyway, as it is probably not a good idea to have different method signatures between parent and child class.

src/tox/tox_env/runner.py
```diff
@@ -121,9 +121,10 @@ def _call_pkg_envs(self, method_name: str, *args: Any) -> None:
...
+        if transitive:
+            self._call_pkg_envs("_clean")
```
jugmac00, line 127: So, do I get this right?  If `_clean` with `transitive` gets called, all existing envs get recreated?

jugmac00 (approved):
Works! 

Looks good, with some inline comments.

src/tox/tox_env/runner.py
gaborbernat: You can't have different interface and expect the call to work so we have no option on this. Furthermore just because it's not using today doesn't mean that will never use it.

src/tox/tox_env/runner.py
gaborbernat: Yes all transitive environments. Currently this means packaging environments attached to the run environment. But you might have additional environments attached to the run environment that might need be cleaned.

docs/changelog/2037.bugfix.rst
gaborbernat: This snippet shows up under the bug fixes section so I think the current wording in that context is clear to me ,🤔

jugmac00: I just scanned our history and it looks like we use both styles. No hard feelings If you keep it that way, though I like the other one better 😉

jugmac00: Ship it!

gaborbernat: I'll merge it as is, but if you put in a PR that converts it to a single format I'll accept it 🙏👍

Issue #1840, opened 2021-01-17
tox4: ignore_basepython_conflict support


gaborbernat:
https://tox.readthedocs.io/en/latest/config.html#conf-ignore_basepython_conflict

PR #2156, opened 2021-08-12
Check base python against env name and raise if differs

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/1840

Allow enforcing the env name via ignore_base_python_conflict config
value.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -103,7 +103,7 @@ Core
    :version_added: 3.2.0
 
    Specify a list of `PEP-508 <https://www.python.org/dev/peps/pep-0508/>`_ compliant dependencies that must be
-   satisfied in the python environment hosting tox when running the tox command. If any of these dependencies are not
+   satisfied in the Python environment hosting tox when running the tox command. If any of these dependencies are not
    satisfied will automatically create a provisioned tox environment that does not have this issue, and run the tox
    command within that environment. See :ref:`provision_tox_env` for more details.
 
@@ -417,10 +417,10 @@ Python
    :default: {package_root}
 
    Name or path to a Python interpreter which will be used for creating the virtual environment, first one found wins.
-   This determines in practice the python for what we'll create a virtual isolated environment. Use this to specify the
-   python version for a tox environment. If not specified, the virtual environments factors (e.g. name part) will be
+   This determines in practice the Python for what we'll create a virtual isolated environment. Use this to specify the
+   Python version for a tox environment. If not specified, the virtual environments factors (e.g. name part) will be
    used to automatically set one. For example, ``py310`` means ``python3.10``, ``py3`` means ``python3`` and ``py``
-   means ``python``. If the name does not match this pattern the same python version tox is installed into will be used.
+   means ``python``. If the name does not match this pattern the same Python version tox is installed into will be used.
 
     .. versionchanged:: 3.1
 
@@ -430,10 +430,10 @@ Python
 
     .. note::
 
-      Leaving this unset will cause an error if the package under test has a different python requires than tox itself
-      and tox is installed into a python that's not supported by the package. For example, if your package requires
-      python 3.9 or later, and you install tox in python 3.8, when you run a tox environment that has left this
-      unspecified tox will use python 3.8 to build and install your package which will fail given it requires 3.9.
+      Leaving this unset will cause an error if the package under test has a different Python requires than tox itself
+      and tox is installed into a Python that's not supported by the package. For example, if your package requires
+      Python 3.9 or later, and you install tox in Python 3.8, when you run a tox environment that has left this
+      unspecified tox will use Python 3.8 to build and install your package which will fail given it requires 3.9.
 
 
 .. conf::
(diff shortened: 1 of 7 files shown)


tests/tox_env/python/test_python_api.py
```diff
@@ -65,3 +66,61 @@ def test_diff_msg_added_removed_changed() -> None:
...
+    project = tox_project({"tox.ini": ini})
+    result = project.run("c", "-e", f"py{py_ver}", "-k", "base_python")
+    result.assert_success()
+    if ignore_base_python_conflict:
+        out = f"[testenv:py{py_ver}]\nbase_python = py{py_ver}\n"
+    else:
+        coma_in_exc = sys.version_info[0:2] <= (3, 6)
+        out = (
+            f"[testenv:py{py_ver}]\nbase_python = # Exception: Fail('env name py{py_ver} conflicting with"
+            f" base python py{py_ver_next}'{',' if coma_in_exc else ''})\n"
```
jugmac00, line 124: Should this be `comma_in_exc`?

tests/tox_env/python/test_python_api.py
```diff
@@ -65,3 +66,61 @@ def test_diff_msg_added_removed_changed() -> None:
...
+        if ignore_base_python_conflict:
+            result = Python._validate_base_python(env_name, base_python, ignore_base_python_conflict)
+            assert result == [env_name]
+        else:
+            msg = f"env name {env_name} conflicting with base python {conflict[0]}"
+            with pytest.raises(Fail, match=msg):
+                Python._validate_base_python(env_name, base_python, ignore_base_python_conflict)
+    else:
+        result = Python._validate_base_python(env_name, base_python, ignore_base_python_conflict)
+        assert result is base_python
```
jugmac00, line 105: I am not a big fan of using `if` and `else` in tests and I would not mind to write three small tests, which would duplicate a couple of lines, but then would be easier to reason about - but that is possible a personal preference :-)

jugmac00 (changes requested):
With a `tox.ini` as the following...

```
[tox]
envlist =
    py36,
    py37,
    py38,
    py39,
    pre-commit,
    mypy,
    coverage,
    docs,
ignore_basepython_conflict = true

[testenv]
basepython = python3.9
description = run the tests with pytest
extras = test
deps = pdbpp
commands = pytest {posargs}

[testenv:py36]
commands = python -c "print('hello')"
```

... there are errors no matter whether `ignore_basepython_conflict` is set to `true` or `false`.

```
❯ tox4dev -e py36 -r  # ignore... = false
py36: remove tox env folder /home/jugmac00/Projects/flask-reuploaded/.tox/4/py36
.pkg: remove tox env folder /home/jugmac00/Projects/flask-reuploaded/.tox/4/.pkg
py36: failed with env name py36 conflicting with base python python3.9
  py36: FAIL code 1 (0.00 seconds)
  evaluation failed :( (0.03 seconds)

❯ tox4dev -e py36 -r  # ignore... = true
py36: remove tox env folder /home/jugmac00/Projects/flask-reuploaded/.tox/4/py36
.pkg: remove tox env folder /home/jugmac00/Projects/flask-reuploaded/.tox/4/.pkg
py36: failed with env name py36 conflicting with base python python3.9
  py36: FAIL code 1 (0.00 seconds)
  evaluation failed :( (0.03 seconds)
```

While the `tox3` docs state this will become an error in a future version independent of the setting, this does not match the documentation in this pull request. Also, I remember you said we try to stay compatible with `tox3` as much as possible.

docs/config.rst
```diff
@@ -448,10 +448,9 @@ Python
     do so. To help with sanity of users a error will be raised whenever the environment name version does not matches
     up with this expectation.
 
-    Furthermore, we allow hard enforcing this rule (and bypassing the warning) by setting this flag to ``true``. In such
-    cases we ignore the :ref:`base_python` and instead always use the base python implied from the Python name. This
-    allows you to configure :ref:`base_python` in the :ref:`base` section without affecting environments that have
-    implied base python versions.
+    Furthermore, we allow hard enforcing this rule by setting this flag to ``true``. In such cases we ignore the
+    :ref:`base_python` and instead always use the base python implied from the Python name. This allows you to configure
```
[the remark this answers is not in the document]
gaborbernat: Sure, capital all the way.

tests/tox_env/python/test_python_api.py
gaborbernat: IMHO It's a balancing act :blush: made some changes, but did not take it all the way.

gaborbernat:
> I remember you said we try to stay compatible with `tox3` as much as possible.

With some exceptions, this is one of those. And you're right, I interpreted this config as env specific, but is global in tox 3 (hence why wasn't working with your example). Now should be ok.

tests/tox_env/python/test_python_api.py
jugmac00: Thank you!

jugmac00 (commented):
Works like a charm!

PR #2162, opened 2021-08-18
Add LC_ALL to implicit tox passenv

ssbarnea:
This is needed because LANG by itself is not enough to help newer pip from failing on py36. Forcing tox users to manually alter not only their system configuration but also their tox.ini files to avoid that bug is not quite desired. As we already pass `LANG` and `LANGUAGE` I see not reason why we should not do the same for `LC_ALL` too.

Related: https://github.com/pypa/pip/issues/10219

## Contribution checklist:

- [x] added/updated test(s)
      in message body

### DIFF
--- a/docs/changelog/2162.bugfix.rst
+++ b/docs/changelog/2162.bugfix.rst
@@ -1,0 +1,1 @@
+include ``LC_ALL`` to implicit list of passenv variables - by :user:`ssbarnea`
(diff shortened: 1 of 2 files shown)


jpodivin:
We could really use this one. Otherwise we have to set the language for each version of tox.ini in our projects.

gaborbernat:
released via https://pypi.org/project/tox/3.24.2/

PR #2164, opened 2021-08-20
ASCII output for --parallel report lines

### LINKED ISSUE #1421: --parallel: Unicode output for status in ASCII mode
In 43d65b177f6cb9a8757554bf8988f59018b92c0f, an ASCII spinner was added if the output does not support unicode:

https://github.com/tox-dev/tox/blob/7a8a6acf5f72472dadcc982c2e418c7e7a52e6fb/src/tox/util/spinner.py#L45-L49

However, the final status outputs still contain unicode characters which might not be rendered properly (or even cause exceptions?):

https://github.com/tox-dev/tox/blob/7a8a6acf5f72472dadcc982c2e418c7e7a52e6fb/src/tox/util/spinner.py#L107-L114

Not sure what an appropriate output would be - I guess no symbols at all, or maybe ":)", ":(" and ":|" similar to what happens without `--parallel` :wink:

brettcs:
This is a branch to fix #1421 by providing ASCII equivalent messages for the OK/FAIL/SKIP lines that summarize the result for each environment.

I chose strings reminiscent of the sysvinit boot service banners on older Linux distros. If you'd rather use something else, of course it's no problem for anyone else to switch out those strings. The code should continue to work no problem.

Hope this helps!

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -20,6 +20,7 @@ Bastien Vallet
 Benoit Pierre
 Bernat Gabor
 Brett Langdon
+Brett Smith
 Bruno Oliveira
 Carl Meyer
 Charles Brunet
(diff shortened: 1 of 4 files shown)


brettcs:
Sorry about the test failures on older Pythons. I can tackle those but it might be a few days.

gaborbernat (approved):

PR #2191, opened 2021-08-30
Internal changes to support tox-gh-actions

gaborbernat:
For https://github.com/gaborbernat/tox-gh-actions/blob/support-v4/src/tox_gh_actions/plugin.py#L1-L108

### DIFF
--- a/src/tox/plugin/manager.py
+++ b/src/tox/plugin/manager.py
@@ -15,6 +15,7 @@ from tox.tox_env.python.virtual_env import runner
 from tox.tox_env.python.virtual_env.package import api
 from tox.tox_env.register import REGISTER, ToxEnvRegister
 
+from ..config.main import Config
 from . import NAME, spec
 from .inline import load_inline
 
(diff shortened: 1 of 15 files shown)


src/tox/plugin/manager.py
```diff
@@ -15,6 +15,7 @@
 from tox.tox_env.python.virtual_env.package import api
 from tox.tox_env.register import REGISTER, ToxEnvRegister
 
+from ..config.main import Config
```
jugmac00, line 18: Why not `from tox.config.main import Config`?

jugmac00 (approved):
Any chance to get the uncovered lines under test?

How did you test this? Did you actually run a modified version of `tox-gh-actions` with your branch?

Is it necessary that `tox-gh-actions` perform any changes or is it already 100% compatible with branch of tox4?

src/tox/plugin/manager.py
gaborbernat: Prefer relative imports where possible :thinking:

gaborbernat:
> Any chance to get the uncovered lines under test?

Yes :blush: 
 
> How did you test this? Did you actually run a modified version of `tox-gh-actions` with your branch?

Yes, locally.

> Is it necessary that `tox-gh-actions` perform any changes or is it already 100% compatible with this branch of tox4?

No, it's already compatible.

jugmac00 (approved):

PR #2192, opened 2021-08-31
Make 3.24.3 a headline in the changelog

hroncok:


hroncok:
I didn't add a changelog fragment for this change, as I consider it trivial.

Here is the current way of how this looks:

PR #2198, opened 2021-09-02
Fix usedevelop without setup.py

### LINKED ISSUE #2197: usedevelop without setup.py failure
When `usedevelop` is enabled with a project that only has a `setup.cfg` but no `setup.py`, on the second tox execution with no changes at all, tox crashes because `VirtualEnv._needs_reinstall` relies on `setup.py` to get the name of the project's package.

A PR is coming next.

AntoineD:
Fix #2197 

I was able to check the fix on an external project but unfortunately the new test does not work when the `cmd` fixture is called the second time (error at teardown).
Please let me know how to fix this test or feel free to change it.

### DIFF
--- a/src/tox/venv.py
+++ b/src/tox/venv.py
@@ -7,6 +7,7 @@ import sys
 from itertools import chain
 
 import py
+import setuptools
 
 import tox
 from tox import reporter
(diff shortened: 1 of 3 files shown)


src/tox/venv.py
```diff
@@ -323,20 +324,27 @@ def finish(self):
...
-            returnout=True,
-            env=env,
-            capture_err=False,
-        )
-        name = next(
-            (i for i in output.split("\n") if i and not i.startswith("pydev debugger:")),
-            "",
-        )
+
+        if setup_py.exists():
```
gaborbernat, line 329: If the setup.py does not exist, this here https://github.com/tox-dev/tox/blob/954399ba18/src/tox/package/builder/legacy.py#L37 will fail earlier

gaborbernat (changes requested):
Not sure how you even get that far.

gaborbernat (commented):
If you can make it work for both old and new world I'm happy to accept this PR.

AntoineD:
What do you mean by *old and new world*?

gaborbernat:
It should work with any setuptools/pip version the user might use.

AntoineD:
It seems to me that it does not worth the effort then, as compared to creating an almost empty `setup.py`.

PR #2203, opened 2021-09-05
venv: do not prepend a truncated shebang interpreter

jdknight:
When a user invokes a `tox` call with `TOX_LIMITED_SHEBANG` set, any shebang-based scripts will have their interpreter's extracted and explicitly executed. An issue with the initial implementation is that if the configured interpreter is truncated to a path of another executable, the wrong executable can be invoked instead (see also \[1\]\[2\]). To prevent this, an extra byte is read to ensure the length of the extracted interpreter is within the configured limit (`MAXINTERP`); otherwise the call will know the interpreter has been truncated and will fallback onto the original argument list.

\[1\]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/fs/binfmt_script.c?id=8099b047ecc431518b9bb6bdbba3549bbecdc343
\[2\]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/fs/binfmt_script.c?id=b5372fe5dc84235dbe04998efdede3c4daa866a9

### DIFF
--- a/docs/changelog/2208.bugfix.rst
+++ b/docs/changelog/2208.bugfix.rst
@@ -1,0 +1,2 @@
+Prevent tox from using a truncated interpreter when using
+``TOX_LIMITED_SHEBANG`` -- by :user:`jdknight`.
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
Please add a changelog for this.

gaborbernat (approved):

gaborbernat:
Released via https://tox.readthedocs.io/en/latest/changelog.html#v3-24-4-2021-09-16

PR #2209, opened 2021-09-10
Make ConfigSet ABC and add contains to Loader

gaborbernat:
The ConfigSet should be always specialized via subclassing. The contains
is a quality of life imporvement for plugins. Also bumped the tools
while at it.

### DIFF
--- a/tests/config/test_sets.py
+++ b/tests/config/test_sets.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Callable, Dict, Optional, Set, TypeVar
 
 import pytest
+from pytest_mock import MockerFixture
 
 from tests.conftest import ToxIniCreator
 from tox.config.cli.parser import Parsed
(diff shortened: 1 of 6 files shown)


tests/config/test_sets.py
```diff
@@ -150,3 +156,8 @@ def __init__(self, conf: Config):
     assert repr(conf) == "MagicConfigSet(loaders=[IniLoader(section=<Section: magic>, overrides={})])"
 
     assert isinstance(result.state.conf.options, Parsed)
+
+
+def test_do_not_allow_create_config_set(mocker: MockerFixture) -> None:
+    with pytest.raises(TypeError, match="Can't instantiate"):
+        ConfigSet(mocker.create_autospec(Config))  # type: ignore # the type checker also warns that ABC
```
jugmac00, line 163: Looks like the comment is cut off... `that ABC`... does what?

jugmac00 (approved):

tests/config/test_sets.py
gaborbernat: It's short but I thought it's enough, so it's on purpose.

PR #2217, opened 2021-09-15
Expose the configuration load mechanism to plugins

### LINKED ISSUE #2200: tox 4: enable plugins to discover config sections & testenvs
My plugin [tox-docker](https://github.com/tox-dev/tox-docker) could benefit from being able to discover all configuration sections and/or to get a list of all testenvs from the `Config`.

tox-docker starts docker containers as configured by each testenv; the testenv references another section (eg `[docker:nginx]` by way of saying `docker=nginx` in the testenv). In order to discover all configuration, I need to enumerate all testenvs or all sections (and filter the list to `docker:*` sections).

I'm currently doing the former with:

```python
class EnvDockerConfigSet(ConfigSet):
    def __init__(self, conf: Config) -> None:
        super().__init__(conf)

        self.add_config(
            keys=["docker"],
            of_type=EnvList,
            default=EnvList([]),
            desc="docker image configs to load",
            post_process=list,
        )

def discover_container_configs(config: Config) -> Sequence[str]:
    docker_configs = set()
    for env_name in config:
        env_config = config.get_section_config(f"testenv:{env_name}", EnvDockerConfigSet)
        docker_configs.update(env_config.load("docker"))
    return list(docker_configs)
```

It would be more explicit to have eg. `config.get_test_envs()` that returns either all testenvs, or better yet, the specific testenvs that will be run (eg due to setting `-e` on the commandline, or the `envlist`, etc).

gaborbernat:
Also, add the `tox_add_env_config` plugin hook and add the main
configuration object to `tox_add_core_config`. Move the env list
manifestation methods from state to the main configuration object to
allow use within plugins.

Resolves #2200.


jugmac00 (approved):

dcrosta:
Since this change, it seems like `get_section_config` is no longer usable without reaching into private members -- the first arg is now a `Section`, which plugins don't have a handy way to get (internal tox code gets it via `self._src.get_tox_env_section(name)`).

If `self._src` were public (or if there were an accessor) I'd feel comfortable using it, but especially given how quickly the API is changing right now, I definitely don't want to reach into private members.

gaborbernat:
it's an abstract section, so you can pass in e.g. `Section('docker', 'nginx')`; we'll do the translation internally to map it to `docker:nginx` for ini files, but for pyproject.toml files likely something like `tool.tox.docker.nginx` :slight_smile:

however, you can query an iteration of available sections via the `sections` on the main configuration object see https://github.com/tox-dev/tox/blob/rewrite/tests/plugin/test_plugin.py#L109

dcrosta:
Ah! I didn't even think to look at what that class is, I assumed it was stateful. Got it, thanks!

PR #2223, opened 2021-09-17
Support for environment files in set_env

### LINKED ISSUE #1938: Support environment files and comments in setenv
As added in #1668.

gaborbernat:
Resolves #1938.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -271,15 +271,15 @@ Base options
    :keys: set_env, setenv
 
    A dictionary of environment variables to set when running commands in the tox environment. Lines starting with a
-   ``file|`` prefix are considered environment files to load with their location being the content minus this prefix.
+   ``file|`` prefix define the location of environment file.
 
     .. note::
 
        Environment files are processed using the following rules:
 
-       - blank lines ignored,
+       - blank lines are ignored,
        - lines starting with the ``#`` character are ignored,
-       - each line is in KEY=VALUE format; both the key and the value is stripped,
+       - each line is in KEY=VALUE format; both the key and the value are stripped,
        - there is no special handling of quotation marks, they are part of the key or value.
 
 .. conf::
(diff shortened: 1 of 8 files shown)


docs/config.rst
```diff
@@ -270,7 +270,17 @@ Base options
...
 
-   A dictionary of environment variables to set when running commands in the tox environment.
+   A dictionary of environment variables to set when running commands in the tox environment. Lines starting with a
+   ``file|`` prefix are considered environment files to load with their location being the content minus this prefix.
+
+    .. note::
+
+       Environment files are processed using the following rules:
+
+       - blank lines ignored,
```
jugmac00, line 280:
```suggestion
       - blank lines are ignored,
```

docs/config.rst
```diff
@@ -270,7 +270,17 @@ Base options
...
+       - lines starting with the ``#`` character are ignored,
+       - each line is in KEY=VALUE format; both the key and the value is stripped,
```
jugmac00, line 282:
```suggestion
       - each line is in KEY=VALUE format; both the key and the value are stripped,
```

docs/config.rst
```diff
@@ -270,7 +270,17 @@ Base options
 .. conf::
    :keys: set_env, setenv
 
-   A dictionary of environment variables to set when running commands in the tox environment.
+   A dictionary of environment variables to set when running commands in the tox environment. Lines starting with a
+   ``file|`` prefix are considered environment files to load with their location being the content minus this prefix.
```
jugmac00, line 274:
```suggestion
   ``file|`` prefix define the location of environment files.
```

I think it is both clear that the file prefix is omitted and that a configuration file will be loaded If given.

What maybe is missing... When using a relative path, which base path is asumed? The directory of the tox configuration file?

And are placeholders like `toxinidir` supported?

jugmac00 (approved):
LGTM with some annotations

docs/config.rst
gaborbernat: Within the file itself no, within the `file|` part yes.

gaborbernat: Actually, it wasn't, but now I've fixed that part.

PR #2253, opened 2021-10-18
Fix typos

kianmeng:


src/tox/logs/env.py
```diff
@@ -28,7 +28,7 @@ def set_installed(self, packages):
 
     def set_header(self, installpkg):
         """
-        :param py.path.local installpkg: Path ot the package.
+        :param py.path.local installpkg: Path of the package.
```
jugmac00, line 31: I am not native speaker, but I think that would be `Path to the package`.

```suggestion
        :param py.path.local installpkg: Path to the package.
```

jugmac00 (approved):

gaborbernat (approved):

PR #2275, opened 2021-11-16
Drop python 3.6 support

gaborbernat:
:+1:


jugmac00:
I'll have a look this evening :+1:

src/tox/execute/local_sub_process/read_via_thread_unix.py
```diff
@@ -1,10 +1,12 @@
 """
 On UNIX we use select.select to ensure we drain in a non-blocking fashion.
 """
+from __future__ import annotations
+
 import errno  # pragma: win32 no cover
 import os  # pragma: win32 no cover
 import select  # pragma: win32 no cover
-from typing import Callable, Optional  # pragma: win32 no cover
```
jugmac00, line 7: Was the `no-cover` drop on purpose?

src/tox/execute/local_sub_process/read_via_thread_windows.py
```diff
@@ -1,21 +1,25 @@
...
 from asyncio.windows_utils import BUFSIZE  # pragma: win32 cover
 from time import sleep  # pragma: win32 cover
-from typing import Callable, Optional  # pragma: win32 cover
+from typing import Callable
 
 import _overlapped  # type: ignore[import]  # pragma: win32 cover
 
 from .read_via_thread import ReadViaThread  # pragma: win32 cover
 
+# mypy: warn-unused-ignores=false
```
jugmac00, line 15: What is the reason we need this now but not when supporting 3.6?

src/tox/plugin/inline.py
```diff
@@ -1,11 +1,12 @@
...
 import sys
 from pathlib import Path
 from types import ModuleType
-from typing import Optional
 
 
-def load_inline(path: Path) -> Optional[ModuleType]:
+def load_inline(path: Path) -> ModuleType | None:
     # nox uses here the importlib.machinery.SourceFileLoader but I consider this similarly good, and we can keep any
     # name for the tox file, it's content will always be loaded in the this module from a system point of view
```
jugmac00, line 11:
```suggestion
    # name for the tox file, its content will always be loaded in the this module from a system point of view
```

jugmac00 (approved):
LGTM with some comments

src/tox/config/loader/convert.py
```diff
@@ -78,10 +80,7 @@ def _to_typing(self, raw: T, of_type: Type[V], factory: Factory[V]) -> V:
                     new_type = next(i for i in args if i != none)  # pragma: no cover # this will always find a element
                     result = self.to(raw, new_type, factory)
         elif origin in (Literal, type(Literal)):
-            if sys.version_info >= (3, 7):  # pragma: no cover (py37+)
-                choice = of_type.__args__
-            else:  # pragma: no cover (py38+)
-                choice = of_type.__values__  # type: ignore[attr-defined]
+            choice = of_type.__args__  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
gaborbernat: There's an if statement that does something different on 3.6 than above. We removed the iff statement :thinking: how is that not related?

src/tox/execute/local_sub_process/read_via_thread_windows.py
gaborbernat: Partially unrelated. The rewrite of files caused some unused ignore to show up. So I moved the global do not warn unusued ignores to file level. We need this settings in this file because it's the only way to make mypy pass on windows and linux too (as some ignores are windows only or linux only).

src/tox/execute/local_sub_process/read_via_thread_unix.py
gaborbernat: I don't think so :laughing:

PR #2289, opened 2021-12-20
Have the --colored default value be yes in the docs

### LINKED ISSUE #2288: tox4: --colored is yes by default
With tox4a9, contrary to the [docs](https://tox.wiki/en/rewrite/cli_interface.html#tox---colored) `--colored` is yes by default.
What is the expected default?

AntoineD:
Fixes #2288.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -5,6 +5,7 @@ from datetime import date, datetime
 from pathlib import Path
 from subprocess import check_output
 from typing import Any, cast
+import os
 
 from docutils.nodes import Element, Node, Text, container, fully_normalize_name, literal, paragraph, reference, strong
 from docutils.parsers.rst.directives import flag, unchanged, unchanged_required
@@ -85,6 +86,8 @@ intersphinx_mapping = {
 }
 nitpicky = True
 nitpick_ignore = []
+
+os.environ["FORCE_COLOR"] = "yes"  # force --colored default value to be yes
 
 
 def skip_member(app: Sphinx, what: str, name: str, obj: Any, would_skip: bool, options: Options) -> bool:  # noqa: U100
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -86,6 +87,9 @@
 nitpicky = True
 nitpick_ignore = []
 
+# Force --colored default value to be yes.
+os.environ["FORCE_COLOR"] = "yes"
```
gaborbernat, line 91: can you put the command to the right-hand side here?

AntoineD:
Is the changelog fragment really needed here?

gaborbernat:
> Is the changelog fragment really needed here?

We can ignore that now.

docs/conf.py
AntoineD: Sorry I do not understand what you mean.

gaborbernat:
```python
os.environ["FORCE_COLOR"] = "yes"   # ensure by default we documented colored
```

AntoineD: I see now (you wrote command instead of comment).

gaborbernat: my bad :laughing:

docs/conf.py
```diff
@@ -86,6 +87,8 @@
 nitpicky = True
 nitpick_ignore = []
 
+os.environ["FORCE_COLOR"] = "yes"  # Force --colored default value to be yes.
```
gaborbernat, line 90: can you make the comment lowercase and remove the period? generally, I don't write it as a sentence unless there's multiple sentences

AntoineD: Sure (I just tend to follow pep8 by default).

gaborbernat (approved):

PR #2293, opened 2021-12-28
Move CI to Github Actions

gaborbernat:


.github/workflows/check.yml
```diff
@@ -0,0 +1,161 @@
...
+          import codecs
+          import os
+          import platform
+          import sys
+          cpy = platform.python_implementation() == "CPython"
+          base =("{}{}{}" if cpy else "{}{}").format("py" if cpy else "pypy", *sys.version_info[0:2])
+          env = "TOXENV={}\n".format(base)
+          print("Picked:\n{}for{}".format(env, sys.version))
+          with codecs.open(os.environ["GITHUB_ENV"], "a", "utf-8") as file_handler:
+               file_handler.write(env)
```
asottile, line 59: this can probably just run `tox -e py` maybe?

asottile (approved):

.github/workflows/check.yml
gaborbernat: That would not work for two reasons:

- we're currently always using tox with 3.10 to run the test suite of the tox tool, and -e py would mean we'd always test under 3.10.
- we'd have to massage the coverage filenames to differentiate between runs, as uploading the same artefact twice operates in overwrite mode.

asottile: ahh makes sense

PR #2296, opened 2022-01-03
avoid double-running for in-repo branches

asottile:
noticed that we ran twice as many checks as needed in #2295

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,6 +1,8 @@
 name: check
 on:
   push:
+    branches: [master, 'test-me-*']
+    tags:
   pull_request:
   schedule:
     - cron: "0 8 * * *"


.github/workflows/check.yml
```diff
@@ -1,6 +1,8 @@
 name: check
 on:
   push:
+    branches: [master, 'test-me-*']
```
gaborbernat, line 4: Can we remove the test-me? and explain why this is needed?

.github/workflows/check.yml
```diff
@@ -1,6 +1,8 @@
...
+    tags:
```
gaborbernat, line 5: Empty tags mean run on all tags? Or is it do not run it on tags?

.github/workflows/check.yml
asottile: this is useful for testing CI without having to open a PR or push to master -- we have a similar setting in pytest

.github/workflows/check.yml
asottile: all tags

.github/workflows/check.yml
gaborbernat: But you shouldn't open PRs to core, you can enable the GHA for your fork if you want to test. And that's not sharing resources with upstream so I'm -1 on this.

asottile: then I have to open junk PRs on my fork which I'd rather not do either -- this will let me just create branches to test them

gaborbernat: Why would you open a PR if you leave branches enabled for GHA, see https://github.com/gaborbernat/tox/actions

I still don't understand why this fixes it, and why we have double runs without this. I don't want you polluting the upstream repositories with stray branches. I understand for pre-commit.ci is a must, but for anything else, I'd rather not. Forks are the place to test what you want.

asottile: release branches (as well as any changes made by github ui by any of the maintainers, such as typo fixes) will double-run everything because `push:` right now is blanket-enabled

gaborbernat: Why is that? What triggers the double-run?

asottile: one run is triggered for `push:` and one run is triggered for `pull_request:`

gaborbernat: They're only the same build if you have rebased your Pr against the branch you're targeting, which is not always the case. In that case helpful to have two builds 🤔

gaborbernat (approved):

PR #2316, opened 2022-01-11
venv: Do not fail for test commands that don't exist with "-" leader

### LINKED ISSUE #2315: Ignore missing commands when ignoring errors
Commands which fail, even at (for example) dynamic loader stage get a pass if there's a "-" leader to indicate to ignore errors. That means that any user using this must accept that there's a chance that the command will not execute whatsoever.

Since these outcomes are the same as the executable not existing at all, we should allow "-" to also silence the command not existing in the first place.

cdown:
Commands which fail, even at (for example) dynamic loader stage get a
pass if there's a "-" leader to indicate to ignore errors. That means
that any user using this must accept that there's a chance that the
command will not execute whatsoever.

Since these outcomes are the same as the executable not existing at all,
allow "-" to also silence the command not existing in the first place.

As an example, with:

    [testenv:bogus]
    commands =
	-bogus

Before:

    ERROR: InvocationError for command could not find executable bogus
    ________________________ summary ________________________
    ERROR:   bogus: commands failed

After:

    WARNING: command not found but explicitly ignored
    cmd: bogus
    ________________________ summary ________________________
    bogus: commands succeeded
    congratulations :)

Closes #2315.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -24,6 +24,7 @@ Brett Smith
 Bruno Oliveira
 Carl Meyer
 Charles Brunet
+Chris Down
 Chris Jerdonek
 Chris Rose
 Clark Boylan
--- a/docs/changelog/2315.feature.rst
+++ b/docs/changelog/2315.feature.rst
@@ -1,0 +1,2 @@
+Ignore missing commands if they are prefixed by ``-``
+-- by :user:`cdown`.
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -264,7 +264,7 @@ Complete list of settings that you can put into ``testenv*`` sections:
 
     Commands will execute one by one in sequential fashion until one of them fails (their exit
     code is non-zero) or all of them succeed. The exit code of a command may be ignored (meaning
-    they are always considered successful) by prefixing the command with a dash (``-``) - this is
+    they are always considered successful even if they don't exist) by prefixing the command with a dash (``-``) - this is
     similar to how ``make`` recipe lines work. The outcome of the environment is considered successful
     only if all commands (these + setup + teardown) succeeded (exit code ignored via the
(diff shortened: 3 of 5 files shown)


jugmac00:
Could you elaborate about your use case?

Usually `-` is used to make a tox run pass even when e.g. a linter fails.

This makes sense, as the linter output is printed to stdout so it does not go by completely unnoticed but CI still succeeds - think of e.g. some whitespace issues detected by flake8.

To me an unavailable command is something completely different. That indicates a broken setup.

Without knowing more about a good reason I am -1 on this.

cdown:
@jugmac00 The use case is from #python IRC:

```
15:02:04 <NoImNotNi> hm. okay, so, using tox, in my commands, i have a bunch of 'optional' commands (where i don't want their failure to prevent a successful tox run), so i prefix them with a hyphen as per tox docs, and this works fine, for the most part
15:02:36 <NoImNotNi> but for some commands, maybe a dependency isn't installed (e.g. npm), and so tox gets a ERROR: InvocationError: could not find executable 'npx'
15:03:09 <NoImNotNi> and this class of error apparently doesn't get ignored by the leading hyphen? because it causes the overall tox job to fail.
```

Using `-` to solely indicate "pass even if [some application] fails" is error-prone, because (for example) we could even be failing to load Python or some FFI dynamic libraries and it would still pass. Thus anyone using `-` must accept that it will also pass on said broken setups. This just codifies that that is the case, it doesn't change the existing paradigm.

gaborbernat (approved):

asottile:
there is at least precedent for this in `make` -- as weird as it seems:

```console
$ cat Makefile 
all:
	-wat
$ make all
wat
make: wat: Command not found
make: [Makefile:2: all] Error 127 (ignored)
```

PR #2383, opened 2022-03-21
Add PROGRAM environment variables on Windows

### LINKED ISSUE #2382: Pass through ProgramFiles* and ProgramData by default
I'd like to follow up on #1673. I couldn't comment or even +1 that issue.

I believe the characterization of the issue as setuptools-specific was incorrect.

Although the issue does affect Setuptools and not Flit, that's only because Setuptools supports building C extensions on Windows.

The issue can't be fixed in Setuptools. By the time tox has masked the ProgramFiles env var, Setuptools has little hope of recovering that setting.

We're not asking tox to patch Setuptools. Instead, we're asking tox to consider honoring the system's intrinsic environment configuration by passing through system-level variables, variables that would be unlikely to be configured by the user and which are necessary for basic operation on the platform (similar to `PATH`, `SYSTEMDRIVE`, and `SYSTEMROOT`).

Failing to support this model will instead require every project that builds extension modules on Windows to bypass this setting in their tox config. I believe this issue was largely missed until recently because most users were still testing on older platforms/compilers that did not rely on vswhere for discovery, but now that Windows 2022 is the default in Github, it's affecting a wide array of users. I contend:

- this exemption is required in every case it affects
- there is no known case that this masking is currently important
- the number of affected projects is large.

I can put together a repro that doesn't involve Setuptools if that helps persuade.

jaraco:
Fixes #2382

## Contribution checklist:

- [x] added/updated test(s)
      in message body
    ```rst
    ```

### DIFF
--- a/docs/changelog/2382.feature.rst
+++ b/docs/changelog/2382.feature.rst
@@ -1,0 +1,1 @@
+On Windows, PROGRAMFILES, PROGRAMFILES(X86), and PROGRAMDATA environment variables are now passed through.
--- a/docs/changelog/2382.feature.rst
+++ b/docs/changelog/2382.feature.rst
@@ -1,1 +1,1 @@
-On Windows, PROGRAMFILES, PROGRAMFILES(X86), and PROGRAMDATA environment variables are now passed through, unmasking system values necessary to locate resources such as a C compiler.
+On Windows ``PROGRAMFILES``, ``PROGRAMFILES(X86)``, and ``PROGRAMDATA`` environment variables are now passed through, unmasking system values necessary to locate resources such as a C compiler.
--- a/docs/changelog/2382.feature.rst
+++ b/docs/changelog/2382.feature.rst
@@ -1,1 +1,1 @@
-On Windows, PROGRAMFILES, PROGRAMFILES(X86), and PROGRAMDATA environment variables are now passed through.
+On Windows, PROGRAMFILES, PROGRAMFILES(X86), and PROGRAMDATA environment variables are now passed through, unmasking system values necessary to locate resources such as a C compiler.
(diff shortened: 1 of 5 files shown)


docs/changelog/2382.feature.rst
```diff
@@ -0,0 +1 @@
+On Windows, PROGRAMFILES, PROGRAMFILES(X86), and PROGRAMDATA environment variables are now passed through.
```
jugmac00, line 1: Maybe you could mention why this is a good thing?

src/tox/config/__init__.py
```diff
@@ -807,6 +807,11 @@ def passenv(testenv_config, value):
             passenv.add("PROCESSOR_ARCHITECTURE")  # platform.machine()
             passenv.add("USERPROFILE")  # needed for `os.path.expanduser()`
             passenv.add("MSYSTEM")  # fixes #429
+
```
jugmac00, line 810: I think this empty line is not necessary :-)

jugmac00 (approved):
Thanks, Jason! Looks good!

docs/changelog/2382.feature.rst
jaraco: Done in 3eff20c7.

src/tox/config/__init__.py
jaraco: Hmm. Perhaps you're right... The thing I'm trying to communicate with the newline is that the comment applies to a group of items. What I really want is for the next variable added to put a blank line before it so that the group of three items is clear from the syntax. I'll just remove it and let future people figure it out however.

jaraco: Fixed in aa94f9d9.

docs/changelog/2382.feature.rst
```diff
@@ -0,0 +1 @@
+On Windows, PROGRAMFILES, PROGRAMFILES(X86), and PROGRAMDATA environment variables are now passed through, unmasking system values necessary to locate resources such as a C compiler.
```
gaborbernat, line 1: Please make the env var names literals and remove the coma after windows

gaborbernat (changes requested):

docs/changelog/2382.feature.rst
jaraco: Done

gaborbernat (approved):

PR #2408, opened 2022-04-28
Migrate tox 3 documentation

gaborbernat:
This is the first pass.


jugmac00 (changes requested):
Thanks a lot!

I had a first pass - mostly from the GitHub ui, and only occasionally on the rendered docs - I will have another one when the PR is finished.

Apart from my inline comments, you certainly noticed that building the documentation is currently broken to an changed anchor in the pytest documentation.

Speaking of the pytest documentation, have you noticed that they restructured their documentation to follow https://diataxis.fr/ documentation framework? (not a technical framework, but a way to structure your docs).

Our documentation looks good, but maybe you have a time to have a look at diataxis whether you like it.

e.g. our faq is not really a faq as I have not seen too many people asking for customizing virtualenv creation - currently our faq is more like a place to put stuff which does not fit anywhere else.

In diataxis these would be all "how-tos". Our user-guide would be a tutorial, `configuration`, `tox cli interface` and `API` would go in a reference section, and so on... 

If we ever would plan to restructure the documentation now would be a great time.

Oh - and I think we need a What's new in tox version4 page - with all highlights.

docs/faq.rst
```diff
@@ -75,3 +119,126 @@ and the following ``tox.ini`` content:
...
+
+.. code-block:: ini
+
+
+   [testenv:py310]
+   commands =
+     - python -c 'import sys; sys.exit(1)'
+     python --version
+
+Customize virtual environment creation
```
jugmac00, line 138:
```suggestion
Customizing virtual environment creation
```

docs/faq.rst
```diff
@@ -75,3 +119,126 @@ and the following ``tox.ini`` content:
...
+
+   [testenv]
+   setenv =
+     VIRTUALENV_PIP==22.1
+     VIRTUALENV_SYSTEM_SITE_PACKAGES=true
+
+Consult the :pypi:`virtualenv` project for supported values (any CLI flag for virtualenv, in all upper case, prefixed
+by the ``VIRTUALENV_`` key).
+
+Build documentation with Sphinx
```
jugmac00, line 158:
```suggestion
Building documentation with Sphinx
```

docs/faq.rst
```diff
@@ -75,3 +119,126 @@ and the following ``tox.ini`` content:
...
+    basepython = python3.10
+    deps =
+      sphinx>=4
+    commands =
+      sphinx-build -d "{envtmpdir}{/}doctree" docs "{toxworkdir}{/}docs_out" --color -b html
+      python -c 'print(r"documentation available under file://{toxworkdir}{/}docs_out{/}index.html")'
+
+Note here we also require python 3.10, allowing us to use f-strings within the sphinx ``conf.py``.
+
+Build documentation with mkdocs
```
jugmac00, line 185:
```suggestion
Building documentation with mkdocs
```

jugmac00 (commented):
some more comments

docs/faq.rst
```diff
@@ -3,6 +3,50 @@ FAQ
...
+    PIP_EXTRA_INDEX_URL = {env:PIP_EXTRA_INDEX_URL:https://tox.wiki/pypi/simple-second}
+
+If the index server defined under ``PIP_INDEX_URL`` does not contain a package pip will attempt to resolve it also from
+the URI from ``PIP_EXTRA_INDEX_URL``.
+
+.. warning::
+
+  Using an extra PyPI index for installing private packages may cause security issues. For example, if ``package1`` is
+  registered with the default PyPI index, pip will install ``package1`` from the default PyPI index, not from the extra
+  one.
```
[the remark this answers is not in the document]
pradyunsg: Pretty much. There's no guarentees around which one pip would use the file from, if it sees two files with the same name from two package indexes.

jugmac00:
It looks like cpython is also adoption the Diátaxis framework for documentation:
https://discuss.python.org/t/adopting-the-diataxis-framework-for-python-documentation/15072

https://diataxis.fr/how-to-use-diataxis/
https://diataxis.fr/

I think it would be a great idea to also adopt it.

gaborbernat:
I'm not against it, but until we have availability for someone (mostly me in the first phase) to work on it, it doesn't really matter 😆

gaborbernat:
One thing I've learned part of the diatrixis framework is to not do big bang release. Considering that we'll merge this as is and do smaller follow-up PRs to make it diatrixis.

Issue #2358, opened 2022-02-16
tox4: regression plugins not listed with --version --version


ssbarnea:
While with tox 3 it was possible to list currently installed plugins, this feature vanished from tox4.

```
tox --version
```

That is confusing as it gives the impression that no plugins are installed.

ssbarnea:
I can confirm that this still happens with recently releases tox `4.0b2`.

PR #2413, opened 2022-05-06
Display registered plugins with --version

mxd4:
Ref #2358

Hi! First time contributing here. 
I tried to display the registered plugins when calling `tox --version`, by looking at how it was done on [master](https://github.com/tox-dev/tox/blob/master/src/tox/config/__init__.py#L359).

Looks like this works, however the formatter used by `argparse` removes `\n` characters, therefore the display is inline:
```
$ tox --version
4.0.0b2.dev15+gd1845907 from /Users/maximed/PycharmProjects/tox-rewrite/src/tox/__init__.py. registered plugins tox-mxd-0.0.0 at /Users/maximed/PycharmProjects/tox-mxd/tox_mxd.py
```

I didn't commit it but changing the formatter to `RawTextHelpFormatter` in [`parser.py`](https://github.com/tox-dev/tox/blob/rewrite/src/tox/config/cli/parser.py#L165) actually fixes it:
```
$ tox --version
4.0.0b2.dev15+gd1845907.d20220506 from /Users/maximed/PycharmProjects/tox-rewrite/src/tox/__init__.py.
registered plugins:
    tox-mxd-0.0.0 at /Users/maximed/PycharmProjects/tox-mxd/tox_mxd.py
```
Any idea how to improve the display without having to change the formatter of the parser? I'm still getting familiar with tox and argparse. And python. :innocent:

### DIFF
--- a/src/tox/config/cli/parser.py
+++ b/src/tox/config/cli/parser.py
@@ -162,7 +162,7 @@ class ToxParser(ArgumentParserWithEnvAndConfig):
             cmd,
             help=help_msg,
             aliases=aliases,
-            formatter_class=HelpFormatter,
+            formatter_class=argparse.RawTextHelpFormatter,
             file_config=self.file_config,
         )
         sub_parser.of_cmd = cmd  # mark it as parser for a sub-command
--- a/src/tox/config/cli/parser.py
+++ b/src/tox/config/cli/parser.py
@@ -99,13 +99,19 @@ class HelpFormatter(ArgumentDefaultsHelpFormatter):
         super().__init__(prog, max_help_position=30, width=240)
 
     def _get_help_string(self, action: Action) -> str | None:
-
         text: str = super()._get_help_string(action) or ""
         if hasattr(action, "default_source"):
             default = " (default: %(default)s)"
             if text.endswith(default):  # pragma: no branch
                 text = f"{text[: -len(default)]} (default: %(default)s -> from %(default_source)s)"
         return text
+
+    def add_raw_text(self, text: str | None) -> None:
+        def keep(content: str) -> str:
+            return content
+
+        if text is not SUPPRESS and text is not None:
+            self._add_item(keep, [text])
 
 
 ToxParserT = TypeVar("ToxParserT", bound="ToxParser")
@@ -162,7 +168,7 @@ class ToxParser(ArgumentParserWithEnvAndConfig):
             cmd,
             help=help_msg,
             aliases=aliases,
-            formatter_class=argparse.RawTextHelpFormatter,
+            formatter_class=HelpFormatter,
             file_config=self.file_config,
         )
         sub_parser.of_cmd = cmd  # mark it as parser for a sub-command
@@ -334,4 +340,5 @@ __all__ = (
     "DEFAULT_VERBOSITY",
     "Parsed",
     "ToxParser",
+    "HelpFormatter",
 )
--- a/src/tox/session/cmd/version_flag.py
+++ b/src/tox/session/cmd/version_flag.py
@@ -22,11 +22,11 @@ def tox_add_option(parser: ToxParser) -> None:
 
 
 def get_version_info() -> str:
-    out = [f"{version} from {Path(tox.__file__).absolute()} "]
+    out = [f"{version} from {Path(tox.__file__).absolute()}"]
     plugin_info = MANAGER.manager.list_plugin_distinfo()
     if plugin_info:
         out.append("registered plugins:")
-        for mod, egg_info in plugin_info:
-            source = getattr(mod, "__file__", repr(mod))
+        for module, egg_info in plugin_info:
+            source = getattr(module, "__file__", repr(module))
             out.append(f"    {egg_info.project_name}-{egg_info.version} at {source}")
     return "\n".join(out)
--- a/src/tox/session/cmd/version_flag.py
+++ b/src/tox/session/cmd/version_flag.py
@@ -18,17 +18,16 @@ def tox_add_option(parser: ToxParser) -> None:
     parser.add_argument(
         "--version",
         action="version",
-        version=f"{version} from {Path(tox.__file__).absolute()}.\n{get_registered_plugins()}",
+        version=f"{version} from {Path(tox.__file__).absolute()}\n{get_registered_plugins()}",
     )
 
 
 def get_registered_plugins():
-    out = ["registered plugins:"]
-    plugin_dist_info = MANAGER.manager.list_plugin_distinfo()
-    if plugin_dist_info:
-        for mod, egg_info in plugin_dist_info:
+    out = []
+    plugin_info = MANAGER.manager.list_plugin_distinfo()
+    if plugin_info:
+        out.append("registered plugins:")
+        for mod, egg_info in plugin_info:
             source = getattr(mod, "__file__", repr(mod))
             out.append(f"    {egg_info.project_name}-{egg_info.version} at {source}")
-    else:
-        out.append("None.")
     return "\n".join(out)
--- a/src/tox/session/cmd/version_flag.py
+++ b/src/tox/session/cmd/version_flag.py
@@ -5,25 +5,24 @@ from __future__ import annotations
 
 from pathlib import Path
 
+import tox
 from tox.config.cli.parser import ToxParser
 from tox.plugin import impl
 from tox.plugin.manager import MANAGER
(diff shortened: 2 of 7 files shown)


src/tox/session/cmd/version_flag.py
```diff
@@ -17,5 +18,17 @@ def tox_add_option(parser: ToxParser) -> None:
...
+
+def get_registered_plugins():
+    out = ["registered plugins:"]
+    plugin_dist_info = MANAGER.manager.list_plugin_distinfo()
+    if plugin_dist_info:
+        for mod, egg_info in plugin_dist_info:
+            source = getattr(mod, "__file__", repr(mod))
+            out.append(f"    {egg_info.project_name}-{egg_info.version} at {source}")
+    else:
+        out.append("None.")
```
jugmac00, line 33: tox v3 does not show anything when no plugins are installed - I think we should keep this behavior. ie only show plugin information when a plugin is installed.

jugmac00 (changes requested):
Hey @mxd4,
Thanks a lot for your contribution!

A couple of things of test runs are failing. It is always a good idea to run `tox` or in this case `tox4` locally before committing, pushing and creating a PR. Though in this case there might be some flaky tests which I need to check closer. But at very least the `tox4 -e fix` env is failing. We use `flake8-spellcheck` which requires all unknown words need to be whitelisted in the file `whitelist.txt`.

You'd also need to create a changelog entry, so we can publish the good news about the bug fix  in our changelog.

Also, you need to write a test case which covers your added / modified lines.

P.S.: It is pretty late here, so I will have a closer look on the PR when you have fixed the current issues.

mxd4:
Hi @jugmac00 , thanks a lot for your guidance!
I've fixed the spellcheck, added a changelog and a test.
I'd like a second test here to display plugins, but I can't manage to register a dummy plugin when running `tox_project`.

jugmac00:
I pretty packed this weekend, so I have not the capacity to dig deeper now.

But I'd have a look at look at how this is tested in the tox3 codebase.

In order to know where to look, I created a virtualenv, installed tox, installed a tox plugin, and ran `tox --version`:

```
❯ tox --version
3.25.1.dev3+g7818b3d5 imported from /home/jugmac00/Projects/tox/src/tox/__init__.py
registered plugins:
    tox-gh-actions-2.9.1 at /home/jugmac00/Projects/tox/venv/lib/python3.8/site-packages/tox_gh_actions/plugin.py
```

Then I searched for "registered plugins" in the test files of tox3:

https://github.com/tox-dev/tox/blob/7818b3d55fadef6fc881e66bc58626e9f6832c34/tests/unit/config/test_config.py#L3323-L3366

Can you have a look whether this works for you?

Also, the type checker shows an error (see failed tests runs):
```
type: commands[0]> mypy src/tox
src/tox/session/cmd/version_flag.py:21: error: Call to untyped function "get_registered_plugins" in typed context  [no-untyped-call]
src/tox/session/cmd/version_flag.py:25: error: Function is missing a return type annotation  [no-untyped-def]
Found 2 errors in 1 file (checked 101 source files)
```

The fix would be possible something like `def get_registered_plugins():` -> `def get_registered_plugins() -> str:`

tests/test_version.py
```diff
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+from tox import __version__
+from tox.pytest import ToxProjectCreator
 
-def test_version() -> None:
```
gaborbernat, line 4: You should not remove this test.

tests/test_version.py
```diff
@@ -1,7 +1,10 @@
...
-    from tox import __version__
 
-    assert __version__
```
gaborbernat, line 7: You'll need to add a test that validates the output with plugins. Check some of our tests that validate the plugin system to find out how.

gaborbernat (changes requested):
> I didn't commit it but changing the formatter to `RawTextHelpFormatter` in [`parser.py`](https://github.com/tox-dev/tox/blob/rewrite/src/tox/config/cli/parser.py#L165) actually fixes it:

I don't think can be done otherwise.

src/tox/session/cmd/version_flag.py
```diff
@@ -5,17 +5,28 @@
...
 from tox.config.cli.parser import ToxParser
 from tox.plugin import impl
+from tox.plugin.manager import MANAGER
+from tox.version import version
 
 
 @impl
 def tox_add_option(parser: ToxParser) -> None:
-    import tox
-    from tox.version import version
```
mxd4, line 15: Was there a reason those where imported locally and not in the import section?

tests/test_version.py
```diff
@@ -1,7 +1,42 @@
...
+        class MockModule:
+            __file__ = "dummy-path"
+
+        class MockEggInfo:
+            project_name = "dummy-project"
+            version = "1.0"
+
+        return [(MockModule, MockEggInfo)]
+
+    monkeypatch.setattr(MANAGER.manager, "list_plugin_distinfo", dummy_plugin)
```
mxd4, line 34: @gaborbernat I dug deep into the different plugin tests but I was not able to create a `tox_project` with registered plugins that would get displayed in `--version`. Maybe I'm missing something with how to properly use `tox_project` and `register_inline_plugin` but the manager could not find them.

Instead I looked at how it was done in `tox3` like @jugmac00 suggested, and used `MonkeyPatch` to get a plugin as a response from the plugin manager.

mxd4:
> > I didn't commit it but changing the formatter to `RawTextHelpFormatter` in [`parser.py`](https://github.com/tox-dev/tox/blob/rewrite/src/tox/config/cli/parser.py#L165) actually fixes it:
> 
> I don't think can be done otherwise.

Is it acceptable to have the plugins displayed inline? or should I go through with it and change the formatter?
Looks like it doesn't break any test but I'm afraid it could mess up some command outputs.

src/tox/session/cmd/version_flag.py
gaborbernat: Perhaps circular import issues 🤔

tests/test_version.py
```diff
@@ -1,7 +1,45 @@
 from __future__ import annotations
 
+from _pytest.monkeypatch import MonkeyPatch
```
gaborbernat, line 3: Please import without the _

tests/test_version.py
```diff
@@ -1,7 +1,45 @@
...
+
+    def dummy_plugin():
+        class MockModule:
+            __file__ = "dummy-path"
+
+        class MockEggInfo:
+            project_name = "dummy-project"
+            version = "1.0"
+
+        return [(MockModule, MockEggInfo)]
```
gaborbernat, line 35: Use mock objects instead

gaborbernat (changes requested):
> Is it acceptable to have the plugins displayed inline? or should I go through with it and change the formatted?

You should go through with it. You also need to fix the type checker errors.

src/tox/session/cmd/version_flag.py
mxd4: I was not familiar with those issues, I ran [pycycle](https://pypi.org/project/pycycle/) and looks like we're fine with them in the import section.

tests/test_version.py
```diff
@@ -1,7 +1,42 @@
...
+    from tox import __version__
+
+    mock_module = mock.Mock(__file__="dummy-path")
+
+    mock_egg_info = mock.Mock(
+        project_name="dummy-project",
+        version="1.0",
+    )
+
+    monkeypatch.setattr(MANAGER.manager, "list_plugin_distinfo", lambda: [(mock_module, mock_egg_info)])
```
gaborbernat, line 34: use mock.patch.object instead 👍

mxd4:
Rebased on top of `rewrite` branch to fix tests

src/tox/config/cli/parser.py
```diff
@@ -162,7 +162,7 @@ def add_command(
             cmd,
             help=help_msg,
             aliases=aliases,
-            formatter_class=HelpFormatter,
+            formatter_class=argparse.RawTextHelpFormatter,
```
gaborbernat, line 165: You cannot do this, you're removing the entire HelpFormatter functionality :thinking:

mxd4: Ah yes, my bad, I should have edited `HelpFormatter` instead. I see you just did, thank you! 😄

gaborbernat (approved):

PR #2419, opened 2022-05-11
fix typo

jugmac00:


jugmac00:
typo fix -> merging without approval

PR #2448, opened 2022-06-28
Unify badges style

DeadNews:
Bringing badges to a unified style.


jugmac00 (approved):
Thanks! Will merge after successful CI - I do not think a changelog entry is necessary.

PR #2456, opened 2022-07-05
Do no longer recommend pipx-in-pipx

jugmac00:
... as it is no longer actively maintained.

We can trust our users to be able to install pipx in a different way.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -10,7 +10,6 @@ affecting other parts of the system.
 
 .. code-block:: bash
 
-    python -m pip install pipx-in-pipx --user
     pipx install tox
     tox --help
 


gaborbernat (changes requested):
I never had any issues with it and I don't see any mention of not being maintained. 🤔 I'd rather adopt its maintainership rather than recommend something else 😆

jugmac00:
Today I was asked how to install tox and I wanted to direct the op to our documentation but I did not feel comfortable doing so because of pipx-in-pipx.

Whatever you call its current maintenance state, I do think it is enough from our side to recommend pipx - they have their own installation documentation.

gaborbernat:
> Today I was asked how to install tox and I wanted to direct the op to our documentation but I did not feel comfortable doing so because of pipx-in-pipx.

Guess we disagree on recommending `pipx-in-pipx`. 🤔 

> Whatever you call its current maintenance state, I do think it is enough from our side to recommend pipx - they have their own installation documentation.

Perhaps, but I found people found it helpful to not have to read another documentation on how to get quickly started 🤔

gaborbernat:
I think we had different experiences. `pipx-in-pipx` worked flawlessly for me on macOS, Arch Linux or Windows. I don't use Ubuntu personally so 🤷

Issue #2462, opened 2022-07-21
Use tomllib from the standard library on Python 3.11+


hroncok:
Hey,
would you accept a pull request that uses tomllib rom the standard library on Python 3.11+ and fallbacks to tomli, instead of toml?

hroncok:
I was testing the changes, so I opened a draft PR without a changelog entry to present what change I meant. If acceptable, I'll add the entry. https://github.com/tox-dev/tox/pull/2463

jugmac00:
Looks like tomli never had support for Python 2.7. I am not sure what are the plans about that in the future. `tox4` will require Python 3.7 as a minimum, but I don't think it makes a lot of sense to drop support for Python 2.7 for tox 3 anymore.

hroncok:
I have not thought about that. Will fallback to toml. Not today though.

hroncok:
Done.

PR #2463, opened 2022-07-21
Use tomllib from the standard library on Python 3.11+

hroncok:
Fallback to tomli, which has the same API.

Fixes https://github.com/tox-dev/tox/issues/2462

## Contribution checklist:

- [ ] added/updated test(s)
    ```rst
    ```

### DIFF
--- a/docs/changelog/2463.feature.rst
+++ b/docs/changelog/2463.feature.rst
@@ -1,1 +1,1 @@
-Use ``tomllib`` on Python 3.11 or later and ``tomli`` instead of ``toml`` library on lower versions - by :user:`gaborbernat`.
+Use ``tomllib`` on Python 3.11 or later and ``tomli`` instead of ``toml`` library on lower versions - by :user:`hroncok`.
--- a/docs/changelog/2463.feature.rst
+++ b/docs/changelog/2463.feature.rst
@@ -1,0 +1,1 @@
+Use ``tomllib`` on Python 3.11 or later and ``tomli`` instead of ``toml`` library on lower versions - by :user:`gaborbernat`.
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ install_requires =
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
-    toml;python_version<"3.11"
+    tomli>=2.0.1;python_version<"3.11"
 python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
 
 [options.packages.find]
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ install_requires =
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
-    toml>=0.9.4;python_version<="3.5"
+    toml>=0.10.2;python_version<="3.5"
     tomli;python_version>="3.6" and python_version<"3.11"
 python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ install_requires =
     pluggy>=0.12.0
     py>=1.4.17
     six>=1.14.0 # required when virtualenv>=20
-    toml>=0.9.4
+    tomli;python_version<"3.11"
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,8 +42,8 @@ install_requires =
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
-    toml>=0.10.2;python_version<="3.5"
-    tomli;python_version>="3.6" and python_version<"3.11"
+    toml>=0.10.2;python_version<="3.6"
+    tomli>=2.0.1;python_version>="3.7" and python_version<"3.11"
 python_requires = >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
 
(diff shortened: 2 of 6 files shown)


jugmac00:
Looks good to me!

But neither tox nor gh actions run Python 3.11.

@gaborbernat What do you think? Are you ok with the changes? This would mean we need to run CI with Python 3.11 beta.

gaborbernat:
> But neither tox nor gh actions run Python 3.11.

Yeah, we need to change that before we merge this. I'll add 3.11 to virtualenv today and then move to tox.

gaborbernat (approved):

hroncok:
Is the Python 2 compatibility no longer a concern?

jugmac00:
> Is the Python 2 compatibility no longer a concern?

The information is split between the issue and the pr.

@gaborbernat The Python 2 comp was discussed at https://github.com/tox-dev/tox/issues/2462#issuecomment-1192926874

gaborbernat:
Seems we'll need to fallback to toml on python 2.

hroncok:
Yes. See my *fixup: Fallback to toml, not tomli, for Python 2 support* commit which can still be seen here at github. It seems to have been erased away.

gaborbernat:
Because you've fallen back to toml on Python 3.11 or less and not just Python 2. 🤔

gaborbernat:
> So you'd prefer a three-way split? I can do that, I just considered it a bit redundant.

Yes.

src/tox/config/__init__.py
```diff
@@ -309,8 +313,8 @@ def parseconfig(args, plugins=()):
 
 
 def get_py_project_toml(path):
-    with io.open(str(path), mode="rb") as file_handler:
-        config_data = tomllib.load(file_handler)
+    with io.open(str(path), mode=toml_mode, encoding=toml_encoding) as file_handler:
```
gaborbernat: Do we need io.open here, can't we just use io?

setup.cfg
```diff
@@ -42,7 +42,7 @@ install_requires =
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
-    tomli;python_version<"3.11"
+    toml;python_version<"3.11"
```
gaborbernat: This should be version lower constrained, and we should use tomli on Python 3.

gaborbernat (changes requested):

src/tox/config/__init__.py
hroncok: Just use io? Do you mean just use open? That's missing some arguments needed on Python 2. IIRC encoding.

hroncok: Exactly encoding.

setup.cfg
```diff
@@ -39,10 +39,11 @@ install_requires =
     pluggy>=0.12.0
     py>=1.4.17
     six>=1.14.0 # required when virtualenv>=20
-    toml>=0.9.4
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
+    toml;python_version=="2.7"
+    tomli>=2.0.1;python_version>="3.5" and python_version<"3.11"
```
hroncok, line 46:
```suggestion
    tomli;python_version>="3.5" and python_version<"3.11"
```

setup.cfg
```diff
@@ -39,10 +39,11 @@ install_requires =
     pluggy>=0.12.0
     py>=1.4.17
     six>=1.14.0 # required when virtualenv>=20
-    toml>=0.9.4
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
+    toml;python_version=="2.7"
```
hroncok, line 45:
```suggestion
    toml>=0.9.4;python_version=="2.7"
```

hroncok:
Looks like 3.5 will also need toml. Will update this later today.

hroncok:
This seems to work. At least on cpythons.

Let me know if you prefer to use toml also on 3.6 and pin tomli to 2+.

Let me know if you want me to rebase or squash somehow.

docs/changelog/2463.feature.rst
```diff
@@ -0,0 +1 @@
+Use ``tomllib`` on Python 3.11 or later and ``tomli`` instead of ``toml`` library on lower versions - by :user:`gaborbernat`.
```
hroncok, line 1:
```suggestion
Use ``tomllib`` on Python 3.11 or later and ``tomli`` instead of ``toml`` library on lower versions - by :user:`hroncok`.
```

setup.cfg
```diff
@@ -39,10 +39,11 @@ install_requires =
     pluggy>=0.12.0
     py>=1.4.17
     six>=1.14.0 # required when virtualenv>=20
-    toml>=0.9.4
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
+    toml>=0.9.4;python_version<="3.5"
+    tomli;python_version>="3.6" and python_version<"3.11"
```
gaborbernat, line 46: Lower pin for this please

gaborbernat (changes requested):

setup.cfg
```diff
@@ -39,10 +39,11 @@ install_requires =
     pluggy>=0.12.0
     py>=1.4.17
     six>=1.14.0 # required when virtualenv>=20
-    toml>=0.9.4
     virtualenv!=20.0.0,!=20.0.1,!=20.0.2,!=20.0.3,!=20.0.4,!=20.0.5,!=20.0.6,!=20.0.7,>=16.0.0
     colorama>=0.4.1 ;platform_system=="Windows"
     importlib-metadata>=0.12;python_version<"3.8"
+    toml>=0.9.4;python_version<="3.5"
```
gaborbernat, line 45: 0.10.2 here per https://pypi.org/project/toml/

setup.cfg
gaborbernat: 2.0.1

setup.cfg
hroncok: IIRC version 2 does not support Python 3.6 so that requires using toml on 3.6. Is that what you prefer?

setup.cfg
hroncok: IIRC version 2 does not support Python 3.6 so that requires using toml on 3.6. Is that what you prefer?

setup.cfg
hroncok: I don't understand this request. It was already pinned to a different version and I kept it the same. I can change it, but I wonder why.

setup.cfg
hroncok: Confirmed. tomli 1.2.3  is the latest released version that supported Python 3.6

setup.cfg
hroncok: Done.

setup.cfg
hroncok: Done.

hroncok:
> Let me know if you prefer to use toml also on 3.6 and pin tomli to 2+.

I've done that now. Feel free to revert the commit.

gaborbernat (approved):

PR #2466, opened 2022-07-26
Check 3.11 support v4

gaborbernat:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.2.0
+    rev: v4.3.0
     hooks:
       - id: check-ast
       - id: check-builtin-literals
@@ -12,12 +12,12 @@ repos:
       - id: end-of-file-fixer
       - id: trailing-whitespace
   - repo: https://github.com/asottile/add-trailing-comma
-    rev: v2.2.2
+    rev: v2.2.3
     hooks:
       - id: add-trailing-comma
         args: [--py36-plus]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.32.0
+    rev: v2.37.2
     hooks:
       - id: pyupgrade
         args: ["--py37-plus"]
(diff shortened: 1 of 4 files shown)


asottile:
wouldn't be a new cpython release if cython weren't broken eh?

gaborbernat:
> wouldn't be a new cpython release if cython weren't broken eh?

Well cython is very brittle to c-extension modifications...

cclauss:
3.11.0-beta.5 should be available now.

.github/workflows/check.yml
```diff
@@ -17,40 +17,40 @@ jobs:
       fail-fast: false
       matrix:
         py:
+          - "3.11.0-beta.4"
```
cclauss, line 20:
```suggestion
          - "3.11.0-beta.5"
```

fschulze:
Why is tox installing devpi-server? I did a GitHub search for it in the repository, but it didn’t come up with anything.

cclauss:
https://github.com/tox-dev/devpi-process is written but the tox-dev team and described here https://devpi.net/docs/devpi/devpi/stable/+d/index.html

gaborbernat:
Uses devpi in the test suite to test self provisioning.

.github/workflows/check.yml
```diff
@@ -17,40 +17,40 @@ jobs:
       fail-fast: false
       matrix:
         py:
+          - "3.11.0-beta.5"
```
cclauss, line 20:
```suggestion
          - "3.11-dev"
```

gaborbernat: No thanks.

cclauss: Why not [rc1](https://www.python.org/download/pre-releases)?

gaborbernat: Until aiofiles makes a new release doesn't matter. You were suggesting dev though not rc1.

cclauss: Understood.  `dev == rc1` until there is an `rc2` ;-)

fschulze:
I found the tests. For now I would suggest to pin devpi-server to <6.3.0 when running with Python 3.11. These should work in pyproject.toml:
```
  "devpi-server<6.3;python_version=='3.11'",
  "devpi-server>=6.6;python_version<'3.11'",
```
There should be no changes in devpi-server between 6.2 and 6.6 which would affect the tests besides usage of asyncio.

PR #2487, opened 2022-09-07
Fix SetuptoolsDeprecationWarning about using --global-option

### LINKED ISSUE #2478: Warning about `"--global-option": ["--formats=gztar"]` emitted when using setuptools 64.0.1
When using tox to build or install a setuptools-based package, if the tox-created virtualenv contains setuptools v64.0.1, then the following warning is emitted:

```
/Users/jwodder/work/pyproj/linesep/.tox/.package/lib/python3.9/site-packages/setuptools/build_meta.py:306: SetuptoolsDeprecationWarning:
            The arguments ['--formats=gztar'] were given via `--global-option`.
            Please use `--build-option` instead,
            `--global-option` is reserved to flags like `--verbose` or `--quiet`.
```

This appears to be due to [this line of code](https://github.com/tox-dev/tox/blob/d98f589197f033fa9846918f309bf66a645f905b/src/tox/helper/build_isolated.py#L41) in tox's source.

# To Reproduce

1. Clone, say, https://github.com/jwodder/linesep
2. Ensure virtualenv uses the latest version of setuptools by running `virtualenv --upgrade-embed-wheels`
3. Run `tox`

adamchainz:
Fixes #2478.

## Contribution checklist:

- [x] added/updated test(s)
      in message body
    ```rst
    ```

### DIFF
--- a/docs/changelog/2478.bugfix.rst
+++ b/docs/changelog/2478.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``SetuptoolsDeprecationWarning`` about using ``--global-option`` -- by :user:`adamchainz`
--- a/src/tox/helper/build_isolated.py
+++ b/src/tox/helper/build_isolated.py
@@ -38,5 +38,5 @@ _ensure_module_in_paths(backend, backend_paths)
 if backend_obj:
     backend = getattr(backend, backend_obj)
 
-basename = backend.build_sdist(dist_folder, {"--global-option": ["--formats=gztar"]})
+basename = backend.build_sdist(dist_folder, {"--build-option": ["--formats=gztar"]})
 print(basename)
--- a/tests/unit/package/builder/test_package_builder_isolated.py
+++ b/tests/unit/package/builder/test_package_builder_isolated.py
@@ -25,6 +25,7 @@ def test_verbose_isolated_build(initproj, mock_venv, cmd):
         },
     )
     result = cmd("--sdistonly", "-v", "-v", "-v", "-e", "py")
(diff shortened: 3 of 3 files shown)


asottile:
what versions of setuptools does the new argument require ?

gaborbernat (approved):

adamchainz:
I tested down to setuptools 50 and got no error.

But looking at the setuptools code, it seems `--build-option` was actually introduced in setuptools 63: https://github.com/pypa/setuptools/pull/3380 . So perhaps when I used older versions, it was just being ignored.

asottile:
@gaborbernat I don't think this could go out as is given the above

gaborbernat:
I guess the question here if older setuptools just ignored this flag or actually used it 🤔

asottile:
I believe on windows the flag is important -- otherwise it'll default to zip ?

gaborbernat:
Would that cause a problem for pip?

asottile:
it shouldn't be a problem for pip -- but I thought tox tries to read the tgz -- maybe I'm misremembering how this works -- not at a computer to check

gaborbernat:
tox does not read the file 🤔

asottile:
this is what I was thinking of but looks like it already supports zip: https://github.com/tox-dev/tox/blob/d98f589197f033fa9846918f309bf66a645f905b/src/tox/package/local.py#L52

we can probably drop the option entirely then

adamchainz:
That seems reasonable to me too, done in #2497.

PR #2489, opened 2022-09-07
Expand installation instruction for common linux distros, macos

alago1:
The instructions for installing tox through linux/macos package managers mention potential name conflicts with other projects but lacks in examples. I also could not find info anywhere else in the docs that expanded upon this.

I looked through the repositories for the most popular linux distro's package managers and homebrew for macos and added a section with the installation commands. Given there's some variation in project's name between `tox`, `python-tox`, and `python3-tox`, I think this section helps to add some clarity to these specific installation instructions.

## Contribution checklist:

- [ ] added/updated test(s)
      in message body
- [X] added news fragment in [changelog folder](../tree/master/docs/changelog)
    ```rst
    ```

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -52,6 +52,39 @@ so that you can do changes and submit patches.
 [Linux/macOS] Install via your package manager
 ----------------------------------------------
 
-You can also find tox packaged for many Linux distributions and Homebrew for macOs - usually under the name of **python-tox** or simply **tox**. Be aware though that there also other projects under the same name (most prominently a `secure chat client <https://tox.chat/>`_ with no affiliation to this project), so make sure you install the correct package.
+You can also find tox packaged for many Linux distributions and Homebrew for macOs.
+
+Debian, Ubuntu
+
+.. code-block:: shell
+
+    $ apt-get install tox
+
+Fedora, CentOS, Red Hat
+
+.. code-block:: shell
+
+    $ yum -y install python-tox
+
+Arch Linux
+
+.. code-block:: shell
+
+    $ pacman -S python-tox
+
+openSUSE
+
+.. code-block:: shell
+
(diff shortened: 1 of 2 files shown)


gaborbernat:
Duplicate of https://github.com/tox-dev/tox/pull/2476#issuecomment-1221596894.

alago1:
Hey @gaborbernat , would you be able to provide some feedback as to why you think this is a duplicate?
The linked PR attempted to add instructions related to installing Python whereas this PR aids in installing the tox package itself.
Within the existing installation instructions there's commands to install with pip or from a clone but the linux/macos package manager installations lack much detail which is why I though this change made sense.

docs/install.rst
```diff
@@ -52,6 +52,39 @@ so that you can do changes and submit patches.
 [Linux/macOS] Install via your package manager
```
gaborbernat, line 52: I'd rather remove this section altogether. Installing tox not in a virtual environment via pipx is always causing just problems because OS distro repackaging always messes up our packaging.

asottile: +1

jugmac00: While I agree that packaged versions of tox, which are usually outdated and missing bugfixes, will cause some problems, both for the user and for us when the users complain about long fixed bugs or even things which are only broken in the packaged version (looking at you, bionic's version of tox), I do think many people will still use the packaged versions, maybe they even have no choice, e.g. in an enterprise environment.

I am slightly in favor to accept this PR, as it will help especially beginners to get started.

Though, I would like to see some words of warning that the preferred way for installation would be pip/pipx, and that packaged versions of tox might be outdated and are not directly supported by us.

Can we agree on one solution to move forward this pr? Though it looks like a clear 2-1 :-)

PR #2501, opened 2022-09-14
Remove read-only files upon cleanup

### LINKED ISSUE #2498: Tox cannot remove readonly files from tmp directory
When there are some read-only files left in tmp directory, they are not cleared upon next execution.

Cleaning is done in ensure_empty_dir.

Changing it to the following code fixes the problem:
```python
import os, stat
import shutil

from tox import reporter


def ensure_empty_dir(path):
    if path.check():
        reporter.info("  removing {}".format(path))
        shutil.rmtree(str(path), onerror=remove_readonly)  # originally without onerror, just with ignore_errors
        path.ensure(dir=1)

def remove_readonly(func, path, _):
    "Clear the readonly bit and reattempt the removal"
    os.chmod(path, stat.S_IWRITE)
    func(path)
```

In our case those read-only files come from externally executed commands and they affect subsequent tests execution.
As a workaround we'd consider cleaning them ourselves.

Reproducible: always
Platform: win32 (Windows 10), Python 3.9
Tox version: tox==3.25.1

Please let me know if you need more information.
I can try providing self-reproducible setup, but might take a while.

robgom:
Delete read-only files in ``ensure_empty_dir``.

fixes #2498

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -107,6 +107,7 @@ Pierre-Jean Campigotto
 Pierre-Luc Tessier Gagné
 Prakhar Gurunani
 Rahul Bangar
+Robert Gomulka
 Ronald Evers
 Ronny Pfannschmidt
 Ryuichi Ohori
--- a/docs/changelog/2498.bugfix.rst
+++ b/docs/changelog/2498.bugfix.rst
@@ -1,0 +1,1 @@
+Remove read-only files in ``ensure_empty_dir``.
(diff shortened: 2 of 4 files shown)


robgom:
Verified manually in my environment - without the fix my tests failed, with fix applied they pass.

asottile:
needs a test demonstrating the feature

robgom:
> needs a test demonstrating the feature

I have added a test proof-of-concept. It's failing, as I don't understand all the relationships in test code (mocks, fixtures).
If someone is able to polish it, great, if not - it'll have to wait until I learn it.

robgom:
Managed to add a test.

gaborbernat (approved):

Issue #2506, opened 2022-09-27
Ability to specify more than one dependency in scoped "deps" config section


robblovell:
In a project, to run the mypy static analyzer, I need to install both mypy and types-requests. While I could specify a separate requirements.txt for this, I would rather just put it as a list in the deps section:
```
[tox]
envlist   =
    tests
    dev
    lint
    compile
skipsdist = true
allowlist_externals =
    tox-echo.sh

[testenv]
deps =
    dev: -r media_analysis/requirements-m1.txt
    tests: -r media_analysis/requirements-m1.txt
    lint: pyflakes
    compile: mypy, types-requests

commands =
    dev: ./tox-echo.sh
    tests: python -m pytest media_analysis/
    lint: pyflakes media_analysis/
    compile: dmypy run --timeout 5 -- media_analysis
```
`tox -e compile` yields an error with installing of the dependencies `mypy` and `types-requests`:
```
ERROR: invocation failed (exit code 1), logfile: .../.tox/compile/log/compile-1.log
=================================================================================================== log start ====================================================================================================
ERROR: Invalid requirement: 'mypy, types-requests'

==================================================================================================== log end =====================================================================================================
ERROR: could not install deps [mypy, types-requests]; 
v = InvocationError(".../.tox/compile/bin/python -m pip install 'mypy, types-requests'", 1)
____________________________________________________________________________________________________ summary _____________________________________________________________________________________________________
ERROR: compile: could not install deps [mypy, types-requests]; 
v = InvocationError(".../.tox/compile/bin/python -m pip install 'mypy, types-requests'", 1)

```
It appears the arguments are not split at the comma and passed as a full string to pip.
```
python -m pip install 'mypy, types-requests'
```
instead of
```
python -m pip install 'mypy' 'types-requests'
```
Note that specifying `compile: mypy, types-requests` with a space (`compile: mypy types-requests`) instead makes no difference. `compile: mypy types-requests` would also be more consistent with the way pip works and might be the right implementation if this feature is included.

My workaround is to use a separate requirements.txt or to specify the extra dependency before the scoped ones:
```
...
[testenv]
deps =
    types-requests
    dev: -r media_analysis/requirements-m1.txt
    tests: -r media_analysis/requirements-m1.txt
    lint: pyflakes
    compile: mypy
...
```
or
```
...
[testenv]
deps =
    dev: -r media_analysis/requirements-m1.txt
    tests: -r media_analysis/requirements-m1.txt
    lint: pyflakes
[...]

PR #2507, opened 2022-09-27
feature: split and install dependencies if a comma-space is in the deps line.

robblovell:
## Description

Implementation of feature request: #2506 

There are two failed tests, one failure I believe is due to this change, the other I believe to be pre-existing.

I have modified the venv.py file's get_resolved_dependencies() function to split the dependency listed if a comma-space exists in the string. I found that comma's only or spaces only failed around 7 tests and given my unfamiliarity with the codebase, I opted for a more stringent syntax. 

I believe that a split on a 'comma only' would be more expected. If someone more familiar with the code might understand why a split on a comma only is problematic and could fix the additional tests, then it should probably be implemented with a split on a comma only.

Giving a list on separate lines is also not supported, but additional work could make that possible.

I have added `import copy` and used copy.copy to clone the dependency and I am unsure if this is problematic for python 2.7.

I am unsure of backward compatibility issues this might introduce given my unfamiliarity with Tox in general and the code base. Please feel free to critique what I have proposed, alternate solutions would be welcome.

## Todo's

TODO: Fix 2 failing tests.
TODO: Multi-line dependency spec
TODO: comma only dependency spec.

## Contribution checklist:

- [ ] added/updated test(s)
      in message body
    ```rst
    ```

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -108,6 +108,7 @@ Pierre-Luc Tessier Gagné
 Prakhar Gurunani
 Rahul Bangar
 Robert Gomulka
+Robb Lovell
 Ronald Evers
 Ronny Pfannschmidt
 Ryuichi Ohori
--- a/docs/changelog/2506.feature.rst
+++ b/docs/changelog/2506.feature.rst
@@ -1,0 +1,1 @@
+Split and install dependencies if a comma-space is in the deps line.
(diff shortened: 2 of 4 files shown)


gaborbernat (changes requested):
I don't thinkl this is backwards compatible.

jugmac00:
> I don't thinkl this is backwards compatible.

I am a bit confused why you did not close this pr - is there anything the op could to make this backwards compatible or what do you expect to happen now?

If there is nothing to do, let's close this pr.

PR #2529, opened 2022-11-03
Fix PYTHONPATH passed to envreport / "pip freeze"

### LINKED ISSUE #2528: PYTHONPATH leaks into "envreport"
Pretty trivial bug.
Even though tox discards PYTHONPATH for `pip install`, it does not discard it for `pip freeze`, causing garbage output if PYTHONPATH has any extra packages (i.e. "pip freeze" will see packages from PYTHONPATH, which it should not do).

https://github.com/tox-dev/tox/blob/3.27.0/src/tox/venv.py#L843

Fix is trivial - fetch env and apply `ensure_pip_os_environ_ok` the same way it is done for `pip install`, pass env to pcall. I will make a PR.

f3flight:
Closes #2528
This fix uses the same approach as used for "pip install" (https://github.com/tox-dev/tox/blob/3.27.0/src/tox/venv.py#L432)

### DIFF
--- a/tests/unit/test_venv.py
+++ b/tests/unit/test_venv.py
@@ -1243,4 +1243,8 @@ def test_runenvreport_pythonpath_discarded(newmocksession, mocker):
     mock_os_environ.return_value = dict(PYTHONPATH="/some/path/")
     mock_pcall = mocker.patch.object(venv, "_pcall")
     tox_runenvreport(venv, None)
-    assert "PYTHONPATH" not in mock_pcall.mock_calls[0].kwargs["env"]
+    try:
+        env = mock_pcall.mock_calls[0].kwargs["env"]
+    except TypeError:  # older pytest (python 3.7 and below)
+        env = mock_pcall.mock_calls[0][2]["env"]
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
Needs test.

tests/unit/test_venv.py
```diff
@@ -1233,3 +1235,15 @@ def test_path_change(tmpdir, mocksession, newconfig, monkeypatch):
         path = x.env["PATH"]
         assert os.environ["PATH"] in path
         assert path.endswith(str(venv.envconfig.config.toxinidir) + "/bin")
+
+
+@pytest.mark.parametrize("pythonpath", (None, "blah"))
```
gaborbernat, line 1240: if we don't add more parameters there's no need for parametrize here; also use mocker.patch from pytest-mock instead imorting unittest.mock

gaborbernat (changes requested):

f3flight:
@gaborbernat would you be able to have another look?

gaborbernat (approved):

PR #2539, opened 2022-11-19
Invoke pytest indirectly

mtelka:
This is to support [tox-current-env](https://pypi.org/project/tox-current-env/).

### DIFF
--- a/docs/example/pytest.rst
+++ b/docs/example/pytest.rst
@@ -24,7 +24,7 @@ and the following ``tox.ini`` content:
 
     [testenv]
     deps = pytest               # PYPI package providing pytest
-    commands = pytest {posargs} # substitute with tox' positional arguments
+    commands = python -m pytest {posargs} # substitute with tox' positional arguments
 
 you can now invoke ``tox`` in the directory where your ``tox.ini`` resides.
 ``tox`` will sdist-package your project, create two virtualenv environments
@@ -53,7 +53,7 @@ and the following ``tox.ini`` content:
     changedir = tests
     deps = pytest
     # change pytest tempdir and add posargs from command line
-    commands = pytest --basetemp="{envtmpdir}" {posargs}
+    commands = python -m pytest --basetemp="{envtmpdir}" {posargs}
 
 you can invoke ``tox`` in the directory where your ``tox.ini`` resides.
 Differently than in the previous example the ``pytest`` command
@@ -75,7 +75,7 @@ to make ``tox`` use this feature:
     deps = pytest-xdist
     changedir = tests
     # use three sub processes
-    commands = pytest --basetemp="{envtmpdir}"  \
+    commands = python -m pytest --basetemp="{envtmpdir}"  \
                       --confcutdir=..         \
                       -n 3                    \
                       {posargs}


docs/example/pytest.rst
```diff
@@ -24,7 +24,7 @@ and the following ``tox.ini`` content:
 
     [testenv]
     deps = pytest               # PYPI package providing pytest
-    commands = pytest {posargs} # substitute with tox' positional arguments
+    commands = python -m pytest {posargs} # substitute with tox' positional arguments
```
asottile, line 27: if this doesn't work with (unsupported) tox-current-env that feels to me a bug on their part.

asottile (changes requested):

docs/example/pytest.rst
mtelka: Yes, this could be saw as a bug on their side, or simply design choice (the latter seems to be true).  BTW, many projects already does this, for example [mypy](https://github.com/python/mypy/blob/master/tox.ini#L19), [wheel](https://github.com/pypa/wheel/blob/main/tox.ini#L13), or [Pillow](https://github.com/python-pillow/Pillow/blob/main/tox.ini#L19).

asottile: their project is unsupported and there's no improvement to doing it this way in tox. if current-env doesn't adjust PATH properly I don't think we should make our docs worse to accommodate them

asottile: btw those are done that way to pass -W, not for any other reason (and that can slash should be done with PYTHONWARNINGS or the pytest setting)

mtelka: Could you please elaborate how this would make the documentation worse?

The problem is not in `PATH`.  Let me show you real example where I need to deal with this.

In [OpenIndiana](https://www.openindiana.org/) we support several Python versions concurrently with one version considered as default.  Currently we support Python 3.7 and 3.9.  This means that we deliver `/usr/bin/python3.7` and `/usr/bin/python3.9` together with `/usr/bin/tox-3.7`, `/usr/bin/tox-3.9` and `/usr/bin/pytest-3.7`, `/usr/bin/pytest-3.9`.  There is also a symlink pointing to the default version, so we do have `/usr/bin/python -> python3.9`, `/usr/bin/tox -> tox-3.9`, `/usr/bin/pytest -> pytest-3.9`.

Now, when we want to test something for non-default Python version (e.g. 3.7), we just run `tox-3.7 --current-env`, then the `tox-current-env` plugin prepares the testing environment by creating symlink `python -> /usr/bin/python3.7`.  I think nothing else is prepared (by design).  When `tox` tries to run `pytest` binary directly it simply cannot find version 3.7 of `pytest` because the only `pytest` in the system is symlink to `pytest-3.9` and there is no way to solve that via `PATH`.  The only reasonable way to find `pytest-3.7` is via `python -m pytest`.

In addition, the `python -m pytest` way of invocation `pytest` is officially documented and supported way of running `pytest`, please see https://docs.pytest.org/en/latest/how-to/usage.html#calling-pytest-through-python-m-pytest.

Thank you.

asottile: I'm a pytest core dev. we should not cargo cult this unnecessarily for an unsupported plugin

mtelka: > btw those are done that way to pass -W, not for any other reason (and that can slash should be done with PYTHONWARNINGS or the pytest setting)

Sorry, but mypy does not pass -W.  Also, to pass -W to pytest you do not need to invoke it via `python -m pytest -W ...`, you can do direct `pytest -W ...` (at least `pytest --help` suggests so).

asottile: yes that's why I said unnecessarily, but if you take a look at the git history that's why it was changed in those projects

gaborbernat:
Closing per the discussion above.

Issue #2340, opened 2022-02-06
Folder support for -c


dvarrazzo:
Hello,

I am giving a spin to tox 4.0.0b1 on the [psycopg 3 project](https://github.com/psycopg/psycopg/).

Using tox 3, you can run tests using `tox -c psycopg -s`.

On tox4, the first error encountered is that the argument of -c cannot be a directory. That's fair: I can see that the help of tox 3 also says that it must be a config file, although it works specifying with a dir (akin to `make -C`, `git -C`, `tar -C`...). However the error is reported as a traceback:

```
(.venv) piro@baloo:~/dev/psycopg3$ tox4 -c psycopg -s
Traceback (most recent call last):
  File "/home/piro/dev/psycopg3/.venv/bin/tox4", line 8, in <module>
    sys.exit(run())
  File "/home/piro/dev/psycopg3/.venv/lib/python3.8/site-packages/tox/run.py", line 18, in run
...
  File "/usr/lib/python3.8/pathlib.py", line 1222, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
IsADirectoryError: [Errno 21] Is a directory: 'psycopg'
```

If this is an user error, probably a clean error message rather than a traceback would be more appropriate (and correctly placing the blame on the user :).

However, once the command line is fixed, a traceback occurs:

```
(.venv) piro@baloo:~/dev/psycopg3$ tox4 -c psycopg/tox.ini -e 3.8
3.8: internal error
Traceback (most recent call last):
  File "/home/piro/dev/psycopg3/.venv/lib/python3.8/site-packages/packaging/requirements.py", line 102, in __init__
    req = REQUIREMENT.parseString(requirement_string)
  File "/home/piro/dev/psycopg3/.venv/lib/python3.8/site-packages/pyparsing/core.py", line 1127, in parse_string
    raise exc.with_traceback(None)
pyparsing.exceptions.ParseException: Expected string_end, found '/'  (at char 7), (line:1, col:8)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/piro/dev/psycopg3/.venv/lib/python3.8/site-packages/tox/tox_env/python/pip/req/file.py", line 37, in __init__
    self._requirement: Requirement | Path | str = Requirement(req)
  File "/home/piro/dev/psycopg3/.venv/lib/python3.8/site-packages/packaging/requirements.py", line 104, in __init__
    raise InvalidRequirement(
[...]

PR #2550, opened 2022-11-26
Allow passing config directory without filename

ssbarnea:
Fixes: #2340

## Contribution checklist:

- [x] added/updated test(s)
      in message body
    ```rst
    ```

### DIFF
--- a/docs/changelog/2340.bugfix.rst
+++ b/docs/changelog/2340.bugfix.rst
@@ -1,0 +1,1 @@
+Allow passing config directory without filename. - by :user:`ssbarnea`.
--- a/src/tox/config/source/discover.py
+++ b/src/tox/config/source/discover.py
@@ -26,6 +26,17 @@ def discover_source(config_file: Path | None, root_dir: Path | None) -> Source:
         src = _locate_source()
         if src is None:
             src = _create_default_source(root_dir)
+    elif config_file.is_dir():
+        src = None
+        for src_type in SOURCE_TYPES:
+            candidate: Path = config_file / src_type.FILENAME
+            try:
+                src = src_type(candidate)
+                break
+            except ValueError:
+                continue
+        if src is None:
+            raise HandledError(f"could not find any config file in {config_file}")
     else:
         src = _load_exact_source(config_file)
     return src
--- a/tests/config/cli/test_cli_ini.py
+++ b/tests/config/cli/test_cli_ini.py
@@ -11,7 +11,10 @@ from pytest_mock import MockerFixture
 
 from tox.config.cli.ini import IniConfig
 from tox.config.cli.parse import get_options
+from tox.config.cli.parser import Parsed
 from tox.config.loader.api import Override
+from tox.config.main import Config
+from tox.config.source import discover_source
 from tox.pytest import CaptureFixture, LogCaptureFixture, MonkeyPatch
 from tox.session.env_select import CliEnv
 from tox.session.state import State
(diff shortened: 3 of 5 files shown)


docs/changelog/2340.bugfix.rst
```diff
@@ -0,0 +1 @@
+Allow call with -c folder. - by :user:`ssbarnea`.
```
gaborbernat, line 1: Make this a reference to the documentation.

src/tox/config/source/discover.py
```diff
@@ -27,6 +27,9 @@ def discover_source(config_file: Path | None, root_dir: Path | None) -> Source:
         if src is None:
             src = _create_default_source(root_dir)
     else:
+        if config_file.is_dir():
+            if (config_file / "tox.ini").is_file():
+                config_file = config_file / "tox.ini"
         src = _load_exact_source(config_file)
```
gaborbernat, line 33: This needs to be generic, so does not belong here. Needs to work for other configuration sources, not just `tox.ini`.  Take a look here for other sources:

https://github.com/tox-dev/tox/blob/rewrite/src/tox/config/source/setup_cfg.py
https://github.com/tox-dev/tox/blob/rewrite/src/tox/config/source/legacy_toml.py

gaborbernat (changes requested):

src/tox/config/source/discover.py
ssbarnea: TBH, I do not want to spend time trying to open N configuration filenames and see if they include tox section when a folder is gives, especially "legacy" ones. As v4 is a major version, I think we safely drop some from the list.

WDYT? Or maybe you have a suggestion about how to implement it and reuse tox own abilities?

If the argument is a folder, we could just do a pushd(folder) and run tox from there, without -c? Is that doable?

gaborbernat: > TBH, I do not want to spend time trying to open N configuration filenames and see if they include tox section when a folder is gives, especially "legacy" ones. As v4 is a major version, I think we safely drop some from the list.

Strong disagree.

gaborbernat: I'd imagine we'd repeat this loop here with the config file name passed is a directory. See how you can get the filename to fill in by using `s.FILENAME`.

src/tox/config/source/discover.py
```diff
@@ -58,9 +69,10 @@ def _create_default_source(root_dir: Path | None) -> Source:
     if root_dir is None:  # if set use that
         empty = Path.cwd()
         for base in chain([empty], empty.parents):
-            if (base / "pyproject.toml").exists():
-                empty = base
-                break
+            for f in SOURCE_TYPES:
```
gaborbernat, line 72: Single char variables for list compression only please

tests/config/cli/test_cli_ini.py
```diff
@@ -199,3 +202,24 @@ def test_cli_ini_with_interpolated(tmp_path: Path, monkeypatch: MonkeyPatch) ->
     monkeypatch.setenv("TOX_CONFIG_FILE", str(to))
     conf = IniConfig()
     assert conf.get("a", str)
+
+
+@pytest.mark.parametrize("conf_arg", ["", "tox.ini"])
```
gaborbernat, line 207: This test should test all other sources, not just tox.ini.

tests/config/cli/test_cli_ini.py
```diff
@@ -199,3 +202,24 @@ def test_cli_ini_with_interpolated(tmp_path: Path, monkeypatch: MonkeyPatch) ->
...
+def test_conf_arg(tmp_path: Path, monkeypatch: MonkeyPatch, conf_arg: str) -> None:
+    dest = tmp_path / "c"
+    dest.mkdir()
+    # write test tox.ini file
+    cfg = dest / "tox.ini"
+    cfg.write_bytes(b"[tox]")
+
+    with monkeypatch.context() as context:
```
gaborbernat, line 215: Why this, and release context right away?

gaborbernat (changes requested):

src/tox/config/source/discover.py
```diff
@@ -26,6 +26,17 @@ def discover_source(config_file: Path | None, root_dir: Path | None) -> Source:
         src = _locate_source()
         if src is None:
             src = _create_default_source(root_dir)
+    elif config_file.is_dir():
+        src = None
+        for src_type in SOURCE_TYPES:
+            candidate: Path = (config_file / src_type.FILENAME).resolve()
```
gaborbernat, line 32: Do we really need the resolve? I think can do without 🤔

src/tox/config/source/discover.py
```diff
@@ -58,9 +69,10 @@ def _create_default_source(root_dir: Path | None) -> Source:
     if root_dir is None:  # if set use that
         empty = Path.cwd()
         for base in chain([empty], empty.parents):
-            if (base / "pyproject.toml").exists():
-                empty = base
-                break
+            for file_path in SOURCE_TYPES:
+                if (base / file_path.FILENAME).exists():
```
gaborbernat, line 73: This change is unrelated 🤔 what's this about?

src/tox/config/source/discover.py
ssbarnea: Not really the code from `_create_default_source` is not compatible with our needs. We would need to do extra changes in it to make it reusable for the "folder" case.

gaborbernat: I wasn't talking about reusing that method here, but about not using resolve in this line.

src/tox/config/source/discover.py
ssbarnea: I think that was another bug I found while testing because it only checked for presence of `pyproject.toml` file, when in fact it should look for all the 3 supported filenames, in particular order from SOURCE_TYPES.

Basically before this it was looking on parents only for pyproject.toml.

src/tox/config/source/discover.py
ssbarnea: True, I removed it. I added it while testing when something did not work right and I wanted to see how it goes.

src/tox/config/source/discover.py
gaborbernat: Do we have a test to validate this then? The one you added I think only deals with the folder parts of it.

gaborbernat (approved):

src/tox/config/source/discover.py
ssbarnea: For some reason I do not see any codecov job reporting coverage. That would make it much easier to decide.

Issue #2366, opened 2022-03-03
tox4: Regression for file system dependencies


daneah:
We have a runtime application that depends on packages from the file system that live up and over from the runtime application's directory:

```
├── package/
└── runtime-app/
    ├── requirements.txt
    └── setup.cfg
```

In tox3, running the tests in the runtime app, from the runtime app's directory, successfully installs the package via `-e ../package` defined in `requirements.txt`.

In tox4, this situation instead produces the following error:

```python
Traceback (most recent call last):
  File "/.../tox/session/cmd/run/single.py", line 45, in _evaluate
    tox_env.setup()
  File "/.../tox/tox_env/api.py", line 226, in setup
    self._setup_env()
  File "/.../tox/tox_env/python/runner.py", line 91, in _setup_env
    self._install_deps()
  File "/.../tox/tox_env/python/runner.py", line 95, in _install_deps
    self.installer.install(requirements_file, PythonRun.__name__, "deps")
  File "/.../tox/tox_env/python/pip/pip_install.py", line 84, in install
    self._install_requirement_file(arguments, section, of_type)
  File "/.../tox/tox_env/python/pip/pip_install.py", line 95, in _install_requirement_file
    raise HandledError(f"{exception} for tox env py within deps")
tox.report.HandledError: '/.../package' is not in the subpath of '/.../runtime-app' OR one path is relative and the other is absolute. for tox env py within deps
```

ssbarnea:
That is caused by https://bugs.python.org/issue23082#msg366433 - I am writing a fix for it.

PR #2554, opened 2022-11-28
Allow installing relative paths that go outside tox root folder

ssbarnea:
Fixes: #2366 which is caused by https://github.com/python/cpython/issues/67271

## Contribution checklist:

- [x] added/updated test(s)
      in message body
    ```rst
    ```

### DIFF
--- a/src/tox/tox_env/python/pip/req/file.py
+++ b/src/tox/tox_env/python/pip/req/file.py
@@ -57,9 +57,15 @@ class ParsedRequirement:
                 else:
                     path = root / req
                 extra_part = f"[{','.join(sorted(extras))}]" if extras else ""
-                rel_path = str(path.resolve().relative_to(root))
-                if rel_path != "." and os.sep not in rel_path:  # prefix paths in cwd to not convert them to requirement
-                    rel_path = f".{os.sep}{rel_path}"
+                try:
+                    rel_path = str(path.resolve().relative_to(root))
+                    if (
+                        rel_path != "." and os.sep not in rel_path
+                    ):  # prefix paths in cwd to not convert them to requirement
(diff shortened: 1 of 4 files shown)


src/tox/tox_env/python/pip/req/file.py
```diff
@@ -57,9 +57,15 @@ def __init__(self, req: str, options: dict[str, Any], from_file: str, lineno: in
...
-                rel_path = str(path.resolve().relative_to(root))
-                if rel_path != "." and os.sep not in rel_path:  # prefix paths in cwd to not convert them to requirement
-                    rel_path = f".{os.sep}{rel_path}"
+                try:
+                    rel_path = str(path.resolve().relative_to(root))
+                    if (
+                        rel_path != "." and os.sep not in rel_path
+                    ):  # prefix paths in cwd to not convert them to requirement
+                        rel_path = f".{os.sep}{rel_path}"
+                except ValueError:
```
gaborbernat, line 66: Can we guard this different behavior with a version filter so we remember to remove it once the CPython fix lands in our lowest supported version?

gaborbernat (changes requested):
Test please 👍

src/tox/tox_env/python/pip/req/file.py
ssbarnea: They do not have the plan to fix the issue in stdlib, only the error message would be different. Basically they will document that `relative_to` is never expected to produce paths that start with `../`, instead it would fail.

gaborbernat: Fair, tests still in place.

gaborbernat (approved):

Issue #2362, opened 2022-02-19
tox4: Generative section names result in additional commands, not overridden commands


garrison:
I have noticed something that behaves differently on tox 4.0.0b1 compared with tox 3.24.5.

Consider the following `tox.ini` with generative section names:

```ini
[tox]
minversion = 3.15
envlist = py37, lint
skipsdist = true

[testenv]
usedevelop = true
install_command = pip install -U {opts} {packages}
deps = black
commands =
  pip freeze

[testenv:{black,lint}]
envdir = {toxworkdir}/lint
commands =
  black: black {posargs} .
  lint: black --check {posargs} .
```

When I run `tox -elint` (on version 3), this runs only the `black --check` line, as expected.

However, if I run `tox4 -elint`, it runs the `pip freeze` command under `[testenv]` before running `black --check`.  Instead of overriding `commands`, I end up with all the commands.

This only seems to happen with the generative section names, i.e. when I change it to `[testenv:lint]`, the behavior is as expected.

gaborbernat:
Well, I wish this wasn't a thing because this makes it really tricky to resolve situations like:

```ini
[testenv:{black,lint}]
envdir = {toxworkdir}/lint

[testenv:black]
envdir = {toxworkdir}/black
```

Which envdir wins here now?

PR #2557, opened 2022-11-28
Support for generative section names

gaborbernat:
For now, the first section found wins, with no conflict resolution when multiple section names qualify.

Resolves https://github.com/tox-dev/tox/issues/2362


jugmac00:
Could you please update our documentation? Probably here... https://tox.wiki/en/rewrite/user_guide.html#generative-environments

PR #2574, opened 2022-12-01
Avoid recreation of virtualenv due to symlinks

ssbarnea:
If people called tox alternating between `python3 -m tox` and `tox`,
they would endup triggering venv recreation, even if there was only
one python. We should always resolve symlinks to avoid such false
env invalidations.

# Thanks for contribution


ssbarnea:
@gaborbernat I would have written a test but I am not sure how to test for that. Maybe you can indicate me an existing test that loads the info and we can check that `executable` is already resolved inside the file?

gaborbernat (approved):

Issue #2568, opened 2022-11-30
Recursive extra support


gaborbernat:
https://github.com/python-attrs/attrs

![image](https://user-images.githubusercontent.com/690238/204834805-d74d6815-0944-413b-9016-4bb979e39186.png)

gaborbernat:
cc @hynek

PR #2579, opened 2022-12-02
Support for recursive extras

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/2568

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       - id: add-trailing-comma
         args: [--py36-plus]
   - repo: https://github.com/asottile/pyupgrade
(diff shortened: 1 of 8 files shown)


hynek:
Dumb question: could this be adapted for pip-tools which also lack the support? It's like the final piece for standards-based workflows: https://github.com/jazzband/pip-tools/issues/1685

gaborbernat:
That's a question for the pip-tools maintainers 👍

ssbarnea:
@hynek I keep an eye on pip-tools, I could help reviewing a PR. Still this in particular might be too tricky to implement and I would prefer to see lots of stuff removed from pip-tools before this would be added.

So far, I prefered to avoid recursivity.

Issue #2595, opened 2022-12-05
Running tox v4 rc2 leaks Python processes


jugmac00:
With tox v4 rc2, when I run `tox4 -rvv -e py310` for e.g. the testtools repository, a Python process will be created and stays there even when tox finishes.

## repro

```
git clone git@github.com:testing-cabal/testtools.git
cd testtools
tox4 -rvv -e py310
```

Afterwards, when tox has finished, when I have a look at top, I see the following output:
```
 185766 jugmac00  20   0   58404  52628   9472 R  99,7   0,3   1:01.93 python 
```

When I then inspect that process with py-spy, I see the following output:
![Screenshot from 2022-12-05 15-33-29](https://user-images.githubusercontent.com/9895620/205663459-9c9d7d93-179d-4a81-9178-40046f7bab4d.png)

jugmac00:
Thanks for fixing!

PR #2597, opened 2022-12-06
Better handling of build backend without editable support and add --exit-and-dump-after flag

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/2567.
Resolves https://github.com/tox-dev/tox/issues/2595.

### DIFF
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -108,9 +108,9 @@ class Python(ToxEnv, ABC):
         candidates: list[str] = []
         for factor in env_name.split("-"):
             spec = PythonSpec.from_string_spec(factor)
-            if spec.implementation is not None:
-                if spec.implementation.lower() in INTERPRETER_SHORT_NAMES and env_name is not None:
-                    candidates.append(factor)
+            impl = spec.implementation or "python"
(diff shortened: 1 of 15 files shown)


src/tox/tox_env/python/api.py
```diff
@@ -108,9 +108,9 @@ def extract_base_python(env_name: str) -> str | None:
         candidates: list[str] = []
         for factor in env_name.split("-"):
             spec = PythonSpec.from_string_spec(factor)
-            if spec.implementation is not None:
-                if spec.implementation.lower() in INTERPRETER_SHORT_NAMES and env_name is not None:
-                    candidates.append(factor)
+            impl = spec.implementation or "python"
```
nsoranzo, line 111: @gaborbernat This change has the unintended (I presume) effect of interpreting a numeric factor (e.g. `2105`) as a candidate base python, breaking previously valid environment names (e.g. `py37-foo-2105`) by raising `ValueError: conflicting factors py37, 2105 in py37-foo-2105`.
Example GitHub workflow build: https://github.com/galaxyproject/planemo/actions/runs/3654111042/jobs/6174245903 for `tox.ini` https://github.com/galaxyproject/planemo/blob/nsoranzo-patch-1/tox.ini

gaborbernat: That error is a bug 🤣 please fill an issue for it.

Issue #2615, opened 2022-12-07
`passenv` regression when multiple variables are named on the same line


masenf:
## Issue

In tox 4, space-separated variables specified for passenv are not passed into the container.

```
[testenv:inline]
passenv = FOO BAR
```
☝️  doesn't work on tox 4

```
[testenv:separate-lines]
passenv =
    FOO
    BAR
```
☝️ works fine on tox3 and tox4

## Environment

Provide at least:

- OS: ubuntu 20.04
- `pip list` of the host Python where `tox` is installed:

```console
% ./tox4/bin/pip list
Package       Version
------------- -------
cachetools    5.2.0
chardet       5.1.0
colorama      0.4.6
distlib       0.3.6
filelock      3.8.2
packaging     21.3
pip           20.0.2
pkg-resources 0.0.0
platformdirs  2.6.0
pluggy        1.0.0
pyparsing     3.0.9
pyproject-api 1.2.1
setuptools    44.0.0
tomli         2.0.1
tox           4.0.0
virtualenv    20.17.1
```

## Output of running tox

Provide the output of `tox -rvv`:

```console
inline: 235 W remove tox env folder /git/tox-passenv/.tox/inline [tox/tox_env/api.py:292]
inline: 292 I find interpreter for spec PythonSpec(path=/git/tox-passenv/tox4/bin/python3) [virtualenv/discovery/builtin.py:56]
inline: 292 D discover exe for PythonInfo(spec=CPython3.8.10.final.0-64, exe=/git/tox-passenv/tox4/bin/python3, platform=linux, version='3.8.10 (default, Jun 22 2022, 20:18:18) \n[GCC 9.4.0]', encoding_fs_io=utf-8-utf-8) in /usr [virtualenv/discovery/py_info.py:437]
inline: 293 D filesystem is case-sensitive [virtualenv/info.py:24]
inline: 294 D Attempting to acquire lock 139846177643632 on /home/mfurer/.local/share/virtualenv/py_info/1/df0893f56f349688326838aaeea0de204df53a132722cbd565e54b24a8fec5f6.lock [filelock/_api.py:172]
inline: 294 D Lock 139846177643632 acquired on /home/mfurer/.local/share/virtualenv/py_info/1/df0893f56f349688326838aaeea0de204df53a132722cbd565e54b24a8fec5f6.lock [filelock/_api.py:176]
[...]

PR #2621, opened 2022-12-07
Document necessity to escape `#` sign used in `commands`

### LINKED ISSUE #2617: Handling of "#" in commands has changed in tox4
## Issue

Sample tox.ini:

```
[testenv]
allowlist_externals = 
  bash
commands =
  bash -c "echo 'foo#bar'"
```

With tox3:

```
python run-test-pre: PYTHONHASHSEED='411589126'
python run-test: commands[0] | bash -c 'echo '"'"'foo#bar'"'"''
foo#bar
__________________________________________________________________________________________ summary ___________________________________________________________________________________________
  python: commands succeeded
  congratulations :)
```

With tox4:
```
py: commands[0]> bash -c '"echo '"'"'foo'
/usr/bin/bash: -c: line 1: unexpected EOF while looking for matching `"'
/usr/bin/bash: -c: line 2: syntax error: unexpected end of file
py: exit 2 (0.00 seconds) /home/jharbott/tmp> bash -c '"echo '"'"'foo' pid=125625
  py: FAIL code 2 (0.03=setup[0.02]+cmd[0.00] seconds)
  evaluation failed :( (0.06 seconds)
```

Escaping the `#` as `\#` solves the issue and seems to be backwards compatible, so maybe this is just the expected behaviour.

## Environment

Provide at least:

- OS: Ubuntu 22.04

jugmac00:
... `commands_pre` or `commands_post`.

This fixes #2617.
Resolves https://github.com/tox-dev/tox/issues/2615

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -409,6 +409,11 @@ Run
    Commands to run before running the :ref:`commands`. All evaluation and configuration logic applies from
    :ref:`commands`.
 
+   .. note::
+
+     When you use the ``#`` sign inside ``commands_pre``, you need to escape it, e.g.
+     ``commands_pre = bash -c "echo 'foo\#bar'"``.
+
 .. conf::
    :keys: commands
    :default: <empty list>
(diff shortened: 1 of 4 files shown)


docs/config.rst
```diff
@@ -409,6 +409,11 @@ Run
    Commands to run before running the :ref:`commands`. All evaluation and configuration logic applies from
    :ref:`commands`.
 
+   .. note::
+
+     When you use the ``#`` sign inside ``commands_pre``, you need to escape it, e.g.
```
gaborbernat, line 414: Please remove these from here. These config flags contain the generic description of the configuration. This behavior is also not bound just to commands. This is a generic ini config behaviour and should be documented in a ini config description area.

docs/faq.rst
```diff
@@ -42,6 +42,23 @@ tox 4 - removed tox.ini keys
 | ``isolated_build``       | Isolated builds are now always used.        |
 +--------------------------+---------------------------------------------+
 
+tox 4 - changed behaviour for tox.ini keys
+++++++++++++++++++++++++++++++++++++++++++
+
+- When using any of the ``commands``, ``commands_pre`` or ``commands_post`` keys, you need to escape the ``#`` sign.
```
gaborbernat, line 48: Remove reference to commands, this is a general ini file behavior; these commands is just an example.

gaborbernat (changes requested):

PR #2641, opened 2022-12-08
Handle encodings properly in SyncWrite, fixes #2422

### LINKED ISSUE #2422: UnicodeDecodeError on line 100 of execute/stream.py
Currently using Version: 4.0.0b2 on Windows, and getting an error whereby Tox is trying to decode everything using "utf-8" alone while there might be other valid encodings as well.

This is the problematic line for me that happened on Windows:
Line 100 onwards of "execute/stream.py"
```
    @property
    def text(self) -> str:
        with self._content_lock:
            return self._content.decode("utf-8")
```
            
When using breakpoint() there it seems that it will produce a UnicodeDecodeError with one of the self._content having an encoding of "Windows-1252" if I use the chardet library to detect it.

My own personal patch is modifying the function this way but I suppose a more elegant fix can be put in by someone else hopefully:

```
import chardet
...
...

    @property
    def text(self) -> str:
        with self._content_lock:
            encoding_detected: str | None = chardet.detect(self._content).get('encoding')
            if encoding_detected:
                return self._content.decode(encoding_detected)
            return self._content.decode('utf-8')
```

maxnoe:
This pull requests adds the `encoding` parameter to `SyncWrite` to hopefully fix #2422

While I have not yet produced a minimal working example that reproduces the bug in 2422, 
it is definitely wrong to always just assume utf-8.

- [ ] ran the linter to address style issues (`tox -e fix_lint`)

### DIFF
--- a/src/tox/execute/api.py
+++ b/src/tox/execute/api.py
@@ -122,7 +122,12 @@ class Execute(ABC):
             out, err = out_err
             out_buffer, err_buffer = out.buffer, err.buffer
             out_sync = SyncWrite(out_buffer.name, out_buffer if show else None, encoding=out.encoding)
-            err_sync = SyncWrite(err_buffer.name, err_buffer if show else None, Fore.RED if self._colored else None, encoding=err.encoding)
+            err_sync = SyncWrite(
+                err_buffer.name,
+                err_buffer if show else None,
+                Fore.RED if self._colored else None,
+                encoding=err.encoding,
+            )
             with out_sync, err_sync:
                 instance = self.build_instance(request, self._option_class(env), out_sync, err_sync)
                 with instance as status:
--- a/src/tox/execute/api.py
+++ b/src/tox/execute/api.py
@@ -119,9 +119,10 @@ class Execute(ABC):
         start = time.monotonic()
         try:
             # collector is what forwards the content from the file streams to the standard streams
-            out, err = out_err[0].buffer, out_err[1].buffer
-            out_sync = SyncWrite(out.name, out if show else None)
-            err_sync = SyncWrite(err.name, err if show else None, Fore.RED if self._colored else None)
+            out, err = out_err
+            out_buffer, err_buffer = out.buffer, err.buffer
+            out_sync = SyncWrite(out_buffer.name, out_buffer if show else None, encoding=out.encoding)
+            err_sync = SyncWrite(err_buffer.name, err_buffer if show else None, Fore.RED if self._colored else None, encoding=err.encoding)
             with out_sync, err_sync:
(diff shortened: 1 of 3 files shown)


maxnoe:
This is unfortunately not enough to fix the issue on windows, only part way there. The encoding of the text stream that `Execute.call` gets is already (wrongly) utf-8. I have a minimal example reproducing this on windows:

`hello.py:`

```
import sys

print(sys.stdout.encoding)
print("Hell\N{LATIN SMALL LETTER O WITH DIAERESIS}")
```

tox.ini
```
[tox]
envlist =
    test

[test]
description = install pytest in a virtual environment and invoke it on the tests folder
commands = python hello.py
```

Running `tox -e` test produces the unicode decode error shown in #2422, as tox wrongly assumes utf8 where the correct encoding (knowably) is cp1252

maxnoe:
What is very weird is that inside the tox process, `sys.stdout.encoding` is `utf-8` but inside the user python script `sys.stdout.encoding` is utf-8. I don't understand how this can happen.

gaborbernat:
> What is very weird is that inside the tox process, `sys.stdout.encoding` is `utf-8` but inside the user python script `sys.stdout.encoding` is utf-8. I don't understand how this can happen.

Yeah need to understand this because our own GIthub Actions against Windows are working https://github.com/tox-dev/tox/actions/runs/3646508070/jobs/6157666644 🤔

maxnoe:
The problem seems to be this: 

> The root cause is with the way python handles STDOUT. Python does some low level detection to get the encoding of the system and then uses a io.TextIOWrapper with the encoding set to what it detects and that's what you get in sys.stdout (stderr and stdin have the same).
> 
> Now, this detection returns UTF-8 when running in the shell because powershell works in UTF-8 and puts a layer of translation between the system and the running program but when piping to another program the communication is direct without the powershell translation, this direct communication uses the system's encoding which for windows is cp1252 (AKA Windows-1252).

From https://stackoverflow.com/a/57280184/

maxnoe:
> Yeah need to understand this because our own GIthub Actions against Windows are working

Are you sure there is non-ascii output in one of the test cases? Where could I add one?

maxnoe:
See also: https://bugs.python.org/issue27179

maxnoe:
This works:

```
[tox]
envlist =
    test


[test]
description = install pytest in a virtual environment and invoke it on the tests folder
commands = python hello.py
setenv =
    PYTHONIOENCODING=utf-8
```

gaborbernat:
Accepting https://github.com/tox-dev/tox/pull/2646 instead for now.

PR #2666, opened 2022-12-09
Add repr() calls to "invalid value" exception message in StrConverter.to_bool

ptmcg:
After upgrading to 4.0, we are getting the following exceptions in our build pipeline:

    File "/tmp/tmpw_2jv6x3/python-test-env/lib/python3.9/site-packages/tox/config/cli/parser.py", line 293, in add_color_flags
        if converter.to_bool(os.environ.get("NO_COLOR", "")):
    File "/tmp/tmpw_2jv6x3/python-test-env/lib/python3.9/site-packages/tox/config/loader/str_convert.py", line 90, in to_bool
        raise TypeError(f"value {value} cannot be transformed to bool, valid: {', '.join(StrConvert.VALID_BOOL)}")
    TypeError: value  cannot be transformed to bool, valid: , 0, 1, false, no, off, on, true, yes

We cannot reproduce the issue locally, else we would have more debugging tools and options.

This PR adds !r to display the repr() of the bad value, as well as repr()'s the values in VALID_BOOL, to help diagnose issues when running locally is not an option.

# Thanks for contribution


src/tox/config/loader/str_convert.py
```diff
@@ -81,13 +81,16 @@ def to_env_list(value: str) -> EnvList:
...
+        norm = str(value).strip().lower()
         if norm in StrConvert.TRUTHFUL_VALUES:
             return True
         elif norm in StrConvert.FALSE_VALUES:
             return False
         else:
-            raise TypeError(f"value {value} cannot be transformed to bool, valid: {', '.join(StrConvert.VALID_BOOL)}")
+            raise TypeError(
+                f"value {value!r} cannot be transformed to bool,"
+                f" valid values: {', '.join(repr(vb) for vb in StrConvert.VALID_BOOL)}",
```
gaborbernat, line 92: I don't want to use repr for valid values

gaborbernat (changes requested):

src/tox/config/loader/str_convert.py
ptmcg: When one of the valid values is the empty string, it makes for a weird message.

gaborbernat:
Also needs a test. My guess is you're passing an empty string for boolean.

ptmcg:
Thanks for the quick review, by the way!

That was our thought as well (passing in an empty string), but the logic in the code should be able to handle an empty string, as that appears to be one of the values in VALID_BOOL.

I looked for other tests, but did not see any for this method.

gaborbernat:
> I looked for other tests, but did not see any for this method.

https://github.com/tox-dev/tox/blob/main/tests/config/test_sets.py#L92

ptmcg:
Ah, I was looking for something specifically testing to_bool(). Found the appropriate test, updated the expected error message.

Issue #2691, opened 2022-12-12
setenv PATH behavior change with tox 4


dairiki:
## Issue

We have some python code that can use certain external binaries to provide extra functionality but is supposed to degrade gracefully if said binaries are not installed.  To test that graceful degradation, we have been using

```ini
[testenv]
setenv =
    noutils: PATH=/dev/null
...
```

with the intention that running `tox -e py311-noutils` would run our tests with `PATH` set to a broken value so that our code under test could not find the external binaries.

This worked up until tox 4.  With tox 4  `tox -e py311-noutils` reports 

```console
py311-noutils: skipped because could not find python interpreter with spec(s): py311
```

Apparently, the broken PATH setting in `setenv` now breaks tox's ability to find available python interpreters.

A minimal example is provided below.

Is this expected?  (If so, is there another suggested approach for my use case?)

## Minimal example

With this `tox.ini`
```ini
[tox]
isolated_build = true

[testenv]
setenv =
    noutils: PATH=/dev/null
commands =
    python -c "from shutil import which; exit(0 if which('ls') is None else 1)"
```

and this minimal `pyproject.toml`
```toml
[build-system]
requires = ["setuptools>=61"]
build-backend = "setuptools.build_meta"

[project]
name = "test-junk"
version = "0.1"
```

Using tox 4.0.5
```console
$ type python3.11
$ tox -e py311-noutils
py311-noutils: skipped because could not find python interpreter with spec(s): py311
  py311-noutils: SKIP (0.01 seconds)
  evaluation failed :( (0.05 seconds)
```

Using tox 3.27.1
```console
$ tox -e py311-noutils
.package create: /path/to/.tox/.package
.package installdeps: setuptools>=61
warning: sdist: standard file not found: should have one of README, README.rst, README.txt, README.md

py311-noutils create: /path/to/.tox/py311-noutils
py311-noutils inst: /path/to/.tox/.tmp/package/1/test-junk-0.1.tar.gz
py311-noutils installed: test-junk @ file:///path/to/.tox/.tmp/package/1/test-junk-0.1.tar.gz
py311-noutils run-test-pre: PYTHONHASHSEED='4193678857'
py311-noutils run-test: commands[0] | python -c 'from shutil import which; exit(0 if which('"'"'ls'"'"') is None else 1)'
___________________________________ summary ____________________________________
  py311-noutils: commands succeeded
  congratulations :)
```

[...]

PR #2704, opened 2022-12-13
Fixes issues when PATH is set in set_env

dairiki:
This PR fixes two issues that have appeared with tox 4 having to due with the handling of the PATH environment variable.

#### Use `os.environ["PATH"]` when locating python executable for environment

When creating a virtual env, use `os.environ["PATH"]` rather than any PATH from a `set_env` to locate the python executable for the environment. (See #2691 for a motivation.)

#### The logic for updating PATH when `ToxEnv._paths` is set was not respecting any PATH from `set_env`

`ToxEnv.environment_variables` caches its result.  There was logic in the *setter* for `ToxEnv._paths` to update the cached value for PATH when `_paths` was set.  (`ToxEnv._paths` contained a list of directories that are to be prepended to PATH.  It is used to add the virtual env's bindir to the PATH.) That logic was not respecting any customization for PATH in `set_env`, with the result that once the virtual env was constructed, any customizations to PATH from `set_env`  was discarded.

The fix here moves the logic for prefixing PATH to `tox.tox_env.python.api.Python`, since it's not used elsewhere.

## Checklist

### DIFF
--- a/src/tox/tox_env/api.py
+++ b/src/tox/tox_env/api.py
@@ -59,7 +59,7 @@ class ToxEnv(ABC):
 
         #: encode the run state of various methods (setup/clean/etc)
         self._run_state = {"setup": False, "clean": False, "teardown": False}
-        self._paths_private: list[Path] = []  #: a property holding the PATH environment variables
+        self._paths: list[Path] = []  #: directories to prepend to the PATH environment variable
         self._hidden_outcomes: list[Outcome] | None = []
         self._env_vars: dict[str, str] | None = None
         self._env_vars_pass_env: list[str] = []
@@ -310,19 +310,19 @@ class ToxEnv(ABC):
         pass_env: list[str] = self.conf["pass_env"]
         set_env: SetEnv = self.conf["set_env"]
         if self._env_vars_pass_env == pass_env and not set_env.changed and self._env_vars is not None:
-            return self._env_vars
+            return self._add_prefixed_paths(self._env_vars)
 
         result = self._load_pass_env(pass_env)
         # load/paths_env might trigger a load of the environment variables, set result here, returns current state
         self._env_vars, self._env_vars_pass_env, set_env.changed = result, pass_env, False
         # set PATH here in case setting and environment variable requires access to the environment variable PATH
-        result["PATH"] = self._make_path()
+        result["PATH"] = os.environ.get("PATH", "")
         for key in set_env:
             result[key] = set_env.load(key)
         result["TOX_ENV_NAME"] = self.name
         result["TOX_WORK_DIR"] = str(self.core["work_dir"])
         result["TOX_ENV_DIR"] = str(self.conf["env_dir"])
-        return result
+        return self._add_prefixed_paths(result)
 
     @staticmethod
     def _load_pass_env(pass_env: list[str]) -> dict[str, str]:
@@ -333,29 +333,17 @@ class ToxEnv(ABC):
                 result[env] = value
         return result
 
-    @property
-    def _paths(self) -> list[Path]:
-        return self._paths_private
-
-    @_paths.setter
-    def _paths(self, value: list[Path]) -> None:
-        self._paths_private = value
-        # also update the environment variable with the new value
-        if self._env_vars is not None:  # pragma: no branch
-            # remove duplicates and prepend the tox env paths
-            result = self._make_path()
-            self._env_vars["PATH"] = result
+    def _add_prefixed_paths(self, env_vars: dict[str, str]) -> dict[str, str]:
+        if not self._paths:
+            return env_vars
+        values = dict.fromkeys(str(i) for i in self._paths)
+        values.update(dict.fromkeys(env_vars["PATH"].split(os.pathsep)))
+        return {**env_vars, "PATH": os.pathsep.join(values)}
 
     @property
     def _allow_externals(self) -> list[str]:
         result: list[str] = [f"{i}{os.sep}*" for i in self._paths]
         result.extend(i.strip() for i in self.conf["allowlist_externals"])
-        return result
-
-    def _make_path(self) -> str:
-        values = dict.fromkeys(str(i) for i in self._paths)
-        values.update(dict.fromkeys(os.environ.get("PATH", "").split(os.pathsep)))
-        result = os.pathsep.join(values)
         return result
 
     def execute(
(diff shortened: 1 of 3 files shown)


gaborbernat (changes requested):

src/tox/tox_env/python/api.py
```diff
@@ -161,7 +163,22 @@ def _setup_env(self) -> None:
...
-        self._paths = self.prepend_env_var_path()  # now that the environment exist we can add them to the path
+        self._env_paths = self.prepend_env_var_path()  # now that the environment exist we can add them to the path
+
+    @property
+    def environment_variables(self) -> dict[str, str]:
+        env_vars = super().environment_variables
+        if self._env_paths:
+            values = dict.fromkeys(str(i) for i in self._env_paths)
+            values.update(dict.fromkeys(env_vars["PATH"].split(os.pathsep)))
+            env_vars = {**env_vars, "PATH": os.pathsep.join(values)}
```
gaborbernat, line 174: This is not python specific logic, so it's just wrong to push it here. Needs to be at the base level.

src/tox/tox_env/python/api.py
```diff
@@ -161,7 +163,22 @@ def _setup_env(self) -> None:
...
+        return env_vars
+
+    @property
+    def _allow_externals(self) -> list[str]:
```
gaborbernat, line 178: Same here.

src/tox/tox_env/python/virtual_env/api.py
```diff
@@ -109,6 +109,9 @@ def session(self) -> Session:
 
     def virtualenv_env_vars(self) -> dict[str, str]:
         env = self.environment_variables.copy()
+        # we want to use PATH from environment rather than any PATH
+        # modifications in set_env to find the python executable
+        env["PATH"] = os.environ.get("PATH", "")
```
gaborbernat, line 114: This is not virtualenv specific logic, so has no place to live here.

tests/tox_env/python/test_python_api.py
```diff
@@ -131,3 +132,18 @@ def test_base_python_env_conflict_show_conf(tox_project: ToxProjectCreator, igno
...
         )
     result.assert_out_err(out, "")
+
+
+def test_path_from_set_env(tox_project: ToxProjectCreator) -> None:
+    tox_ini = """
+        [testenv]
+        package = skip
+        set_env =
+            PATH=/dev/null
```
gaborbernat, line 142: I'm not sure this is a valid use case. By definition of tox path is where we find the executables used in commands; if you set it to null means no commands are valid. I'm willing to make this use case just not supported.

src/tox/tox_env/api.py
```diff
@@ -310,13 +309,13 @@ def environment_variables(self) -> dict[str, str]:
...
         set_env: SetEnv = self.conf["set_env"]
         if self._env_vars_pass_env == pass_env and not set_env.changed and self._env_vars is not None:
-            return self._env_vars
+            self._env_vars
 
         result = self._load_pass_env(pass_env)
         # load/paths_env might trigger a load of the environment variables, set result here, returns current state
         self._env_vars, self._env_vars_pass_env, set_env.changed = result, pass_env, False
         # set PATH here in case setting and environment variable requires access to the environment variable PATH
-        result["PATH"] = self._make_path()
```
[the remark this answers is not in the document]
gaborbernat: Please keep the current architecture, that was done on purpose.

tests/tox_env/python/test_python_api.py
dairiki: > I'm willing to make this use case just not supported.

Whether intended or not, this did work in tox 3.

That is the PATH from the set_env (prefixed by the virtual env's bindir) is used to find the executables used in the commands executed within that environment (as well as affecting the PATH environment variable seen by any of those commands.)
However, when *constructing* the virtual environment, tox 3 searches `os.environ["PATH"]` (unmodified by any `set_env` setting) to find the python executable.  Thereafter, the prefixing of any PATH with the virtual env's bindir ensures that python, as well as any other commands installed in the virtual env can be found regardless of how badly set_env mangles PATH.

(I do want it to mean "no commands, except those directly installed in the virtual env, are valid".)

----

This is exactly the use case I described in issue #2691.  Where I also asked, "Is this expected?".
You [replied](https://github.com/tox-dev/tox/issues/2691#issuecomment-1347224886), "PR to fix this is welcome", which I took to mean that this is in fact a valid use case.
If it's not, say so, and I will find some other way forward.

src/tox/tox_env/python/virtual_env/api.py
dairiki: This is within the method `virtualenv_env_vars` which is used to construct the environment variables used when constructing the virtual env.  Specifically, these are not the environment variables used when running commands within the virtual environment.

The PATH set here is used to find the python executable to be used to construct the virtual environment.

As such, it seems virtualenv-specific to me.

tests/tox_env/python/test_python_api.py
gaborbernat: > Whether intended or not, this did work in tox 3.

True, but that could have been by chance instead by design. 

> This is exactly the use case I described in issue #2691. Where I also asked, "Is this expected?".
> You [replied](https://github.com/tox-dev/tox/issues/2691#issuecomment-1347224886), "PR to fix this is welcome", which I took to mean that this is in fact a valid use case.
> If it's not, say so, and I will find some other way forward

Sorry for that, did not think through initially of its implications.

gaborbernat: I think your request is valid and sane; I'm just not sure this is the right solution.

src/tox/tox_env/python/virtual_env/api.py
dairiki: Wouldn't `base_python` be the better way to specify a custom interpreter?

src/tox/tox_env/api.py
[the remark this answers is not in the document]
dairiki: See if you like this better...

I've moved the prefixing of PATH to happen outside of the caching of `_env_vars`.

The other option would be to move the extraction of PATH from `set_env` into `_make_path`.  I think what I've done is cleaner, but either would work.

src/tox/tox_env/python/virtual_env/api.py
gaborbernat: That would be short-sighted because those custom python interpreters could not be discovered by virtualenv within the commands, e.g. imagine your test tries creating a new virtual environment nested. In general is better to use PATH as that makes it more nestable.

gaborbernat:
Sorry I'll need to reject this PR for reasons explained in the issue. This path is not desirable.

Issue #2661, opened 2022-12-09
Unerstand "min_version = 4.0" from tox 3 or use "minversion = 4.0" in tox 4 documentation


hroncok:
## What's the problem this feature will solve?

Considering a new project follows the current documentation and they set:

```ini
[tox]
min_version = 4.0
```

When this is read by tox3, it is ignored, because tox3 only reads `minversion`, not `min_version`.
A project maintainer that knows this will adjust the config as follows:

```ini
[tox]
minversion = 4.0
```

This is respected by both tox3 and tox4, and hence it works.

However, a project maintainer who does not know this and tries to follow the documentation will be confused.

## Describe the solution you'd like / Alternative Solutions

Either tox3 needs to be updated to also read min_version (this will work but only for users who have the latest version of tox3)
or the documentation needs to be updated to use `minversion` in the initial examples.

Users following the documentation for tox4 should be guided to create config files where the minial tox version requirement is respected in tox3 and tox4 both.

gaborbernat:
The without underscore version is kept for backwards compatibility and I don't want to encourage its use. So I don't want to do what you're asking. It's not our goal to create configurations going ahead that works for both tox 3 and 4. 3 is dead, let it die.

gaborbernat:
See https://tox.wiki/en/latest/config.html#min_version, specifying we support both, but the ordering is intentional.

hroncok:
> The without underscore version is kept for backwards compatibility and I don't want to encourage its use. 

The alternative is to backport the underscore version to a maintenance release of 3.

gaborbernat:
The transition path is to set min_version 4 and stop using version 3 😅 anything else is not supported. We plan to do no further development for version three.

gaborbernat:
@hroncok if you think this would be useful feel free to put in a PR against the legacy branch for min_version 👍

PR #2714, opened 2022-12-14
tox3: Support provision of tox 4 with the min_version option

hroncok:
Fixes https://github.com/tox-dev/tox/issues/2661

# Thanks for contribution

### DIFF
--- a/tests/integration/test_provision_int.py
+++ b/tests/integration/test_provision_int.py
@@ -25,7 +25,7 @@ def test_provision_missing(initproj, cmd):
             "tox.ini": """\
                 [tox]
                 skipsdist=True
-                minversion = 3.7.0
+                minversion = 3.7.0,<4
                 requires =
                     setuptools == 40.6.3
                 [testenv]
@@ -54,7 +54,7 @@ def test_provision_from_pyvenv(initproj, cmd, monkeypatch):
             "tox.ini": """\
                 [tox]
                 skipsdist=True
-                minversion = 3.7.0
+                minversion = 3.7.0,<4
                 requires =
                     setuptools == 40.6.3
                 [testenv]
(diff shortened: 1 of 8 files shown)


hroncok:
The new test needs to be skipped on old Pythons.

tests/unit/session/test_provision.py
```diff
@@ -27,14 +27,21 @@ def next_tox_major():
...
+
+
+def test_provision_min_version_is_requires(newconfig, minversion_option, next_tox_major):
     with pytest.raises(MissingRequirement) as context:
         newconfig(
             [],
             """\
             [tox]
-            minversion = {}
+            {} = {}
```
hroncok, line 42: I wonder if adding a new fixture that evaluates to this expression is worth it or not.

tests/integration/test_provision_int.py
```diff
@@ -83,7 +108,7 @@ def test_provision_interrupt_child(initproj, monkeypatch, capfd, signal_type):
             "tox.ini": """
                     [tox]
                     skipsdist=True
-                    minversion = 3.7.0
+                    minversion = 3.7.0,<4
```
hroncok, line 111: I don't know if this is explicitly supported but thankfully it works. Without it the tests were failing (as they were provisioning tox 4).

hroncok:
I'm not sure if the Windows failures are related or not. I don't have a Windows development environment to reproduce.

gaborbernat:
> I'm not sure if the Windows failures are related or not. I don't have a Windows development environment to reproduce.

Can't accept the PR without passing CI so would need to figure it out 👍

hroncok:
Attempting to solve the `UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 164: character maps to <undefined>` thing by reading files as utf-8.

hroncok:
Adding encoding=utf-8 to subprocess.check_output in the failing test seems to fix the problem however Python 3.5 does not support that parameter. I could try working around that by monkeypatching locale.getpreferredencoding to return utf-8 or by skipping the test on Python 3.5. Honestly, I have no idea what is going on here and why did this test not fail previously.

hroncok:
I went too far trying to bend the code for Python 3.5 and 2.7 and I completely forgot I can run subprocess on bytes and decode the output explicitly. Hopefully that'll work.

hroncok:
Now the coverage job failed with:

>  Could not find the branch to compare to. Does 'origin/master' exist?

Added another commit.

tests/integration/test_provision_int.py
```diff
@@ -68,6 +68,31 @@ def test_provision_from_pyvenv(initproj, cmd, monkeypatch):
...
+def test_provision_tox_4(initproj, cmd, monkeypatch):
+    initproj(
+        "pkg123-0.7",
+        filedefs={
+            "tox.ini": """\
+                [tox]
+                no_package=True
+                min_version = 4
+                [testenv]
+                commands=python -c "import os; print('assert this')"
```
hroncok, line 84: The import os part is a leftover from debugging a d should have been removed. Sorry about that.

hroncok:
Thanks for merging.

Issue #2690, opened 2022-12-12
The --pre option isn't passed to pip


fschulze:
With tox 4.0.8 the ``--pre`` option is not passed to the ``pip install`` command anymore, unlike tox 3. This is easily seen in the output of the ran commands with the ``-v`` option. As I see nothing in the changelog about this and the option isn't throwing an error I think this is a regression.

gaborbernat:
How did you set the pre-option?

fschulze:
I used ``--pre`` on the command line. It seems like the option doesn't exist anymore. It isn't listed in the cli help output of either ``tox`` or ``tox run``. It is still in the documentation though: https://tox.wiki/en/latest/config.html#pip-installer I would expect tox to fail if it encounters an unknown option. Is there a new alternative to that option?

fschulze:
I just saw that there is also a ``tox legacy`` command and the option is still listed there, but doesn't work either.

gaborbernat:
We're deprecating with v4 all flags controlling pip. You should be using the PIP environment variables instead, so here it would be `PIP_PRE=1` I think, we do already recommend it for custom index servers, see https://tox.wiki/en/latest/faq.html#using-a-custom-pypi-server. We should document this though and fix the legacy command --pre not working.

fschulze:
Thanks for pointing out ``PIP_PRE``, I didn't think of that.

Czaki:
@gaborbernat could you point out where PIP_PRE is documented in pip documentation? I cannot find them

gaborbernat:
That's a question for the pip developers.

Czaki:
> That's a question for the pip developers.

It is your decision to switch from pip documented feature (`--pre` flag) to the undocumented one (`PIP_PRE` env var).

gaborbernat:
Did you read the pip documentation before you started accusing for a free service? https://pip.pypa.io/en/stable/topics/configuration/#environment-variables I'll accept 5 dollar donation for doing your work myself.

Czaki:
I try to google it without any results. 
Maybe without hope that such basic things could be easily googlable, I could read the whole documentation. especially in tox documentation the `--pre` flag is still mentioned and no `PIP_PRE`, but `pip_pre` without any information that it is an environment variable that should be capitalized. 


I go back to debugging my tox configuration.

PR #2731, opened 2022-12-15
Fix legacy flags, deprecate them and remove --index-url

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/2690


tests/util/test_spinner.py
```diff
@@ -72,6 +72,7 @@ def test_spinner_atty(capfd: CaptureFixture, monkeypatch: MonkeyPatch) -> None:
 
 
 @time_machine.travel("2012-01-14", tick=False)
+@pytest.mark.flaky(max_runs=3, min_passes=1)
```
jayaddison, line 75: Reported upstream to adamchainz/time-machine#311

jayaddison: ...and looped back around to #2407 :recycle:

Issue #2720, opened 2022-12-14
test_result_json_sequential fails on macOS homebrew Python 3.11


jaraco:
Against b8b0803c:

<details>

```
 tox main $ tox -- --run-integration -k json_sequential -vv
.pkg: _optional_hooks> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: get_requires_for_build_editable> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: get_requires_for_build_wheel> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True hatchling.build
.pkg: build_wheel> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True hatchling.build
python: install_package> python -I -m pip install --force-reinstall --no-deps /Users/jaraco/code/tox-dev/tox/.tox/.tmp/package/9/tox-4.0.10.dev2+gb8b0803c-py3-none-any.whl
python: commands[0]> pytest --color=yes --run-integration -k json_sequential -vv
======================================================================= test session starts ========================================================================
platform darwin -- Python 3.11.0, pytest-7.2.0, pluggy-1.0.0 -- /Users/jaraco/code/tox-dev/tox/.tox/python/bin/python
cachedir: .tox/python/.pytest_cache
rootdir: /Users/jaraco/code/tox-dev/tox, configfile: pyproject.toml, testpaths: tests
plugins: xdist-3.1.0, flaky-3.7.0, mock-3.10.0, time-machine-2.8.2, devpi-server-6.8.0, cov-4.0.0
collecting ... collected 968 items / 967 deselected / 1 selected

tests/session/cmd/test_sequential.py::test_result_json_sequential FAILED                                                                                     [100%]

============================================================================= FAILURES =============================================================================
___________________________________________________________________ test_result_json_sequential ____________________________________________________________________

tox_project = <function init_fixture.<locals>._init at 0x104cdcb80>, enable_pip_pypi_access = None

    @pytest.mark.integration()
    def test_result_json_sequential(
        tox_project: ToxProjectCreator,
        enable_pip_pypi_access: str | None,  # noqa: U100
    ) -> None:
        cmd = [
[...]

PR #2749, opened 2022-12-18
test_result_json_sequential: use PythonInfo.current

masenf:
fixes test issue #2720 when using macOS homebrew installed python versions.

Getting `PythonInfo.current()` sets the returned `py_info.system_executable` to the real, resolved interpreter path that also gets returned by tox's `--result-json`, which deep in the internals, [calls `resolve()`](https://github.com/tox-dev/tox/blob/b365892901f81cf7a831d5e41dca7f7f43c4c107/src/tox/tox_env/python/virtual_env/api.py#L143) on the interpreter path.

In my experiments, the problem can also be resolved (without this patch) by NOT calling `resolve()` down in `tox_env.python.virtuale_env.api`, but that seems like a heavier fix than changing how an asserted value is constructed inside test code.

That said, I have NO IDEA what the difference really is between `PythonInfo.current` and `PythonInfo.current_system` other than the latter triggers a call to [`_resolve_to_system`](https://github.com/pypa/virtualenv/blob/6845f6f2660fc786c92a01e077b4ae07a87d5a6b/src/virtualenv/discovery/py_info.py#L407-L428); if you could enlighten me on why this change may be a bad idea, I'd be happy to correct my mental model.

# Thanks for contribution

### DIFF
--- a/tests/session/cmd/test_sequential.py
+++ b/tests/session/cmd/test_sequential.py
@@ -68,7 +68,7 @@ def test_result_json_sequential(
     with log.open("rt") as file_handler:
         log_report = json.load(file_handler)
 
-    py_info = PythonInfo.current_system()
+    py_info = PythonInfo.current()
     host_python = {
         "executable": py_info.system_executable,
         "extra_version_info": None,


masenf:
not quite :[

Issue #2657, opened 2022-12-09
Numeric factor in environment name wrongly causes "conflicting factors" error


nsoranzo:
https://github.com/tox-dev/tox/pull/2597 had the unintended effect of interpreting a numeric factor (e.g. `2105`) as a candidate base python, breaking previously valid environment names (e.g. `py37-foo-2105`) by raising `ValueError: conflicting factors py37, 2105 in py37-foo-2105`.
Example GitHub workflow build: https://github.com/galaxyproject/planemo/actions/runs/3654111042/jobs/6174245903 for `tox.ini` https://github.com/galaxyproject/planemo/blob/nsoranzo-patch-1/tox.ini

_Originally posted by @nsoranzo in https://github.com/tox-dev/tox/pull/2597#discussion_r1044065606_

gaborbernat:
> For an environment named `py310-foo-py`, I'm getting:

This is not the same case, and in your case the failure is expected as explained in other issue.

PR #2764, opened 2022-12-21
Skip factors that don't start with an alphabetic character

nsoranzo:
A numeric factor in an environment name can wrongly cause a "conflicting factors" error.
Fix https://github.com/tox-dev/tox/issues/2657 .

# Thanks for contribution

### DIFF
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -128,6 +128,9 @@ class Python(ToxEnv, ABC):
     def extract_base_python(env_name: str) -> str | None:
         candidates: list[str] = []
         for factor in env_name.split("-"):
+            if not factor[0].isalpha():
+                # This may be considered a valid spec by virtualenv, but not in tox
+                continue
             spec = PythonSpec.from_string_spec(factor)
             impl = spec.implementation or "python"
             if impl.lower() in INTERPRETER_SHORT_NAMES and env_name is not None and spec.path is None:
--- a/docs/changelog/2657.bugfix.rst
+++ b/docs/changelog/2657.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a numeric factor in an environment name can wrongly cause a "conflicting factors" error - by :user:`nsoranzo`.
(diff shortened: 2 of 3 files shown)


src/tox/tox_env/python/api.py
```diff
@@ -128,6 +128,9 @@ def default_base_python(self, conf: Config, env_name: str | None) -> list[str]:
     def extract_base_python(env_name: str) -> str | None:
         candidates: list[str] = []
         for factor in env_name.split("-"):
+            if not factor[0].isalpha():
```
gaborbernat, line 131: Wouldn't this break names like `3.7` that is valid (and used **https://github.com/psycopg/psycopg/actions/runs/3642624184/jobs/6149941052) ?** Or `3`?

nsoranzo: @gaborbernat Sorry for the late reply, holiday season!
I've looked at your example (had to go back to commit https://github.com/psycopg/psycopg/commit/14bfb96c4ad53a0c426406a587bea5d45fd605e4 because they just removed their use of tox) and if their use of `3.7` as environment name was supposed to run tests under Python 3.7 it didn't work under tox 3 (see https://gist.github.com/nsoranzo/a291d87bbfa43c37a9c01a6aaf0d887b , where tox3 creates the `3.7` environment using Python 3.8) and I don't think it was never documented as supported, only `py37`.

gaborbernat: It is supported with tox 4 though 🤔so we should add this feature by supporting it. 😅

nsoranzo: tox 4 has been use in for just 20 days, this change is [not documented](https://tox.wiki/en/latest/config.html#base_python) (what environment factor may now be used as Python specifier? py37, 3.7, 37, 3?) and, most importantly, breaks environment names (as reported in https://github.com/tox-dev/tox/issues/2657) which have been valid for years.
I don't think we can implement this feature without breaking some tox configuration that was working under the assumption that a number as environment factor is not interpreted as a Python specifier, but happy to help if you have a reasonable specification in mind.

gaborbernat: py37, 3.7, 37, 3

mostly these 🤔

gaborbernat: An easy fix will be that if we find `3.x.y`,  `3.x`, ' 3x`, or `3` factors, we still accept those here and pass them through but not other numbers.

gaborbernat:
Replaced by https://github.com/tox-dev/tox/pull/2849#pullrequestreview-1244158620

nsoranzo:
@gaborbernat Sorry I didn't have time to follow up on your suggestions!

@stephenfin Thanks a lot for working on this and for your explanation, my issue is indeed fixed! :tada:

PR #2786, opened 2022-12-28
Improved factor selection to allow multiple uses of `-f` for "OR" and to allow hyphenated factors

### LINKED ISSUE #2766: Doc request: how to do explicit invocations with multiple factors

Under tox v3, I was able to pass a string with multiple factors to `tox -e`.
For example, in a project with a `tomli` factor to control the installation of `tomli` for TOML support on py<3.11, the following invocation was valid in v3:
```
tox -e 'py{37,310}{,-tomli}'
```

The above expression would expand to an environment list in the same way that those factors would expand in the `envlist`.

This is particularly useful for declaring CI builds and trying to run a large test matrix efficiently.
A simple project with one factor may have an envlist which reads with multiple python versions and the factor enabled/disabled:
```
envlist = py{37,38,39,310,311}{,-tomli}
```
and in CI, `tox` may be invoked with `tox -e 'py{,-tomli}`.

I have tried to reimplement this kind of logic with tox v4 using labels and `tox r -m`, but for mildly complex cases it becomes very verbose.

The following usage fails with a `ValueError`:
```
tox r -e 'py37{,-tomli}'
```
As does `tox -m ci` with the config
```
labels =
    ci = py37{,-tomli}
```

On the assumption that this is possible and I simply haven't figured it out yet
- if someone could share the solution here, that would help
- let's get it into the documentation as part of the tox4 vs tox3 differences (I'm happy to open a PR once I know how to do this)

## Alternative Solutions

On the assumption that this is no longer possible and it isn't considered a desirable feature
- let's document that it's not supported anymore on the tox4 vs tox3 differences page
- some reasonable usage should be defined which covers this kind of usage, and offered in docs as an alternative

I think at a minimum, the docs should say that you used to be able to write `tox -e 'py{,-foo}'` and now you need to have an explicit `labels` config.

## Additional context

For a concrete example of a project where this broke, and the change which fixes it, refer to
https://github.com/python-jsonschema/check-jsonschema/pull/204

sirosen:
### Summary

To enable better factor selection, this PR applies two changes. (resolves #2766)

1. `-f` can now be passed multiple times. Envs matching any `-f` usage are selected (OR) on top of the existing behavior that envs must match all factors in a given usage (AND).

2. `-f` will now split factors with hyphens.

(1) is the key functional change. (2) is not strictly necessary, but makes usage conform a little better to user expectations (or at least the expectations of one user! 👋 😉 ).

### Example Usage

Consider an env list as follows:
```
envlist = py3{10,9}-django2{0,1}{,-cov}
```

Prior to this change, the following usage was valid: `tox l -f py39 django20` and would select `py39-django20 py39-django20-cov`.
After this change, the same selection can be expressed with `tox l -f py39-django20`. Furthermore, the following selection which was previously *not* possible can be expressed:

```
$ tox l -f py39-django20 -f django21-cov
py310-django21-cov
py39-django20
py39-django20-cov
py39-django21-cov
```

The groups of selections are combined with OR semantics, but the individual selections are combined with AND.

### Implementation Notes

I separated these two features into two commits, anticipating that (2) might be contentious. I would like to keep it as I don't think it adds much complexity, but I'm ready and willing to pull it out if necessary.

The parsing/interpretation of factors is now separated out into a dedicated helper method, which I note most other CLI args do not have. It looks like the CLI code avoids breaking things apart into too many small functions, but I thought this case benefits enough in terms of readability to be worth it. Again, I'm happy to change this in response to feedback.

The help message is rather long for this now, and I'm still not sure it's fully descriptive. Perhaps a future FAQ on how to use factors and labels would help.

### Checklist

### DIFF
--- a/docs/changelog/2766.feature.rst
+++ b/docs/changelog/2766.feature.rst
@@ -1,2 +1,1 @@
-* Allow ``-f`` to be used multiple times and to be used on hyphenated factors
-  (e.g. ``-f py311-django -f py39``)
+``-f`` can be used multiple times and on hyphenated factors (e.g. ``-f py311-django -f py39``) - by :user:`sirosen`.
--- a/docs/changelog/2766.feature.rst
+++ b/docs/changelog/2766.feature.rst
@@ -1,0 +1,2 @@
+* Allow ``-f`` to be used multiple times and to be used on hyphenated factors
+  (e.g. ``-f py311-django -f py39``)
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -91,8 +91,12 @@ def register_env_select_flags(
     if multiple:
         help_msg = "labels to evaluate"
         add_to.add_argument("-m", dest="labels", metavar="label", help=help_msg, default=[], type=str, nargs="+")
-        help_msg = "factors to evaluate (passing multiple factors means 'AND', passing this option multiple times means 'OR')"
-        add_to.add_argument("-f", dest="factors", metavar="factor", help=help_msg, default=[], type=str, nargs="+", action="append")
+        help_msg = (
+            "factors to evaluate (passing multiple factors means 'AND', passing this option multiple times means 'OR')"
+        )
+        add_to.add_argument(
+            "-f", dest="factors", metavar="factor", help=help_msg, default=[], type=str, nargs="+", action="append"
+        )
     help_msg = "exclude all environments selected that match this regular expression"
     add_to.add_argument("--skip-env", dest="skip_env", metavar="re", help=help_msg, default="", type=str)
     return add_to
@@ -288,10 +292,16 @@ class EnvSelector:
         self._manager.tox_add_env_config(pkg_conf, self._state)
         return pkg_env
 
+    def _parse_factors(self) -> list[set[str]]:
+        # factors is a list of lists, from the combination of nargs="+" and action="append"
+        # also parse hyphenated factors into lists of factors
+        # so that `-f foo-bar` and `-f foo bar` are treated equivalently
+        raw_factors = getattr(self._state.conf.options, "factors", [])
+        return [set(f for factor in factor_list for f in factor.split("-")) for factor_list in raw_factors]
+
     def _mark_active(self) -> None:
         labels = set(getattr(self._state.conf.options, "labels", []))
-        # factors is a list of lists, from the combination of nargs="+" and action="append"
-        factors = [set(factor_list) for factor_list in getattr(self._state.conf.options, "factors", [])]
+        factors = self._parse_factors()
 
         assert self._defined_envs_ is not None
         if labels or factors:
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -299,12 +299,12 @@ class EnvSelector:
         self._manager.tox_add_env_config(pkg_conf, self._state)
         return pkg_env
 
-    def _parse_factors(self) -> list[set[str]]:
+    def _parse_factors(self) -> tuple[set[str], ...]:
         # factors is a list of lists, from the combination of nargs="+" and action="append"
         # also parse hyphenated factors into lists of factors
         # so that `-f foo-bar` and `-f foo bar` are treated equivalently
         raw_factors = getattr(self._state.conf.options, "factors", [])
-        return [{f for factor in factor_list for f in factor.split("-")} for factor_list in raw_factors]
+        return tuple({f for factor in factor_list for f in factor.split("-")} for factor_list in raw_factors)
 
     def _mark_active(self) -> None:
         labels = set(getattr(self._state.conf.options, "labels", []))
@@ -323,9 +323,12 @@ class EnvSelector:
                         env_info.is_active = True
             if factors:  # if matches mark it active
                 for name, env_info in self._defined_envs_.items():
+                    if env_info.is_active:
+                        continue
                     for factor_set in factors:
                         if factor_set.issubset(set(name.split("-"))):
                             env_info.is_active = True
+                            break
 
     def __getitem__(self, item: str) -> RunToxEnv | PackageToxEnv:
         """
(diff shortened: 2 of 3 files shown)


docs/changelog/2766.feature.rst
```diff
@@ -0,0 +1,2 @@
+* Allow ``-f`` to be used multiple times and to be used on hyphenated factors
```
gaborbernat, line 1: Remove the leading start, and finish your sentence with a period.

gaborbernat: Also add by user directive at the end, see other entries in the changelog template.

gaborbernat (changes requested):
Also, fix the type checker issues.

docs/changelog/2766.feature.rst
sirosen: I wasn't sure what you meant, but judging from the template it looks like `- ` gets prepended, so I remove the `*` bullet-point.

I also did a little rephrase to be shorter and fixed the missing annotation.

tests/session/test_env_select.py
```diff
@@ -61,15 +63,32 @@ def test_label_core_and_trait(tox_project: ToxProjectCreator) -> None:
...
     ini = """
         [tox]
         env_list = py3{10,9}-{django20,django21}{-cov,}
         """
     project = tox_project({"tox.ini": ini})
-    outcome = project.run("l", "--no-desc", "-f", "cov", "django20")
+    outcome = project.run("l", "--no-desc", *selection_arguments)
     outcome.assert_success()
-    outcome.assert_out_err("py310-django20-cov\npy39-django20-cov\n", "")
+    outcome.assert_out_err("\n".join(expect_envs) + "\n", "")
```
gaborbernat, line 91: Instead of + can we use an f-string here?

tests/session/test_env_select.py
```diff
@@ -61,15 +63,32 @@ def test_label_core_and_trait(tox_project: ToxProjectCreator) -> None:
     outcome.assert_out_err("py310\npy39\nflake8\ntype\n", "")
 
 
-def test_factor_select(tox_project: ToxProjectCreator) -> None:
+@pytest.mark.parametrize(
+    ("selection_arguments", "expect_envs"),
+    [
+        (
+            ["-f", "cov", "django20"],
+            ["py310-django20-cov", "py39-django20-cov"],
```
gaborbernat, line 71: Think these should be a tuple as we never extend them, rather than list 🤔

src/tox/session/env_select.py
```diff
@@ -302,10 +321,11 @@ def _mark_active(self) -> None:
...
                     if labels.intersection(env_info.env.conf["labels"]):
                         env_info.is_active = True
-            if self._state.conf.options.factors:  # if matches mark it active
+            if factors:  # if matches mark it active
                 for name, env_info in self._defined_envs_.items():
-                    if factors.issubset(set(name.split("-"))):
-                        env_info.is_active = True
+                    for factor_set in factors:
+                        if factor_set.issubset(set(name.split("-"))):
+                            env_info.is_active = True
```
gaborbernat, line 328: We should probably add here a `break` 🤔 as there's no point setting this more than once in any circumstance

src/tox/session/env_select.py
```diff
@@ -288,9 +299,17 @@ def _get_package_env(self, packager: str, name: str, is_active: bool) -> Package
         self._manager.tox_add_env_config(pkg_conf, self._state)
         return pkg_env
 
+    def _parse_factors(self) -> list[set[str]]:
```
gaborbernat, line 302: this should be tuple not list

tests/session/test_env_select.py
sirosen: I'm trying to come up with some highly academic counterexample in which I prefer `+` but nothing comes to mind! 😁 

I'll switch it.

src/tox/session/env_select.py
sirosen: Yes, I noticed that about the previous logic as well -- I was thinking about adding an `if env_info.is_active: continue` to the top of the loop.

I'll add both optimizations: there's no point in continuing either loop when we know there's no possible change.

tests/session/test_env_select.py
sirosen: Actually, I hit a surprise with this.
When I try to write this as `f'{"\n".join(expect_envs)}\n'`, I find that's a SyntaxError (backslash in f-string expression).

So the possible ways I see to write this segment are...
1. As above, with `+`
2. With `.format()`: `"{}\n".format("\n".join(expected_envs))`
3. With an auxiliary var: `envlist = "\n".join(expected_envs); f"{envlist}\n"`
4. Append `""` to the env lists! 😝 

I think I might have found my case for using `+` for str concat. (I also think this example should be made prominent in PEP701... I may pipe up about it on the discourse thread if it hasn't been mentioned already.)

Am I missing an option? Which of these do you prefer?

gaborbernat: Probably format is my prefered here, but will take as is 🤷

gaborbernat (changes requested):

src/tox/session/env_select.py
```diff
@@ -302,10 +321,14 @@ def _mark_active(self) -> None:
                 for env_info in self._defined_envs_.values():
                     if labels.intersection(env_info.env.conf["labels"]):
                         env_info.is_active = True
-            if self._state.conf.options.factors:  # if matches mark it active
+            if factors:  # if matches mark it active
                 for name, env_info in self._defined_envs_.items():
-                    if factors.issubset(set(name.split("-"))):
-                        env_info.is_active = True
+                    if env_info.is_active:
+                        continue
```
gaborbernat, line 327: This line seems uncovered by any test:

```
src/tox/session/env_select.py (92.3%): Missing lines 327
```

sirosen: I went back to try to add a test and realized it's not possible because you can't use factors and labels in combination. Maybe it would be valuable above (it looks to me like it would apply on the `labels` trait check) but rather than overthinking it I'm just going to remove this.

sirosen:
I think I've addressed the outstanding changes needed. If this is going to squash merge, I think it's ready. Otherwise, I'd like to rebase it to tidy the history a little bit.

gaborbernat (approved):

PR #2792, opened 2022-12-30
Fix extracting extras from markers with many extras

q0w:
# Thanks for contribution

### DIFF
--- a/tests/tox_env/python/virtual_env/package/test_python_package_util.py
+++ b/tests/tox_env/python/virtual_env/package/test_python_package_util.py
@@ -51,6 +51,15 @@ def test_load_dependency_many_extra(pkg_with_extras: PathDistribution) -> None:
         assert str(left) == str(right)
 
 
+def test_load_dependency_poetry_many_extra() -> None:
+    requires = [Requirement('filelock<4.0.0,>=3.9.0; extra == "extras1" or extra == "extras2"')]
+    result = dependencies_with_extras(requires, {"extras1"}, "")
+    exp = [Requirement("filelock<4.0.0,>=3.9.0")]
+    for left, right in zip_longest(result, exp):
+        assert isinstance(right, Requirement)
+        assert str(left) == str(right)
+
+
 def test_loads_deps_recursive_extras() -> None:
     requires = [
         Requirement("no-extra"),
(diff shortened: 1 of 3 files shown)


q0w:
What should be done with requirements like ` 'sphinx>=3; extra == "extras1" and extra == "extras2"'`
```python
requires = [
        Requirement('filelock<4.0.0,>=3.9.0; extra == "extras1" or extra == "extras2"'),
        Requirement('sphinx>=3; extra == "extras1" and extra == "extras2"')
    ]
    result = dependencies_with_extras(requires, {"extras1"}, "")
```
What result is expected?

q0w:
Maybe its a bug in a poetry build backend, because setuptools make different METADATA Require-Dists
 ```
Provides-Extra: extras1
Requires-Dist: filelock (<4.0.0,>=3.9.0) ; extra == 'extras1'
Provides-Extra: extras2
Requires-Dist: filelock (<4.0.0,>=3.9.0) ; extra == 'extras2'

gaborbernat:
> What result is expected?

I'd say only filelock to show up, as sphinx needs both extra enabled.

> Maybe its a bug in the poetry build backend, because setuptools makes different METADATA Requires-Dists

Worth reporting it to them, but I think is valid to expect if something generates such to behave as my previous paragraph here describes. Can you add a test for it?

tests/tox_env/python/virtual_env/package/test_python_package_util.py
```diff
@@ -51,6 +51,15 @@ def test_load_dependency_many_extra(pkg_with_extras: PathDistribution) -> None:
         assert str(left) == str(right)
 
 
+def test_load_dependency_poetry_many_extra() -> None:
+    requires = [Requirement('filelock<4.0.0,>=3.9.0; extra == "extras1" or extra == "extras2"')]
```
gaborbernat, line 55: instead of naming it poetry let's just say it `extra_or` :+1: And lets add an `extra_and` variant please?

gaborbernat (changes requested):

gaborbernat (approved):
Let's ship as is and do the `and` handling as a follow-up.

q0w:
How many logic markers should tox handle? One?

PR #2808, opened 2023-01-03
Document ambiguous usages in v3-to-v4

### LINKED ISSUE #2728: tox4: Unable to run env called `run`
## Issue

I have tox environment called `run` I use to run some commands inside virtual environment where my project is installed (i.e. conveniently testing CLI of the development builds) and since tox version 4.0 it fails with error.

As I can see, there were subcommands now introduced to tox – and apparently the `run` is being intercepted from the `-e`. This is not the correct behavior.

## Environment

Provide at least:

- OS: Arch Linux (up to date)
- `pip list`

```console
Package       Version
------------- -------
cachetools    5.2.0
chardet       5.1.0
colorama      0.4.6
distlib       0.3.6
filelock      3.8.2
packaging     22.0
pip           22.2.2
platformdirs  2.6.0
pluggy        1.0.0
pyproject_api 1.2.1
setuptools    63.2.0
tomli         2.0.1
tox           4.0.11
virtualenv    20.17.1
```

## Output of running tox

```console
[sdatko@polluks tox]$ tox3 -e run -- echo 'I am in venv'
run recreate: /home/sdatko/Development/tox/.tox/run
run run-test-pre: PYTHONHASHSEED='350351148'
run run-test: commands[0] | echo 'I am in venv'
I am in venv
__________________ summary __________________
  run: commands succeeded
  congratulations :)
```
```console
[sdatko@polluks tox]$ tox3 -e other -- echo 'I am in venv'
other create: /home/sdatko/Development/tox/.tox/other
other run-test-pre: PYTHONHASHSEED='1845399330'
other run-test: commands[0] | echo 'I am in venv'
I am in venv
__________________ summary __________________
  other: commands succeeded
  congratulations :)
```
```console
[sdatko@polluks tox]$ tox4 -e run -- echo 'I am in venv'
usage: tox run [-h] [--colored {yes,no}] [--exit-and-dump-after seconds] [-c file] [--workdir dir] [--root dir] [--runner {virtualenv}] [-v | -q] [-e ENV | -m label [label ...] | -f factor [factor ...] | --skip-env re] [--result-json path]
               [-s [v]] [-n] [-b] [--installpkg INSTALL_PKG] [--develop] [--hashseed SEED] [--discover path [path ...]] [--no-recreate-pkg] [--skip-pkg-install] [--version] [--no-provision [REQ_JSON]] [--no-recreate-provision] [-r]
               [-x OVERRIDE]
tox run: error: argument -e: expected one argument
```
```console
[sdatko@polluks tox]$ tox4 -e other -- echo 'I am in venv'
other: commands[0]> echo 'I am in venv'
I am in venv
  other: OK (0.13=setup[0.13]+cmd[0.00] seconds)
  congratulations :) (0.23 seconds)
```

## Minimal example

If possible, provide a minimal reproducer for the issue:

The `tox.ini` file:

```console
[tox]
env_list = py310
skipsdist = True

[testenv]
allowlist_externals = echo
commands = echo default

[testenv:run]
commands = {posargs}

[testenv:other]
commands = {posargs}
```

sirosen:
Add a section to the upgrading doc, "Legacy CLI entry point may result in ambiguous usages"

The new section resolves #2728

---

Upon review, I'm not sure if this should be written as "ambiguous usages" or "environments whose names match subcommands". I'm open to renaming or rewriting it as appropriate.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -151,13 +151,39 @@ Re-use of environments
     deps = pytest<7
 
 
-Legacy CLI entry point may result in ambiguous usages
------------------------------------------------------
+CLI command compatibility
+-------------------------
 
 ``tox`` 4 introduced dedicated subcommands for various usages.
-However, when no recognized command is given the legacy entry point which imitates ``tox`` 3 is used.
+However, when no subcommand is given the legacy entry point which imitates ``tox`` 3 is used.
 
-As a result, commands which were previously unambiguous may have become ambiguous.
+This compatibility feature makes most ``tox`` 3 commands work in ``tox`` 4, but there are some exceptions.
+
+Updating usage with ``-e``
+++++++++++++++++++++++++++
+
+In ``tox`` 3, environments could be specified to run with the ``-e`` flag.
+In ``tox`` 4, environments should always be specified using the ``-e`` flag to the ``run`` subcommand.
+
+Rewrite usages as follows
+
+.. code:: bash
+
+    # tox 3
+    tox -e py310,style
+
+    # tox 4
+    tox run -e py310,style
+
+    # or, tox 4 with the short alias
+    tox r -e py310,style
+
+Environment names matching commands
++++++++++++++++++++++++++++++++++++
+
+Now that ``tox`` has subcommands, it is possible for arguments to ``tox`` or its options to match those subcommand
+names.
+When that happens, parsing can become ambiguous between the ``tox`` 4 usage and the legacy fallback behavior.
 
 For example, consider the following tox config:
 
@@ -177,8 +203,9 @@ For example, consider the following tox config:
 This defines an environment whose name matches a ``tox`` 4 command, ``list``.
 
 Under ``tox`` 3, ``tox -e list`` specified the ``list`` environment.
+However, under ``tox`` 4, the parse of this usage as an invocation of ``tox list`` takes precedence over the legacy
+behavior.
 
-Under ``tox`` 4, this usage parses as an invocation of ``tox list``.
 Therefore, attempting that same usage results in an error:
 
 .. code:: bash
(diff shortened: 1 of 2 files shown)


docs/upgrading.rst
```diff
@@ -149,3 +149,51 @@ Re-use of environments
...
+    $ tox -e list
+    ...
+    tox: error: unrecognized arguments: -e
+
+It can be run either with an explicit ``tox run`` or an explicit use of the
+legacy command:
+
+.. code:: bash
+
+    $ tox run -e list
```
gaborbernat, line 195: For what's worth this is my recommended way to run unambiguous command names too, so let's State that. Assuming legacy by default is just a backwards compatibility layer and might be removed with tox 5.

gaborbernat: Worth mentioning that you don't have to type run fully, the r letter acts as a shorter alias 👍

docs/upgrading.rst
```diff
@@ -149,3 +149,51 @@ Re-use of environments
...
+Therefore, attempting that same usage results in an error:
+
+.. code:: bash
+
+    $ tox -e list
+    ...
+    tox: error: unrecognized arguments: -e
+
+It can be run either with an explicit ``tox run`` or an explicit use of the
+legacy command:
```
gaborbernat, line 191: Please align all lines to 120 width .

docs/upgrading.rst
```diff
@@ -149,3 +149,51 @@ Re-use of environments
 
     [testenv:b]
     deps = pytest<7
+
+
+Legacy CLI entry point may result in ambiguous usages
```
gaborbernat, line 154: Probably should be shorter: CLI compatibility is enough.

docs/upgrading.rst
sirosen: Will do! 👍

docs/upgrading.rst
sirosen: Okay, I'll come back to you with something that states that more clearly, and we can wordsmith it more if necessary.

sirosen: In order to put this front-and-center, I decided to split the section up a little more.
First it states that you should rewrite any use of `tox -e foo` to `tox run -e foo`/`tox r -e foo` without trying to explain why. The second subsection focuses on the bad case of an environment matching a command name.

Hope you like it this way!

docs/upgrading.rst
sirosen: There's another section on "CLI arguments", so I called it "CLI command compatibility", but we could also combine the sections later maybe.

gaborbernat (approved):

PR #2812, opened 2023-01-03
update how extras are extracted to handle cases with more than 2 groups

### LINKED ISSUE #2791: Bug with poetry and overlapping extras
## Issue

I'm not sure if this is a poetry bug or a tox bug, but I figured I'd start here.  The issue seems to be with how:
https://github.com/tox-dev/tox/blob/c83819262b4b739899af8a0bd7b6f32442344e4a/src/tox/tox_env/python/virtual_env/package/util.py#L31
is parsing extras that fall into multiple extra groups.  That is if you have the same requirement in 2+ extra groups, poetry will e.g., produce the line:

```
Requires-Dist: filelock (>=3.9.0,<4.0.0) ; extra == "extras1" or extra == "extras2"
```
in the `PKG-INFO`

Then when `tox` is run with e.g., `extras = extras1`, you get:

```
py-extras1: install_package_deps> python -I -m pip install 'filelock<4.0.0,>=3.9.0; or extra == "extras2"'
...
pip._vendor.packaging.markers.InvalidMarker: Invalid marker: 'or extra == "extras2"', parse error at 'or extra'
```

## Environment

Provide at least:

- OS: macos

```console
./.venv/bin/pip list
Package       Version
------------- -------
cachetools    5.2.0
chardet       5.1.0
colorama      0.4.6
distlib       0.3.6
filelock      3.9.0
packaging     22.0
pip           22.3.1
platformdirs  2.6.2
pluggy        1.0.0
pyproject_api 1.2.1
setuptools    65.5.0
tox           4.1.0
virtualenv    20.17.1
```

## Minimal example

I've created a minimal poetry project that reproduces this bug: https://github.com/dconathan/poetry-tox-bug

dconathan:
Resolves #2791.

- [ ] ~updated/extended the documentation~ (n/a)

This PR makes somewhat significant (?) changes to how extras are extracted from markers to accommodate situations where an extra falls into >2 extras groups (fixes #2791).  The previous approach involved mutating the `_markers` list while iterating over it, which was problematic when dealing with longer lists.  The new approach builds a new `_markers` list.


tests/tox_env/python/virtual_env/package/test_python_package_util.py
```diff
@@ -67,5 +67,13 @@ def test_loads_deps_recursive_extras() -> None:
...
     requires = [Requirement('filelock<4.0.0,>=3.9.0; extra == "extras1" or extra == "extras2"')]
-    result = dependencies_with_extras(requires, {"extras1"}, "")
-    assert [str(r) for r in result] == ["filelock<4.0.0,>=3.9.0"]
+    for extras in ["extras1", "extras2"]:
+        result = dependencies_with_extras(requires, {extras}, "")
+        assert [str(r) for r in result] == ["filelock<4.0.0,>=3.9.0"]
+
+
+def test_load_dependency_requirement_many_or_extras() -> None:
+    requires = [Requirement('filelock<4.0.0,>=3.9.0; extra == "extras1" or extra == "extras2" or extra == "extras3"')]
```
gaborbernat, line 76: While we're at it, can we also handle and expression? 🤔 we should also add in the test a python version marker to make sure that's kept and we only strip extras

gaborbernat (changes requested):

dconathan:
Resorted to some `type: ignore` fixes for mypy.  I don't have much mypy experience so I'm open to feedback/suggestions for alternatives for those cases...

tests/tox_env/python/virtual_env/package/test_python_package_util.py
dconathan: Yeah I was thinking there are probably all kinds of corner cases / nasty expressions we can come up with to break this... I'll add some more tests

Do we have an example of a realistic "and" expression we want to support?  Or is it purely hypothetical/synthetic?

gaborbernat: I haven't seen it but is possible 🤔

gaborbernat (approved):

gaborbernat (approved):

Issue #2754, opened 2022-12-19
ignore_basepython_conflict option is not in effect in tox4


dkolvakh:
## Issue

Seems ignore_basepython_conflict option is not in effect in tox4. Setting it up to True or False doesn't change anything if envlist contains py310 but python3 points to python3.8.

## Environment

Provide at least:

- OS: Linux Mint 20.3, running in VirtualBox
- `pip list` of the host Python where `tox` is installed:

```console
Package               Version             
--------------------- --------------------
appdirs               1.4.3               
apt-clone             0.2.1               
apturl                0.5.2               
astunparse            1.6.3               
attrs                 22.1.0              
autopage              0.5.1               
beautifulsoup4        4.8.2               
blinker               1.4                 
Brlapi                0.7.0               
ccsm                  0.9.14.1            
certifi               2019.11.28          
chardet               3.0.4               
Click                 7.0                 
cliff                 4.0.0               
cmd2                  2.4.2               
colorama              0.4.3               
command-not-found     0.3                 
compizconfig-python   0.9.14.1            
configobj             5.0.6               
coverage              6.4.4               
cryptography          2.8                 
cupshelpers           1.0                 
dbus-python           1.2.16              
ddt                   1.6.0               
defer                 1.0.6               
distlib               0.3.0               
distro                1.4.0               
entrypoints           0.3                 
execnet               1.9.0               
extras                1.0.0               
filelock              3.0.12              
fixtures              4.0.1               
flake8                6.0.0               
future                0.18.2              
glob2                 0.7                 
grpcio                1.16.1              
hammett               0.9.3               
httplib2              0.14.0              
idna                  2.8                 
ifaddr                0.1.6               
IMDbPY                6.8                 
importlib-metadata    4.12.0              
iniconfig             1.1.1               
junit-xml             1.8                 
keyring               18.0.1              
launchpadlib          1.10.13             
lazr.restfulclient    0.14.2              
lazr.uri              1.0.3               
linecache2            1.0.0               
logutils              0.3.5               
louis                 3.12.0              
lxml                  4.5.0               
Mako                  1.1.0               
MarkupSafe            1.1.0               
mccabe                0.7.0               
more-itertools        4.2.0               
mutmut                2.4.1               
netaddr               0.7.19              
netifaces             0.10.4              
oauthlib              3.1.0               
onboard               1.4.1               
packaging             20.3                
parso                 0.8.3               
pbr                   5.4.5               
pexpect               4.6.0               
[...]

PR #2824, opened 2023-01-05
Fail on mismatched python spec attributes

stephenfin:
The following is intended to be a correct tox configuration:

```ini
[tox]
min_version = 4.1
env_list = py{38,39,310,311},docs

[testenv]
base_python = python3.8

[testenv:docs]
commands =
  sphinx-build ...
```

The goal of this is to use `python3.8` (i.e. the value of `[testenv] base_python`) for all environments except those with specific `pyXX` factors in them. This helps eliminate issues where testenvs that do not specify a `pyXX` factor run with different Python versions in different environments.

An earlier bug, #477, prevented us from doing this. Due to #477, the interpreter version indicated by `[testenv] base_python` (or rather `[testenv] basepython` - the underscore-separated variant was only introduced in tox 4) would override the value indicated by the `pyXX` factor. This was resolved with a PR, #841, which started warning users if they were unintentionally running against the wrong interpreter and introduced the `ignore_basepython_conflict` value to opt into the correct behavior.

Unfortunately, this has been partially broken in the move to tox 4. While the above configuration still works, the following no longer does:

```ini
[tox]
min_version = 4.1
env_list = py{38,39,310,311},docs

[testenv]
base_python = python3

[testenv:docs]
commands =
  sphinx-build ...
```

This configuration was common back during the move from Python 2 to Python 3. It ensured that `python3` was used for all testenvs that did not request an explicit Python version via a factor or their own `base_python` version. While it's no longer necessary, it is still quite common. Currently, running with this configuration will result in `python3` being used for every environment, rather than e.g. `python3.8` for a testenv with the `py38` factor. This is clearly incorrect. This issue stems from an errant bit of logic. When comparing interpreter versions as part of the `_validate_base_python`, we ignore various attributes if they are unset on either `[testenv] base_python` or the `pyXX` factor. This allows e.g. `python3` to match `python3.8`, since the minor version is unset for the `[testenv] base_python` value, `python3`. The fix is simple: while we can ignore unset attributes for factor-derived versions (to allow a `py3` factor to work with e.g. `base_python = python3.8`), we should insist on them for `base_python`-derived versions.

Closes: #2754

### DIFF
--- a/docs/changelog/2754.bugfix.rst
+++ b/docs/changelog/2754.bugfix.rst
@@ -1,2 +1,2 @@
-Setting ``[testenv] basepython = python3`` will no longer override the Python
-interpreter version requested by a factor, such as ``py311``.
+Setting ``[testenv] basepython = python3`` will no longer override the Python interpreter version requested by a factor,
+such as ``py311`` - by :user:`stephenfin`.
--- a/docs/changelog/2754.bugfix.rst
+++ b/docs/changelog/2754.bugfix.rst
@@ -1,0 +1,2 @@
+Setting ``[testenv] basepython = python3`` will no longer override the Python
+interpreter version requested by a factor, such as ``py311``.
--- a/tests/tox_env/python/test_python_api.py
+++ b/tests/tox_env/python/test_python_api.py
@@ -95,6 +95,7 @@ def test_base_python_env_no_conflict(env: str, base_python: list[str], ignore_co
         ("py3", ["py2"], ["py2"]),
         ("py38", ["py39"], ["py39"]),
         ("py38", ["py38", "py39"], ["py39"]),
+        ("py38", ["python3"], ["python3"]),
         ("py310", ["py38", "py39"], ["py38", "py39"]),
         ("py3.11.1", ["py3.11.2"], ["py3.11.2"]),
         ("py3-64", ["py3-32"], ["py3-32"]),
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -154,7 +154,7 @@ class Python(ToxEnv, ABC):
                 if any(
                     getattr(spec_base, key) != getattr(spec_name, key)
                     for key in ("implementation", "major", "minor", "micro", "architecture")
-                    if getattr(spec_base, key) is not None and getattr(spec_name, key) is not None
+                    if getattr(spec_name, key) is not None
                 ):
                     msg = f"env name {env_name} conflicting with base python {base_python}"
                     if ignore_base_python_conflict:


stephenfin:
Docs failures don't appear to be due to this...

docs/changelog/2754.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Setting ``[testenv] basepython = python3`` will no longer override the Python
```
gaborbernat, line 1: Wrap at 120 and add ```- by :user:`x`. ```

tests/tox_env/python/test_python_api.py
```diff
@@ -95,6 +95,7 @@ def test_base_python_env_no_conflict(env: str, base_python: list[str], ignore_co
         ("py3", ["py2"], ["py2"]),
         ("py38", ["py39"], ["py39"]),
         ("py38", ["py38", "py39"], ["py39"]),
+        ("py38", ["python3"], ["python3"]),
```
gaborbernat, line 98: This shouldn't conflict, but `py27` and `python3` should 🤔

gaborbernat (changes requested):

docs/changelog/2754.bugfix.rst
gaborbernat: > The following is intended to be a correct tox configuration:
> 
> ```ini
> [tox]
> min_version = 4.1
> env_list = py{38,39,310,311},docs
> 
> [testenv]
> base_python = python3.8
> 
> [testenv:docs]
> commands =
>   sphinx-build ...
> ```
> 
> The goal of this is to use `python3.8` (i.e. the value of `[testenv] base_python`) for all environments except those with specific `pyXX` factors in them. This helps eliminate issues where testenvs that do not specify a `pyXX` factor run with different Python versions in different environments.

Strong disagree. This is not a good way to express this. This config should hard fail for all envs expect py38.

tests/tox_env/python/test_python_api.py
stephenfin: Both should conflict. `python3` on my system == `python3.11.1`. That is decidedly _not_ `python3.8`, which is what I intended. I should be exploding on that combination - which I do (unless the user explicitly said not to, via `ignore_base_python_conflict`)

gaborbernat: You're miss understanding. `base_python` **is not** an executable match. It is a specification. And the specification states `Python 3`, without resolving it.

docs/changelog/2754.bugfix.rst
gaborbernat: > But that makes `base_python` a special-case? For everything else, the idea is that `[testenv]` allows you to set "default" values that you can later override (or extend) in `[testenv:foo]` sections. tox itself does this for various things like `command`:

It's not a special case. Follows the same rule.
```ini
[testenv]
base_python = python3.8
[testenv:py39]
base_python = python3.9 
```

Would work. Where you're mistaken is that `py38` only sets `base_python` to `py38` if at no point is `base_python` specified. If you set it at `testenv` or `[testenv:py39]` level that no longer activates.

tests/tox_env/python/test_python_api.py
stephenfin: That may be the case, but providing a *specification* of `python3` (e.g. `base_python = python3`) causes the `python3` *executable* to be used. For example, using a `tox.ini` like this:

```ini
[tox]
minversion = 3.1
envlist = py{37,38,39,310,311}

[testenv]
basepython = python3
```

If I run `tox -e py38`, the `py38` venv is using `python3.11` :

```
❯ source .tox/py38/bin/activate
❯ python --version
Python 3.11.1
```

No warnings. Nothing. That's got to be incorrect behaviour, surely?

gaborbernat: Indeed, so we have two options:

- for py38 `python3` spec should conflict and hard fail,
- hard fail once we resolve `python3` and see its something else.

I'm inclined to go for option 1.

stephenfin: Agree. Option 1 is the way to go. Option 2 will result in different behaviour depending on the environment making tox less deterministic. That's A Bad Thing :tm:

Without this change, the `py38` venv uses whatever `python3` is on my host (`python3.11`). With this change, things hard fail as expected.

```
❯ tox -e py38
py38: failed with env name py38 conflicting with base python python3
  py38: FAIL code 1 (0.00 seconds)
  evaluation failed :( (0.08 seconds)
```

So it seems I have implemented option 1, yes?

gaborbernat: What happens for env `py3`?

stephenfin:
```
❯ tox -e py3
py3: install_deps> python -I -m pip install -r /tmp/tox-issue-2717/test-requirements.txt
.pkg: _optional_hooks> python /tmp/tox-venv/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist> python /tmp/tox-venv/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: prepare_metadata_for_build_wheel> python /tmp/tox-venv/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist> python /tmp/tox-venv/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
py3: install_package_deps> python -I -m pip install requests
py3: install_package> python -I -m pip install --force-reinstall --no-deps /tmp/tox-issue-2717/.tox/.tmp/package/11/foo-0.0.0.tar.gz
.pkg: _exit> python /tmp/tox-venv/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  py3: OK (4.39 seconds)
  congratulations :) (4.42 seconds)
```

gaborbernat (approved):

stephenfin:
:partying_face: Thanks. Happy to have gotten this resolved :smile:

PR #2830, opened 2023-01-06
Be stricter about our Python factors

### LINKED ISSUE #2657: Numeric factor in environment name wrongly causes "conflicting factors" error

stephenfin:
Fixes #2657 by only using non-prefixed factors that look like Python versions, rather than accepting virtually anything.

We also opt to discourage non-prefixed factors since these are far too likely to conflict with other non-Python factors and the current behavior is very magical. We can fully remove this functionality in a future release.

# Thanks for contribution

### DIFF
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -147,7 +147,22 @@ class Python(ToxEnv, ABC):
             ):
                 continue
 
-            impl = spec.implementation or "python"
+            impl = spec.implementation
+
+            # TODO: Drop in a future version
+            # PythonSpec.from_string_spec set implementation to None for py and python
+            if impl is None and not factor.startswith("py") and not factor.startswith("python"):
+                logging.warning(
+                    "identified the %s factor in the %s testenv as a Python spec but the interpreter implementation "
+                    "was not specified. This will be a ignored in future versions of tox. Please prefix this factor "
+                    "with an interpreter short name (one of: %s)",
+                    factor,
+                    env_name,
+                    ", ".join(INTERPRETER_SHORT_NAMES + ["py"]),
+                )
+
+            impl = impl or "python"
+
             if impl.lower() in INTERPRETER_SHORT_NAMES and spec.path is None:
                 candidates.append(factor)
 
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -132,9 +132,25 @@ class Python(ToxEnv, ABC):
         candidates: list[str] = []
         for factor in env_name.split("-"):
             spec = PythonSpec.from_string_spec(factor)
+
+            # The regex used to parse a Python spec in PythonSpec.from_string_spec is extraordinarily loose so we do our
+            # own sanity checks. In short, if the user does not specify an implementation then a major version will
+            # be present and we insist that it must look like a known Python major version number. If they did specify
+            # an implementation then a major version might not be present, but we will verify it if it is (for
+            # consistency).
+            if (
+                factor != "py"
+                and (  # py is a special case where version info can be unset
+                    (spec.implementation and spec.major) or not spec.implementation
+                )
+                and spec.major not in (2, 3)
+            ):
+                continue
+
             impl = spec.implementation or "python"
-            if impl.lower() in INTERPRETER_SHORT_NAMES and env_name is not None and spec.path is None:
+            if impl.lower() in INTERPRETER_SHORT_NAMES and spec.path is None:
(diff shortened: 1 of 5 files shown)


stephenfin:
Need to add tests and a release note. Next week's work though.

gaborbernat:
@stephenfin do you plan picking up this today, or shall I cut the new release without this?

stephenfin:
I'll wrap this up shortly.

gaborbernat (approved):

stephenfin:
This is done now. However, I think https://github.com/tox-dev/tox/issues/2848 is a far better, less complex approach. If we did merge this, I'd be proposing removing all of what was added here if/when we'd merge #2848 or some variant of same.

gaborbernat:
I cannot accept #2848 as is for the reasons outlined in it. We can fine-tune the range of valid specs, though; but that's a conversation for that PR.

src/tox/tox_env/python/api.py
```diff
@@ -132,9 +132,40 @@ def extract_base_python(env_name: str) -> str | None:
...
+                    (spec.implementation and spec.major) or not spec.implementation
+                )
+                and spec.major not in (2, 3)
+            ):
+                continue
+
+            impl = spec.implementation
+
+            # TODO: Drop in a future version
+            # PythonSpec.from_string_spec set implementation to None for py and python
```
gaborbernat, line 153: Remove the TODO. We haven't decided if we'll do or not.

src/tox/tox_env/python/api.py
```diff
@@ -132,9 +132,40 @@ def extract_base_python(env_name: str) -> str | None:
...
+            if impl is None and not factor.startswith("py") and not factor.startswith("python"):
+                logging.warning(
+                    "identified the %s factor in the %s testenv as a Python spec but the interpreter implementation "
+                    "was not specified. This will be a ignored in future versions of tox. Please prefix this factor "
+                    "with an interpreter short name (one of: %s)",
+                    factor,
+                    env_name,
+                    ", ".join(list(INTERPRETER_SHORT_NAMES) + ["py"]),
+                )
```
gaborbernat, line 162: Remove this warning please. `tox r -e 3.8` I think should be valid and we should default to `py` prefix in that case.

tests/tox_env/python/test_python_api.py
```diff
@@ -71,6 +71,28 @@ def test_diff_msg_no_diff() -> None:
...
+    [
+        ("py3", "py3"),
+        ("py311", "py311"),
+        ("pypy2", "pypy2"),
+        ("functional-py310", "py310"),
+        ("bar-pypy2-foo", "pypy2"),
+        ("310", "310"),
+        ("5", None),
+        # TODO: This is acceptable to 'PythonSpec.from_string_spec' but is
+        # clearly not valid :(
```
gaborbernat, line 85: Remove todo 👍 we can discuss this in an issue, don't need code comments for it.

tests/tox_env/python/test_python_api.py
```diff
@@ -71,6 +71,28 @@ def test_diff_msg_no_diff() -> None:
...
+        ("2000", "2000"),
+        ("4000", None),
+    ],
+    ids=lambda a: "|".join(a) if isinstance(a, list) else str(a),
```
gaborbernat, line 89: use i instead of a?

gaborbernat (commented):
Some stylistic feedback, but otherwise looks good.

stephenfin:
Thanks for the review. I think doing our own parsing for factors [*] is a better than plugging holes in virtualenv's `PythonSpec.from_string_spec`, which really is not suitable for this use case. I'm going to close this in favour of https://github.com/tox-dev/tox/pull/2849. I know that's not quite ready yet but we can work through the kinks together. Obviously, please feel free to cut a new tox release now while we discuss it.

[*] Specifically factors. `base_python` parsing can(/should?) continue to be done by `PythonSpec.from_string_spec`

PR #2843, opened 2023-01-09
The tests actually require wheel

hroncok:
When wheel is not installed, several tests error like this:

    distutils.errors.DistutilsModuleError: invalid command 'bdist_wheel'
    ...
    SystemExit: error: invalid command 'bdist_wheel'

Or like this:

    ERROR Missing dependencies:
    	wheel

When the tests are executed in a virtualenv, wheel is implicitly installed, which is why this have never happened on the CI.

List of tests that error:

     ERROR tests/tox_env/python/virtual_env/package/test_package_cmd_builder.py::test_tox_install_pkg_wheel
     ERROR tests/tox_env/python/virtual_env/package/test_package_cmd_builder.py::test_tox_install_pkg_sdist
     ERROR tests/tox_env/python/virtual_env/package/test_package_cmd_builder.py::test_install_pkg_via[p]
     ERROR tests/tox_env/python/virtual_env/package/test_package_cmd_builder.py::test_install_pkg_via[le]
     ERROR tests/tox_env/python/virtual_env/package/test_python_package_util.py::test_load_dependency_no_extra
     ERROR tests/tox_env/python/virtual_env/package/test_python_package_util.py::test_load_dependency_many_extra

# Thanks for contribution

- [-] ensured there are test(s) validating the fix
- [-] updated/extended the documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ optional-dependencies.testing = [
   "pytest-mock>=3.10",
   "pytest-xdist>=3.1",
   "re-assert>=1.1",
+  "wheel>=0.38.4",
   "time-machine>=2.8.2; implementation_name != \"pypy\"",
 ]
 scripts.tox = "tox.run:run"
--- a/docs/changelog/2843.bugfix.rst
+++ b/docs/changelog/2843.bugfix.rst
@@ -1,0 +1,1 @@
+Explicitly list ``wheel`` as requirement for the tests, as some of the tests error without it.


pyproject.toml
```diff
@@ -60,6 +60,7 @@ optional-dependencies.testing = [
   "pytest-mock>=3.10",
   "pytest-xdist>=3.1",
   "re-assert>=1.1",
+  "wheel",
```
gaborbernat, line 63: `wheel>=0.38.4` please

gaborbernat (changes requested):

pyproject.toml
hroncok: I don't understand this requirement. It works fine with wheel 0.37.1 as well.

gaborbernat: yet we only test in CI with 0.38.4, so 0.37.1 is happenstance not actually validated

hroncok: I see. Wouldn't that be the same issue the minute 0.38.5 is released?

hroncok: Anyway, amended.

hroncok: That explains why we need to patch the requirement versions out so often in Fedora.

gaborbernat: You can do that, but I'm washing my hands if people submit bugs for it 😆

hroncok: Absolutely undestood.

gaborbernat (changes requested):
Please add a changelog 👍

hroncok:
Added a bugfix changelog entry.

gaborbernat (approved):

PR #2854, opened 2023-01-12
Add Documentation URL for quick access via PyPI web

kerel-fs:
This PR adds a link to the Documentation in the pyproject.toml. This link will be shown in the PyPI web quick links, rendered with a book icon.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ readme.content-type = "text/markdown"
 keywords = ["virtual", "environments", "isolated", "testing"]
 license = "MIT"
 urls.Homepage = "http://tox.readthedocs.org"
+urls.Documentation = "https://tox.wiki"
 urls.Source = "https://github.com/tox-dev/tox"
 urls.Tracker = "https://github.com/tox-dev/tox/issues"
 urls."Release Notes" = "https://tox.wiki/en/latest/changelog.html"


pyproject.toml
```diff
@@ -10,6 +10,7 @@ readme.content-type = "text/markdown"
 keywords = ["virtual", "environments", "isolated", "testing"]
 license = "MIT"
 urls.Homepage = "http://tox.readthedocs.org"
+url.Documentation = "https://tox.wiki/"
```
gaborbernat, line 13: No need for the trailing /

gaborbernat (changes requested):

pyproject.toml
jugmac00, line 13: I am not super familiar with pyproject.toml to configure meta data for projects, but I am pretty sure this should be `urls.Documentation`, right?

See https://hatch.pypa.io/latest/config/metadata/#urls

jugmac00 (changes requested):

pyproject.toml
kerel-fs: indeed, fixed.

kerel-fs: how many errors can I put in a one-line change...

thanks! Fixed.

gaborbernat (approved):

Issue #2732, opened 2022-12-15
tox 3.27.1 to 4.0.9, environment var substitution failing


techpriest002:
## Issue

All of our build pipelines are failing due to what appears to be a bug with environment variable substitution in our tox.ini.  The variable substitution works as expected with tox==3.27.1, but we start having issues with tox==4.0.9.

## Environment

Provide at least:

- OS: Windows
- `pip list` of the host Python where `tox` is installed:

```console
(.venv) C:\tmp\tox_4.0.9_bug>pip list
Package       Version
------------- -------
cachetools    5.2.0
chardet       5.1.0
colorama      0.4.6
distlib       0.3.6
filelock      3.8.2
packaging     22.0
pip           22.3.1
platformdirs  2.6.0
pluggy        1.0.0
pyproject_api 1.2.1
setuptools    65.6.3
tomli         2.0.1
tox           4.0.9
virtualenv    20.17.1
wheel         0.38.4
```

## Output of running tox

Provide the output of `tox -rvv`:

```console
(.venv) C:\tmp\tox_4.0.9_bug>tox -rvv
default: 2108 I find interpreter for spec PythonSpec(path=C:\tmp\tox_4.0.9_bug\.venv\Scripts\python.exe) [virtualenv\discovery\builtin.py:56]
default: 2134 D got python info of C:\Users\nws2293\AppData\Local\Programs\Python\Python39\python.exe from C:\Users\nws2293\AppData\Local\pypa\virtualenv\py_info\1\72aa99cb99a91779d7fa96aaf09f6d3b1207abe0138f99ab0af57c8a95fa9102.json [virtualenv\app_data\via_disk_folder.py:129]
default: 2173 I proposed PythonInfo(spec=CPython3.9.10.final.0-64, system=C:\Users\nws2293\AppData\Local\Programs\Python\Python39\python.exe, exe=C:\tmp\tox_4.0.9_bug\.venv\Scripts\python.exe, platform=win32, version='3.9.10 (tags/v3.9.10:f2f3f53, Jan 17 2022, 15:14:21) [MSC v.1929 64 bit (AMD64)]', encoding_fs_io=utf-8-cp1252) [virtualenv\discovery\builtin.py:63]
[...]

PR #2861, opened 2023-01-15
Rewrite substitution parser

masenf:
* Fix issue #2732 by rewriting the regex-based substitution expression parser with a recursive-style parser.
* Substitutions are processed left to right, inner-most to outer-most.
* After a replacement is made in a string, the replaced part of the string will NOT affect characters after it. the result of a replacement will never be used in further substitution parsing.
  * On windows `{/}` will expand to a literal backslash `\`, which will NOT act as an escape for the next character.
* Additional test cases and regression tests are added. 
* Migrate and update "Substitutions" section of the "Configuration" doc from tox 3

# Thanks for contribution

### DIFF
--- a/src/tox/config/loader/ini/replace.py
+++ b/src/tox/config/loader/ini/replace.py
@@ -9,7 +9,7 @@ import sys
 from configparser import SectionProxy
 from functools import lru_cache
 from pathlib import Path
-from typing import TYPE_CHECKING, Iterator, Pattern
+from typing import TYPE_CHECKING, Any, Iterator, Pattern, Sequence, Union
 
 from tox.config.loader.api import ConfigLoadArgs
 from tox.config.loader.stringify import stringify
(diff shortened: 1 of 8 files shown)


masenf:
This is a rather large, fiddly PR. The more review I can get on this the better 🙏.

src/tox/config/loader/ini/replace.py
```diff
@@ -21,74 +21,171 @@
...
-        replace_value = replace_env(conf, args, conf_args)
-    elif of_type == "tty":
-        replace_value = replace_tty(args)
-    elif of_type == "posargs":
-        replace_value = replace_pos_args(conf, args, conf_args)
-    else:
-        replace_value = replace_reference(conf, loader, value, conf_args)
-    return replace_value
+class MatchError(Exception):
+    pass
```
gaborbernat, line 45: Docstring instead of pass.

tests/config/loader/ini/replace/test_replace.py
```diff
@@ -2,26 +2,79 @@
...
-        (r"[\] []", (4, 5, "posargs")),
-        (r"\[] []", (4, 5, "posargs")),
-        ("{foo}", (0, 4, "foo")),
-        (r"\{foo} {bar}", (7, 11, "bar")),
-        ("{foo} {bar}", (0, 4, "foo")),
-        (r"{foo\} {bar}", (7, 11, "bar")),
-        (r"{foo:{bar}}", (5, 9, "bar")),
-        (r"{\{}", (0, 3, r"\{")),
-        (r"{\}}", (0, 3, r"\}")),
+        ("[]", [MatchExpression([["posargs"]])]),
```
gaborbernat, line 13: Oh my, because you've changed the tests makes this pr infinitely harder to review. How am I supposed to know you did not introduce a lot of breaking changes 😅

masenf: i'm realizing now that `find_replace_part` was actually a public API of the `tox.config.loader.ini.replace` module, so I don't know if I can just "replace" it like this (even though the only caller in setenv was updated).

I took some time tonight to try to reimplement the `find_replace_part` API on top of the new `find_replace_expr` recursive implementation, and it's definitely possible, but the `MatchExpression` needs to track position information for where in the original value the match was found to allow reconstruction of the`(start, end, replacement)` tuple.

with that shim in place, I'd be able to bring back the original tests with no modification (only new tests, or those with clear behavior changes) to hopefully improve the reviewability here.

the other option would be keeping the regex parser, or at least the `find_replace_expr` part as it is currently and leave these tests in place... even though the ini-loader would actually be using the new parser under the hood.

thanks for looking.

gaborbernat: I'll accept this as is, for now; we don't have public API besides the plugin interface, but I ask you to stick around for any regression bug reports and help out with that please.

masenf: > I ask you to stick around for any regression bug reports and help out with that please.

100% please ping me if anything comes through that I don't notice and I'll jump on it.

rwols:
It seems that this changed behavior for the `TOX_ENV_NAME`, `TOX_WORK_DIR` injected environment variables.

Before this PR, you could reference these variables in the `commands` section with `{env:TOX_ENV_NAME}`. But after this PR, these placeholders resolve to the empty string.

PR #2865, opened 2023-01-16
Provision: ignore other test environments

### LINKED ISSUE #2862: testenv cannot specify `runner = X` when `X` runner type is provided by a plugin in `[tox] requires` list
## Issue

Please see the minimal [repro repo](https://github.com/masenf/repro-requires-runner-plugin-repro)

In short:
  * My plugin is registering a new runner type, call it "example"
  * A `tox.ini` file has a test environment that wants to set `runner = example`
  * The `tox.ini` file depends on the plugin via [`requires`](https://tox.wiki/en/latest/config.html#requires)

### Expect:

`tox` provisions a new virtualenv, `.tox`, containing the named dependencies from `requires` and then re-executes in the new environment.

### Actual:

`tox` fails with the following stack

```
ROOT: 71 D setup logging to DEBUG on pid 12802 [tox/report.py:221]
ROOT: 99 W will run in automatically provisioned tox, host /Users/masen/code/tox-dev/venv/tox-upstream/bin/python3.11 is missing [requires (has)]: tox-example-plugin [tox/provision.py:124]
  File "/Users/masen/code/tox-dev/minimal-plugin2/../venv/tox-upstream/bin/tox", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/run.py", line 19, in run
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/run.py", line 41, in main
    result = provision(state)
             ^^^^^^^^^^^^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/provision.py", line 125, in provision
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/provision.py", line 143, in run_provision
    tox_env: PythonRun = cast(PythonRun, state.envs[name])
                                         ~~~~~~~~~~^^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/session/env_select.py", line 336, in __getitem__
    return self._defined_envs[item].env
           ^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/session/env_select.py", line 192, in _defined_envs
    run_env = self._build_run_env(name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/session/env_select.py", line 251, in _build_run_env
    runner = REGISTER.runner(cast(str, env_conf["runner"]))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/masen/code/tox-dev/venv/tox-upstream/lib/python3.11/site-packages/tox/tox_env/register.py", line 72, in runner
    return self._run_envs[name]
           ~~~~~~~~~~~~~~^^^^^^
KeyError: 'example'
```

## Environment

Provide at least:

- OS: macOS ventura 13.1, homebrew python 3.11

```console
cachetools==5.2.1
chardet==5.1.0
colorama==0.4.6
distlib==0.3.6
filelock==3.9.0
packaging==23.0
platformdirs==2.6.2
pluggy==1.0.0
pyproject_api==1.4.0
tox==4.2.8
virtualenv==20.17.1
```

## Output of running tox / minimal example

https://github.com/masenf/repro-requires-runner-plugin-repro

## Possible Fix

With the following patch, I no longer see the issue (and all existing tests pass)

```
diff --git a/src/tox/session/env_select.py b/src/tox/session/env_select.py
index f926beda..786d3d4b 100644
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -248,6 +248,14 @@ class EnvSelector:
             env_conf.loaders.insert(0, self._provision[2])
         desc = "the tox execute used to evaluate this environment"
         env_conf.add_config(keys="runner", desc=desc, of_type=str, default=self._state.conf.options.default_runner)
+        runner_name = env_conf["runner"]
+        try:
+            runner = REGISTER.runner(cast(str, runner_name))
+        except KeyError:
+            if self._provision and self._provision[0]:
+                LOGGER.warning("provision: skip environment %s with unknown runner %r", name, runner_name)
+                return None
+            raise
         runner = REGISTER.runner(cast(str, env_conf["runner"]))
         journal = self._journal.get_env_journal(name)
         args = ToxEnvCreateArgs(env_conf, self._state.conf.core, self._state.conf.options, journal, self._log_handler)
```

Not sure if this is the "correct" fix, so @gaborbernat, I'll see if you have any comment before opening PR.

masenf:
When running provision, ignore other environments, since their configuration might not be valid until running in the provisioned environment.

When creating provision environment, seed the loader early via `memory_seed_loader` to ensure it's available when the environment is created. The provision environment must be configured explicitly, it will not inherit from `[testenv]`.

Fix #2862 

# Thanks for contribution

### DIFF
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -13,7 +13,6 @@ from tox.tox_env.api import ToxEnvCreateArgs
 from tox.tox_env.register import REGISTER
 from tox.tox_env.runner import RunToxEnv
 
-from ..config.loader.memory import MemoryLoader
 from ..config.types import EnvList
 from ..report import HandledError
 from ..tox_env.errors import Skip
@@ -134,7 +133,7 @@ class EnvSelector:
         self._manager = MANAGER
         self._log_handler = self._state._options.log_handler
         self._journal = self._state._journal
-        self._provision: None | tuple[bool, str, MemoryLoader] = None
+        self._provision: None | tuple[bool, str] = None
 
         self._state.conf.core.add_config("labels", Dict[str, EnvList], {}, "core labels")
         tox_env_filter_regex = getattr(state.conf.options, "skip_env", "").strip()
@@ -248,8 +247,6 @@ class EnvSelector:
             # ignore other envs when this is a provision run
             return None
         env_conf = self._state.conf.get_env(name, package=False)
-        if self._provision is not None and self._provision[1] == name:
-            env_conf.loaders.insert(0, self._provision[2])
         desc = "the tox execute used to evaluate this environment"
         env_conf.add_config(keys="runner", desc=desc, of_type=str, default=self._state.conf.options.default_runner)
         runner = REGISTER.runner(cast(str, env_conf["runner"]))
@@ -371,8 +368,8 @@ class EnvSelector:
         if invalid:
             raise HandledError(f"cannot run packaging environment(s) {','.join(invalid)}")
 
-    def _mark_provision(self, on: bool, provision_tox_env: str, loader: MemoryLoader) -> None:
-        self._provision = on, provision_tox_env, loader
+    def _mark_provision(self, on: bool, provision_tox_env: str) -> None:
+        self._provision = on, provision_tox_env
 
 
 __all__ = [
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -242,6 +242,10 @@ class EnvSelector:
 
     def _build_run_env(self, name: str) -> RunToxEnv | None:
         if self._provision is not None and self._provision[0] is False and name == self._provision[1]:
+            # ignore provision env unless this is a provision run
+            return None
+        if self._provision is not None and self._provision[0] and name != self._provision[1]:
+            # ignore other envs when this is a provision run
             return None
         env_conf = self._state.conf.get_env(name, package=False)
         if self._provision is not None and self._provision[1] == name:
--- a/docs/changelog/2862.bugfix.rst
+++ b/docs/changelog/2862.bugfix.rst
@@ -1,0 +1,4 @@
+The provision environment (``.tox``) will never inherit from ``testenv``.
+During provisioning, other test environments are not processed, allowing the
+use of keys and values that may be registered by later tox version or
+provisioned plugins - by :user:`masenf`.
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -317,3 +317,10 @@ version is above the version this feature was added to it, for example for setup
     [testenv:dev]
     deps = setuptools>=64
     package = editable
+
+Provisioning environment
+------------------------
+
+The provisioning environment is triggered when ``minversion`` or ``requires`` are specified and the current environment
+does not satisfy the requirement. In tox 4, the provisioning environment (``.tox`` by default) must be explicitly
+configured and will not inherit values from ``[testenv]`` section.
--- a/src/tox/provision.py
+++ b/src/tox/provision.py
@@ -98,7 +98,7 @@ def provision(state: State) -> int | bool:
     )
     provision_tox_env: str = state.conf.core["provision_tox_env"]
     state.conf.memory_seed_loaders[provision_tox_env].append(loader)
-    state.envs._mark_provision(bool(missing), provision_tox_env, loader)
+    state.envs._mark_provision(bool(missing), provision_tox_env)
 
     from tox.plugin.manager import MANAGER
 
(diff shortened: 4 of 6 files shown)


src/tox/session/env_select.py
```diff
@@ -241,13 +242,29 @@ def _finalize_config(self) -> None:
...
-        if self._provision is not None and self._provision[1] == name:
-            env_conf.loaders.insert(0, self._provision[2])
-        desc = "the tox execute used to evaluate this environment"
-        env_conf.add_config(keys="runner", desc=desc, of_type=str, default=self._state.conf.options.default_runner)
+        runner_key_desc = "the tox execute used to evaluate this environment"
+        if name == provision_tox_env:
+            if provision_loader is not None:
+                env_conf.loaders.insert(0, provision_loader)
+            # provision environment always runs with VirtualEnvRunner
+            env_conf.add_constant(keys="runner", desc=runner_key_desc, value=VirtualEnvRunner.id())
```
gaborbernat, line 260: Virtualenv should be the default but users should be able to override it within .tox testenv section.

src/tox/session/env_select.py
```diff
@@ -241,13 +242,29 @@ def _finalize_config(self) -> None:
         self._state.conf.core.mark_finalized()
 
     def _build_run_env(self, name: str) -> RunToxEnv | None:
-        if self._provision is not None and self._provision[0] is False and name == self._provision[1]:
+        provision_on = provision_tox_env = provision_loader = None
```
gaborbernat, line 245: I think we just need to change the base for the provision env.

gaborbernat (changes requested):

src/tox/session/env_select.py
masenf: not sure how to proceed on that one... but i'll think about it.

i'm off for the night here. it's been really fun hacking on `tox` today, thanks for the opportunity.

gaborbernat: See https://github.com/tox-dev/tox/blob/a2222e98b320467e821cf9444113e9b40ab3ad1b/src/tox/provision.py#L92; we should already not use inheritance from testenv here, we need to find out why it still happens. Likely a bug somewhere.

masenf: Right in the middle of that now. I _think_ it's related to caching the `"base"` value from the Ini before the provision loader is even added to `env_conf`.

```
> /Users/masen/code/tox-dev/tox/.tox/py311/lib/python3.11/site-packages/tox/session/env_select.py(259)_build_run_env()
-> env_conf.loaders.insert(0, provision_loader)
(Pdb) env_conf._defined["base"]
ConfigDynamicDefinition(keys=('base',), desc=inherit missing keys from these sections, of_type=typing.List[str], default=['testenv'], _cache=['testenv'])
(Pdb) env_conf.loaders
[IniLoader(section=testenv, overrides={})]
```

If I reset the `_cache` values for the following keys after adding the `provision_loader` at index 0, then they can get the new values from the `provision_loader`.

```
(Pdb) tuple(env_conf._defined)
('set_env', 'setenv', 'base')
```

Something like

```
for config_def in env_conf._defined:
    if isinstance(config_def, ConfigDynamicDefinition):
        config_def._cache = _PLACE_HOLDER
```

But this actually doesn't work, because the loader for the `testenv` section is already in the chain at this point; removing it expicitly would be messy.

--------------

So the approach I'm taking now is to update `Config.get_env` to accept a `base` parameter, and if that's passed through it overrides the defaults. I couldn't find a way to reliably NOT include the base via config only, since shimming in the `provision_loader` occurs after `Config.get_env`, so the `testenv` default base just gets baked into the env's `loaders`.

gaborbernat: I think instead insert we need to use https://github.com/tox-dev/tox/blob/main/src/tox/session/cmd/devenv.py#L38, as it's too late for inserting the loader 🤔

src/tox/session/env_select.py
```diff
@@ -241,13 +241,29 @@ def _finalize_config(self) -> None:
...
```
gaborbernat, line 244: I think this file needs to be untouched, only the provision.py should be changed.

gaborbernat (approved):

PR #2883, opened 2023-01-18
Add more explanation to `list_dependencies_command` configuration option

Czaki:
Whileing debugging different behavior of tests locally and on, CI I noticed thI, installed packages are listed on CI, and when running locally, they are not. 

It make it harder to use diff tools. I need to go through the tox code to understand why it happens. So it is valuable to add information about the condition required to run a command from `list_dependencies_command`


gaborbernat (approved):

gaborbernat (approved):

gaborbernat (changes requested):
You'll need to fix the CI.

Czaki:
I was creating this PR outside my main machine without pre-commit installed. Sorry for such stupid typos that block CI.

gaborbernat (approved):

gaborbernat:
No worries, all good.

PR #2885, opened 2023-01-20
Update infromation about `pip_pre` configuration options 

Czaki:
The docs for tox configuration still mention `--pre` flag even if tox uses `PIP_PRE` environment variables. This may lead to confusion, as in #2690. 
In this PR, I updated documentation to allow understanding of this change without looking for closed issues.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -726,8 +726,7 @@ Pip installer
 
    Determines the command used for installing packages into the virtual environment; both the package under test and its
    dependencies (defined with :ref:`deps`). Must contain the substitution key ``{packages}`` which will be replaced by
-   the package(s) to install.  You should also accept ``{opts}`` -- it will contain index server options such as
-   ``--pre`` (configured as ``pip_pre``).
+   the package(s) to install.  You should also accept ``{opts}`` -- it will contain index server options.
 
    .. note::
 
@@ -751,9 +750,10 @@ Pip installer
    :default: false
    :version_added: 1.9
 
-   If ``true``, adds ``--pre`` to the ``opts`` passed to :ref:`install_command`. This will cause it to install the
+   If ``true``, adds ``PIP_PRE=1`` to the `environment <https://pip.pypa.io/en/stable/topics/configuration/#environment-variables>`_. This will cause pip it to install the
    latest available pre-release of any dependencies without a specified version. If ``false``, pip will only install
-   final releases of unpinned dependencies.
+   final releases of unpinned dependencies. For other tool than pip there may be need to directly modify ``install_command``
+   to install prerelease packages version.
 
 User configuration
 ------------------
--- a/docs/changelog/2885.doc.rst
+++ b/docs/changelog/2885.doc.rst
@@ -1,0 +1,1 @@
+Update documentation of ``pip_pre`` to point to ``PIP_PRE`` environment variable.


docs/config.rst
```diff
@@ -726,8 +726,7 @@ Pip installer
 
    Determines the command used for installing packages into the virtual environment; both the package under test and its
    dependencies (defined with :ref:`deps`). Must contain the substitution key ``{packages}`` which will be replaced by
-   the package(s) to install.  You should also accept ``{opts}`` -- it will contain index server options such as
-   ``--pre`` (configured as ``pip_pre``).
+   the package(s) to install.  You should also accept ``{opts}`` -- it will contain index server options.
```
Czaki, line 729: As pre-release is configured now with the environment variable, I removed it from the example but do not have any idea what could be the replacement. 

As I understand, currently, all pip options could be configured using environment variables, so I'm not sure if `opts` is really required and maybe should be removed.

Czaki:
I cannot find information if pip 20.3.4 (last supporting python 2.7) supports configuration via environment variables.

gaborbernat (approved):

docs/config.rst
```diff
@@ -751,9 +750,10 @@ Pip installer
    :default: false
    :version_added: 1.9
 
-   If ``true``, adds ``--pre`` to the ``opts`` passed to :ref:`install_command`. This will cause it to install the
+   If ``true``, adds ``PIP_PRE=1`` to the `environment <https://pip.pypa.io/en/stable/topics/configuration/#environment-variables>`_. This will cause pip it to install the
```
masenf, line 753: This is not correct.

Setting `pip_pre = true` in a `[testenv]` _does_ cause tox to add `--pre` to the `opts`: https://github.com/tox-dev/tox/blob/997128ced47936adabb2f267d2606f8eac253ad3/src/tox/tox_env/python/pip/pip_install.py#L54-L67

The environment variable `PIP_PRE` is only set when using the _legacy_ CLI flag `--pre`:

https://github.com/tox-dev/tox/blob/4558fee5ac372c6842e5cb2ccd64c441d5a1aac5/src/tox/session/cmd/legacy.py#L60-L65

https://github.com/tox-dev/tox/blob/4558fee5ac372c6842e5cb2ccd64c441d5a1aac5/src/tox/session/cmd/legacy.py#L121-L122

masenf (changes requested):

docs/config.rst
Czaki: > The environment variable `PIP_PRE` is only set when using the _legacy_ CLI flag `--pre`:

Is there a plan to remove inconsistency? Or should I update the description in this PR to inform about this?

masenf: It's my understanding that the "pip flags" accepted by the tox CLI are deprecated and will be eventually removed. The pip environment variables should be used to influence pip instead of passing (some accepted subset) of pip flags to tox.

Basically tox has no reason to maintain these pip flags now that pip exposes all of its options via environment variable.

Czaki: Ok.
But I still need to figure out what this means for this PR. Should it be closed? Modified (ex. add an explanation to avoid confusion)?

masenf: @Czaki we have the following doc for the cli flag already: https://tox.wiki/en/latest/cli_interface.html#tox-legacy---pre

So I don't think this patch is strictly needed. However if you wanted to mention the legacy CLI flag in the testenv config section as "this config value may be overridden externally via the --pre cli flag (deprecated) or by setting PIP_PRE=1". Also note that setting PIP_PRE=0 does NOT override the testenv config, since that will use the pip cli flag, which takes precedence over envvars.

gaborbernat: If the user can set the pre flag, we should fix it so that it works. However, we should also document that this flag is deprecated and we encourage using the environment variables instead.

docs/config.rst
```diff
@@ -726,8 +726,7 @@ Pip installer
 
    Determines the command used for installing packages into the virtual environment; both the package under test and its
    dependencies (defined with :ref:`deps`). Must contain the substitution key ``{packages}`` which will be replaced by
-   the package(s) to install.  You should also accept ``{opts}`` -- it will contain index server options such as
```
MeCorey, line 729: - [ ] Thu, 26 Jan 2023 18:07:00 GMT  /usr/local/bin/git config --global --add safe.directory /Users/runner/work/tox/tox

MeCorey (approved):

gaborbernat:
Seems this has stalled, closing for now we can reopen when you get time for it again.

Czaki:
> Seems this has stalled, closing for now we can reopen when you get time for it again.

It is stalled because I'm confused what is proper solution. 

Did best option is to describe difference between legacy `--pre` flag and `pip_pre` configuration option, or add code changes to uniform behavior (also marking `pip_pre` as deprecated as it could be configured using `setenv` configuration option?

docs/config.rst
gaborbernat: See here

Czaki:
As I read `tox` source code, it looks like it should be handled in the same way as `setenv`, some variant of `EnvConfigSet` that will return `SetEnv` instance with `"PIP_PRE=1"` as input text. 

It is correct assumption? 

If it us correct then where I should register for such a class? Or could you point proper part of the documentation where I could read?

PR #2888, opened 2023-01-23
Enforce constraints during install_package_deps

### LINKED ISSUE #2386: Handle version conflicts between deps and package deps
When using tox to run tests for a Python package and installing testing dependencies through `deps`, there is potential for the package's `install_requires` dependency resolution to result in a package version that doesn't match the constraint in `deps`.

Because `deps` is explicitly defined, it feels like it should take precedence (get installed after the package?) or at the very least tox should warn about this situation.

I ran into this when testing a package against two Django versions:

```ini
# setup.cfg

...

[options]
install_requires =
    # a dependency that specifies Django>=1.11,<4

[tox:tox]
envlist = py{39,310}-django{3.2,4}
isolated_build = True

[testenv]
deps =
    pytest
    pytest-cov
    pytest-django
    pytest-randomly
    django3.2: Django>=3.2,<3.3
    django4: Django>=4.0,<5

...
```

The package's `install_requires` ultimately had an abstract dependency who needed `Django>=1.11,<4`, so even for the envs with the `django4` factor, `Django==3.2.12` was being installed. The only real knowledge of this fact was gotten by reading the output list of installed packages, and the lack of a failure I expected to see in Django 4.

I was almost inclined to call this a bug, but I have to imagine there are some design decisions at play here that I'm not aware of.

masenf:
Create a constraints file based on the listed or installed ``deps`` and use it during ``install_package_deps`` to ensure the package dependencies do not override the environment's dependencies.

Adds new config keys:
  * ``constrain_package_deps`` - chicken switch to go back to the legacy behavior of allowing the package deps to override the testenv deps
  * ``use_frozen_constraints`` - a more strict mode, enforced that the package deps do not conflict in any way with the set of packages installed by ``deps``

Fix #2386

# Thanks for contribution


gaborbernat (approved):

src/tox/tox_env/python/pip/pip_install.py
```diff
@@ -100,9 +133,20 @@ def _install_requirement_file(self, arguments: PythonDeps, section: str, of_type
                     missing_requirement = set(old["requirements"]) - set(new_requirements)
                     if missing_requirement:
                         raise Recreate(f"requirements removed: {' '.join(missing_requirement)}")
+                    if old.get("constraint_options") != constraint_options:
+                        raise Recreate(
+                            "constraint options changed: old={} new={}".format(
```
gaborbernat, line 138: why is this not an f-string?

src/tox/tox_env/python/pip/pip_install.py
```diff
@@ -100,9 +133,20 @@ def _install_requirement_file(self, arguments: PythonDeps, section: str, of_type
...
+                                old.get("constraint_options"),
+                                constraint_options,
+                            ),
+                        )
                 args = arguments.as_root_args
                 if args:  # pragma: no branch
                     self._execute_installer(args, of_type)
+                    if self.constrain_package_deps and not self.use_frozen_constraints:
+                        self.constraints_file().write_text(
+                            "\n".join(new_requirements + [c.lstrip("-c ") for c in new_constraints]),
```
gaborbernat, line 148: Can we keep this shorter by assigning the requirements to a variable?

src/tox/tox_env/python/pip/pip_install.py
masenf: mainly because i'm calling `old.get("constraint_options"),`, but let me do that on the line above.

gaborbernat: can still do that in f-string just use ' instead?

src/tox/tox_env/python/pip/pip_install.py
```diff
@@ -135,18 +135,14 @@ def _install_requirement_file(self, arguments: PythonDeps, section: str, of_type
                         raise Recreate(f"requirements removed: {' '.join(missing_requirement)}")
                     if old.get("constraint_options") != constraint_options:
                         raise Recreate(
```
gaborbernat, line 137:
```
msg =  f"constraint options changed: old={old.get('constraint_options')} new={constraint_options}"
raise Recreate(msg)
```
would save you a line here 🤔

masenf: i ended up making it a little cleaner by assigning `old_constraint_options = old.get("constraint_options")`

gaborbernat (approved):

PR #2902, opened 2023-01-26
docs/faq.rst: clarify how constrain_package_deps works

masenf:
Specify that certain types of deps only work with use_frozen_constraints.

Change the wording now that constrain_package_deps is not the default.

# Thanks for contribution


gaborbernat (approved):

docs/faq.rst
```diff
@@ -124,16 +124,16 @@ install. While creating a test environment tox will invoke pip multiple times, i
...
-dependency, an error will be raised.
-
-Ensure that ``constrain_package_deps = true`` is set in the test environment in order to use the constraints file
-generated by processing the ``deps`` section when performing ``package_deps``.
+Starting in tox 4.4.0, when ``constrain_package_deps = true`` is set in the test environment,
+``{env_dir}{/}constraints.txt`` will be generated during ``install_deps`` based on the package specifications listed
+under ``deps``. These constraints are subsequently passed to pip during the ``install_package_deps`` stage, causing an
+error to be raised when the package dependencies conflict with the test environment dependencies.
+
+For stronger guarantees, set ``use_frozen_constraints = true`` in the test environment to generate the constraints file
```
gaborbernat, line 132: Should one day we make this the default, we should probably do something like True if all in deps is valid package spec; otherwise False.

PR #2917, opened 2023-02-08
Ignore labels when provisioning

### LINKED ISSUE #2916: Labeled environments cannot use runners from provisioned plugins
## Issue

When a tox environment has a `runner` derived from a plugin that is named in the `requires` key, the environment cannot appear in a top-level label expression.

See minimal reproducer: https://github.com/masenf/tox-label-provision-plugin-runner

```ini
[tox]
requires = foo-runner
labels =
    foo = bar, baz
envlist = bar, baz

[testenv:{bar,baz}]
runner = foo
```

running `tox` uses the `envlist`, which works fine.

running `tox -m foo` uses the `label` foo, and blows up before passing off control to the provisioned tox

## Environment

Provide at least:

- OS: ubuntu 20.04

```console
cachetools==5.3.0
chardet==5.1.0
colorama==0.4.6
distlib==0.3.6
filelock==3.9.0
packaging==23.0
platformdirs==3.0.0
pluggy==1.0.0
pyproject-api==1.5.0
tomli==2.0.1
tox==4.4.4
virtualenv==20.19.0
```

## Output of running tox

Provide the output of `tox -rvv`:

```console
tox -rvv -m foo
ROOT: 1297 W will run in automatically provisioned tox, host /home/mfurer/.pytools/bin/python is missing [requires (has)]: foo-runner [tox/provision.py:125]
  File "/home/mfurer/.pytools/bin/tox", line 8, in <module>
    sys.exit(run())
  File "/home/mfurer/.pytools/lib/python3.8/site-packages/tox/run.py", line 19, in run
  File "/home/mfurer/.pytools/lib/python3.8/site-packages/tox/run.py", line 41, in main
    result = provision(state)
  File "/home/mfurer/.pytools/lib/python3.8/site-packages/tox/provision.py", line 126, in provision
  File "/home/mfurer/.pytools/lib/python3.8/site-packages/tox/provision.py", line 144, in run_provision
    tox_env: PythonRun = cast(PythonRun, state.envs[name])
  File "/home/mfurer/.pytools/lib/python3.8/site-packages/tox/session/env_select.py", line 337, in __getitem__
    return self._defined_envs[item].env
  File "/home/mfurer/.pytools/lib/python3.8/site-packages/tox/session/env_select.py", line 233, in _defined_envs
    self._mark_active()
  File "/home/mfurer/.pytools/lib/python3.8/site-packages/tox/session/env_select.py", line 321, in _mark_active
    self._defined_envs_[env_name].is_active = True
KeyError: 'bar'
```

## Minimal example

https://github.com/masenf/tox-label-provision-plugin-runner

masenf:
# Thanks for contribution

Ignore labels when tox will provision a runtime environment (``.tox``) so that environment configurations which depend on provisioned plugins or specific tox versions are not accessed in the outer tox process where the configuration would be invalid.

Extend existing [`test_provision_plugin_runner`](https://github.com/tox-dev/tox/commit/b26a1b95b9aa54d70ba9a1f5bd084a5fe70f08bc) to also run using labels, which catches this regression.

Fix #2916 

### DIFF
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -315,7 +315,8 @@ class EnvSelector:
         if labels or factors:
             for env_info in self._defined_envs_.values():
                 env_info.is_active = False  # if any was selected reset
-            if labels:
+            # ignore labels when provisioning will occur
+            if labels and (self._provision is None or not self._provision[0]):
                 for label in labels:
                     for env_name in self._state.conf.core["labels"].get(label, []):
                         self._defined_envs_[env_name].is_active = True
--- a/docs/changelog/2916.bugfix.rst
+++ b/docs/changelog/2916.bugfix.rst
@@ -1,0 +1,3 @@
+Ignore labels when tox will provision a runtime environment (``.tox``) so that environment configurations which depend
+on provisioned plugins or specific tox versions are not accessed in the outer tox process where the configuration would
+be invalid - by :user:`masenf`.
--- a/tests/test_provision.py
+++ b/tests/test_provision.py
@@ -195,14 +195,21 @@ def test_provision_no_recreate_json(tox_project: ToxProjectCreator) -> None:
 def test_provision_plugin_runner(tox_project: ToxProjectCreator, tmp_path: Path, plugin_testenv: str) -> None:
     """Ensure that testenv runner doesn't affect the provision env."""
     log = tmp_path / "out.log"
-    proj = tox_project({"tox.ini": f"[tox]\nrequires=demo-pkg-inline\n[{plugin_testenv}]\nrunner=example"})
-    result_first = proj.run("r", "-e", "py", "--result-json", str(log))
-    result_first.assert_success()
+    proj = tox_project(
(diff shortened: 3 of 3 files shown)


src/tox/session/env_select.py
```diff
@@ -315,7 +315,8 @@ def _mark_active(self) -> None:
         if labels or factors:
             for env_info in self._defined_envs_.values():
                 env_info.is_active = False  # if any was selected reset
-            if labels:
+            # ignore labels when provisioning will occur
+            if labels and (self._provision is None or not self._provision[0]):
```
gaborbernat, line 319: While at it should we do the same for factors?

masenf: for the factors piece, the code iterates through the environments, setting the attribute on the object(s) it gets back, so there's no risk of dereferencing an env name that doesn't exist. Above in `_build_env`, we're already NOT building the envs when provisioning is active, so the factors loop would be iterating through a mostly empty `self._defined_envs_`

gaborbernat (approved):

Issue #1035, opened 2018-10-05
provide Docker packaging


gaborbernat:
Beside providing tox as a Python package it could be helpful to also provide an "official" Docker image that's set up with correct Pythons and tox installed. This came out of https://github.com/tox-dev/tox/pull/943. Any version of this needs to:

- automate docker image update as part of the release process (Azure Pipelines has docker support),
- add image files,
- add tests (Azure Pipelines has docker support).

It's to be determined if this should leave as part of this repository; or should it instead be a separate project.

obestwalter:
Who would be the intended audience for this?

obestwalter:
That sound indeed handy.

thenewguy:
Did anything ever come of this?

thenewguy:
Sounds like something that would be accepted to the official docker library as well https://docs.docker.com/docker-hub/official_images/#creating-an-official-image

gaborbernat:
Would be great if someone could make a contribution for this.

31z4:
Hi guys, I'm happy to share an initial version of my [Docker image packaging for tox](https://github.com/31z4/tox-docker) 😊

This multi-arch Docker image neatly packages tox v4 along with all currently [active Python versions](https://devguide.python.org/versions/#status-of-python-versions). The image is secure, compact, and easy to use.

At this moment, the image only supports the following platforms:
* `linux/arm64/v8`
* `linux/amd64`

Please give it a try and share your feedback 🙏

I'm especially interested in discussing current image limitations and would love to hear from maintainers and users about the following:

* tox v3 is not supported. Does it make sense to have the v3 version of the image? For how long v3 will be maintained?
* It is not possible to install additional Python versions into a running container because the image doesn't include any build dependencies. I thought that for most cases it would be sufficient to have active CPython versions installed. Otherwise, it should be possible to build a custom image based on `31z4/tox`.
* For the same reason as above testing Python code that requires building C extensions is not supported.

As for the official Docker image, I could help with that too. I've been maintaining official images for [Zookeeper](http://hub.docker.com/_/zookeeper) and [Storm](http://hub.docker.com/_/storm) since 2016, so I have some relevant experience ☺️ Although, before contributing to the official docker library I'd love to gather some feedback from the community about the current version, fix discovered issues and address limitations if needed.

jugmac00:
toy v3 is no longer officially supported by the tox core team. So imho there is need for a docker image.

31z4:
[...]

PR #2923, opened 2023-02-10
Document running tox within a Docker container

31z4:
Created this PR to continue work on #1035. The idea is to mention `31z4/tox` Docker image in the docs to gather more feedback from users. What do you guys think? Is it OK to refer to the current version of the image? Or moving the image under the `tox-dev` organisation or making it [official](http://github.com/docker-library/official-images) would be a blocker?

# Thanks for contribution


docs/faq.rst
```diff
@@ -306,3 +306,45 @@ Access full logs
...
+    docker run -v `pwd`:/home/tox/tests -it --rm 31z4/tox run-parallel -e black,py311
+
+Note, that the image is configured with a working directory at ``/home/tox/tests``.
+
+If you want to install additional Python versions/implementations or Ubuntu packages you can create a derivative image.
+Just make sure you switch the user to ``root`` when needed and switch back to ``tox`` afterwards:
+
+.. code-block:: Dockerfile
+
+    FROM 31z4/tox
```
gaborbernat, line 339: Could we create a tox-dev or tox org instead to host under? You can be one of it's owner's 😊

docs/faq.rst
```diff
@@ -306,3 +306,45 @@ Access full logs
...
+
+.. code-block:: shell
+
+    docker run -v `pwd`:/home/tox/tests -it --rm 31z4/tox
+
+Because an entry point of the image is ``tox``, you can easily pass subcommands and flags:
+
+.. code-block:: shell
+
+    docker run -v `pwd`:/home/tox/tests -it --rm 31z4/tox run-parallel -e black,py311
```
gaborbernat, line 330: Would prefer to mount the project under the /w folder, personal preference 😅

docs/faq.rst
31z4: Yes I think it makes sense. Although, if we aim to make the image "official" maybe it's better to work towards this right now. This way we can save some maintenance efforts. Because we don't have to deprecate `tox-dev/tox` in favour of official `tox`. And also users don't have to migrate from the deprecated image to the official one.

Some benefits of an official image that I'd like to highlight:
* Automated multi-arch builds.
* Automated periodic re-builds when a base image is updated.
* Better structured docs and build artefacts (e.g.. https://github.com/docker-library/repo-info/tree/master/repos/zookeeper).

docs/faq.rst
31z4: That's possible too ☺️ But you have to specify the workdir in that case (`-w /w`):

    docker run -v `pwd`:/w -w /w -it --rm 31z4/tox run-parallel -e black,py311

I can change image workdir from `/home/tox/tests` to something more concise, like `/tests` maybe. I agree that `/home/tox/tests` might be too long to type 😅

gaborbernat: I'd make/w the default, concise better 😅

docs/faq.rst
gaborbernat: Yeah let's make it so. What do you need from me to get there?

31z4: I guess nothing at this point. I'll make some usability changes to the image to conform with https://github.com/docker-library/official-images#consistency. And then will work on PRs to https://github.com/docker-library/official-images and https://github.com/docker-library/docs.

docs/faq.rst
gaborbernat: @31z4 ping on this?

31z4: I've been pretty busy recently. Hopefully will get a chance to work on the official image later this month. In the meantime https://github.com/31z4/tox-docker should be good enough for production already.

Issue #2926, opened 2023-02-14
Plugins cannot use custom tox.ini config sections with a suffix name shared by a testenv


dcrosta:
## Issue

I noticed this with recent tox 4.x failing in CI for tox-docker plugin, which preivously worked.

The issue seems to be deep inside the config parsing code, if both a `[testenv:foo]` and `[docker:foo]` section exist with the same suffix ("foo"). Somewhere inside the parser code it seems to get confused and is looking for the docker plugin's keys within the testenv config section, and fails.

Here's a minimal(ish) script & tox.ini which reproduces it:

tox.ini
```ini
[testenv:foo]
custom = foo
command = python -c ""

[custom:foo]
bar = baz
```

tox_bug.py
```python
from typing import List
from tox.config.loader.section import Section
from tox.config.sets import ConfigSet
from tox.config.sets import EnvConfigSet
from tox.plugin import impl
from tox.session.state import State
from tox.tox_env.api import ToxEnv


def bar_required(bar: str) -> str:
    assert bar
    return bar


class CustomConfigSet(ConfigSet):
    def register_config(self) -> None:
        self.add_config(
            keys=["bar"],
            of_type=str,
            default="",
            desc="description",
            post_process=bar_required,
        )

def parse_custom_config(custom_conf: CustomConfigSet) -> None:
    print(custom_conf["bar"])


@impl
def tox_before_run_commands(tox_env: ToxEnv) -> None:
    custom_configs = tox_env.conf.load("custom")
    # breakpoint()
    for conf in custom_configs:
        parse_custom_config(conf)


@impl
def tox_add_env_config(env_conf: EnvConfigSet, state: State) -> None:
    def build_config_set(suffix: object) -> CustomConfigSet:
        assert isinstance(suffix, str)
        custom_conf = state.conf.get_section_config(
            section=Section("custom", suffix),
            base=[],
            of_type=CustomConfigSet,
            for_env=None,
        )
        if not custom_conf.loaders:
            raise ValueError(f"Missing [custom:{suffix}] in tox.ini")
        return custom_conf

    env_conf.add_config(
        keys=["custom"],
        of_type=List[CustomConfigSet],
        default=[],
        desc="description",
        factory=build_config_set,
    )
```

setup.py
```python
from setuptools import setup

setup(
    name="tox-bug",
    description="desc",
    maintainer="Dan Crosta",
    maintainer_email="dcrosta@example.com",
    py_modules=["tox_bug"],
[...]

PR #2927, opened 2023-02-21
Custom prefix in .ini source

masenf:
Fix bug #2926.

For plugin developers to have more flexibility when loading configuration from tox.ini using the mechanisms provided by tox.

As an implementation note: I've tried to keep the door open for future support of generative environments for custom prefixes (factors) as well. it's not publicly plumbed through at the moment, but in the future, if the `_section_mapping` contained custom keys (perhaps expanded according to plugin logic), then it could auto-dispatch to a base environment just like a generative `testenv` does.

# Thanks for contribution


gaborbernat (approved):

gaborbernat:
@masenf can you cut a release? Just run tox -e release -- version to trigger it. and then the CI will do the rest 😅 you can manually cut a GitHub release against the generated git tag as a final step.

masenf:
I need to learn to make releases sooner or later. Looks like you already handled 4.4.6 before I had a chance to pick it up. I'll try to get the next one, just to learn how it works.

PR #2944, opened 2023-03-14
Change stderr color to grey instead of red

ssbarnea:
As stderr is more use for logging than display errors, we better avoid using red and go for a neutral gray color.

This should avoid visual strain on users or confusions about presence of red colour on logs, which is almost always a sign of a important, usually fatal, error.

# Thanks for contribution

# Before

# After

### DIFF
--- a/src/tox/execute/api.py
+++ b/src/tox/execute/api.py
@@ -121,7 +121,11 @@ class Execute(ABC):
             # collector is what forwards the content from the file streams to the standard streams
             out, err = out_err[0].buffer, out_err[1].buffer
             out_sync = SyncWrite(out.name, out if show else None)
-            err_sync = SyncWrite(err.name, err if show else None, Fore.RED if self._colored else None)
+            err_sync = SyncWrite(
+                err.name,
+                err if show else None,
+                Fore.LIGHTBLACK_EX if self._colored else None,  # noqa: SC200
+            )
             with out_sync, err_sync:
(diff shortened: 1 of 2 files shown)


ssbarnea:
@gaborbernat I can open a poll about the subject, maybe we get other ideas about how we could about the extensive red output on screen, maybe other default color to use than the one I proposed and a lot more feedback from the users. WDYT?

gaborbernat:
I personally prefer the status quo but will listen to other maintainers opinion. The red does not bother me personally, and feels appropiate.

masenf:
I don't have strong opinions on the matter, but I agree it is nice to visually distinguish what is stdout and stderr. In my experience with test systems, I've mostly seen stderr colored red, or red background (in web views).

I'll try to broaden my perspective a bit and ask around my network see if anyone does have strong opinions and why.

jugmac00:
Thanks to the now red color for stderr of tox 4, I was made aware of many little issues, warnings, and so on, which were overlooked for years, but could lead to issues, like e.g.
Screenshot from 2023-03-15 14-46-37

I am +1 for keeping the red color, and I am avers to change, especially after the recent-ish tox 4 release. We had years of alpha releases for people to give feedback. Now I am happy that things have settled.

@ssbarnea I think you are aware of `mkdocs -q`?

Instead of changing the color for all, would it make sense to make the color configurable? Not that I am a fan of too much customization, but that would be a reasonable compromise.

Issue #2933, opened 2023-03-04
Modifying `tox_root` does not propagate to dependent fields.


kdestin:
## Issue

Changing the value of `tox_root` does not propagate to fields whose default value also depends on `tox_root` (i.e `work_dir`, `temp_dir`, `env_dir`, etc...). 

This reproducible in tox `4.4.6` and in the latest commit (which is https://github.com/tox-dev/tox/commit/177a227889a93416bd154938361977486d14b0bb at time of writing) when trying to modify `tox_root` using the CLI interface (e.g `tox --root ~`).

I assume this would also be the case when using a plugin to modify tox_root, but the following (using an approach based on this [discussion](https://github.com/tox-dev/tox/discussions/2882#discussioncomment-4713455)) doesn't even change the value of `tox_root`, so I assume that `tox_root` might be getting cached before this hook can run:

```python
@impl
def tox_add_core_config(core_conf: CoreConfigSet, state: State):
    core_conf.loaders.insert(0, MemoryLoader(tox_root=Path.home())
```

## Environment

Provide at least:

- OS:
```console
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy
```
- `pip list` of the host Python where `tox` is installed:

```console
Package               Version
--------------------- -------
cachetools            5.3.0
chardet               5.1.0
colorama              0.4.6
distlib               0.3.6
docstring-to-markdown 0.11
filelock              3.9.0
jedi                  0.18.2
packaging             23.0
parso                 0.8.3
pip                   22.0.2
platformdirs          3.0.0
pluggy                1.0.0
pylsp-rope            0.1.11
pyproject_api         1.5.0
python-lsp-jsonrpc    1.0.0
python-lsp-server     1.7.1
pytoolconfig          1.2.5
rope                  1.7.0
setuptools            59.6.0
tomli                 2.0.1
tox                   4.4.6
ujson                 5.7.0
virtualenv            20.20.0

```

## Output of running tox

Provide the output of `tox -rvv`:

```console
py: 603 W remove tox env folder /home/kd/Documents/tox/repro/.tox/py [tox/tox_env/api.py:322]
py: 683 I find interpreter for spec PythonSpec(path=/home/kd/Documents/tox/venv/bin/python3) [virtualenv/discovery/builtin.py:56]
[...]

PR #2962, opened 2023-03-29
Fix `tox_root` propagation to `work_dir`

Tbruno25:
This applies a fix for issue #2933 where `work_dir` was not accurately including the current `tox_root` during test runs

# Thanks for contribution

### DIFF
--- a/docs/changelog/2933.bugfix.rst
+++ b/docs/changelog/2933.bugfix.rst
@@ -1,1 +1,1 @@
-Fix issue where ``work_dir`` was not correctly including ``tox_root`` for test runs
+Fix issue where ``work_dir`` was not correctly including ``tox_root`` for test runs.
--- a/docs/changelog/2933.bugfix.rst
+++ b/docs/changelog/2933.bugfix.rst
@@ -1,0 +1,1 @@
+Fix issue where ``work_dir`` was not correctly including ``tox_root`` for test runs
--- a/src/tox/config/sets.py
+++ b/src/tox/config/sets.py
@@ -188,23 +188,17 @@ class CoreConfigSet(ConfigSet):
             desc="the root directory (where the configuration file is found)",
         )
 
-        def work_dir_builder(conf: Config, env_name: str | None) -> Path:  # noqa: U100
-            return (conf.work_dir if conf.work_dir is not None else cast(Path, self["tox_root"])) / ".tox"
-
-        def work_dir_post_process(value: Path) -> Path:
-            return self._conf.work_dir if self._conf.options.work_dir else value
-
         self.add_config(
             keys=["work_dir", "toxworkdir"],
             of_type=Path,
-            default=work_dir_builder,
-            post_process=work_dir_post_process,
+            default=lambda *_: self["tox_root"] / ".tox",  # noqa: U101
+            post_process=lambda dir: self._conf.work_dir if self._conf.options.work_dir else dir,
             desc="working directory",
         )
         self.add_config(
             keys=["temp_dir"],
             of_type=Path,
-            default=lambda conf, _: cast(Path, self["work_dir"]) / ".tmp",  # noqa: U100, U101
+            default=lambda *_: self["work_dir"] / ".tmp",  # noqa: U101
             desc="a folder for temporary files (is not cleaned at start)",
         )
         self.add_constant("host_python", "the host python executable path", sys.executable)
--- a/src/tox/config/sets.py
+++ b/src/tox/config/sets.py
@@ -179,6 +179,15 @@ class CoreConfigSet(ConfigSet):
         desc = "define environments to automatically run"
         self.add_config(keys=["env_list", "envlist"], of_type=EnvList, default=EnvList([]), desc=desc)
 
+    def _default_work_dir(self, conf: Config, env_name: str | None) -> Path:  # noqa: U100
+        return cast(Path, self["tox_root"] / ".tox")
+
+    def _default_temp_dir(self, conf: Config, env_name: str | None) -> Path:  # noqa: U100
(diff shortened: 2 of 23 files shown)


gaborbernat (changes requested):
CI is failing.

Tbruno25:
> Let me know if you'd like me to open a PR onto your branch, to incorporate these changes.

Let's merge it in!
 
>   I may be wrong, but my understanding of the control flow is that the `conf.work_dir` can never actually be `None`. In `Config.make`, it's always set to a non-`None` value:
> ...
>   * The default case gets handled by returning `case(Path, self["tox_root"])) / ".tox"`
>   * The case when `work_dir` is set by `--workdir` is handled here:

Agree with your conclusion.

>   https://github.com/tox-dev/tox/blob/a4be5d8720d2976a9e8b59ee2c0379db355b60ea/src/tox/config/sets.py#L196-L197
> * I wrote tests in `tests/test_run.py` (instead of `tests/test_main.py`), since there was already a test for `--workdir` there which would've been similar to the test of `--root`.
> * I believe I have some fixes for some of your test failures

I was _really_ hoping to avoid altering existing tests which is why I was "guess and checking" with the different fixes. If you've already re-worked these then even more reason to pull it in :+1:

Tbruno25:
> CI is failing.

Should be all set thanks to @kdestin :rocket:

src/tox/config/sets.py
```diff
@@ -188,23 +188,17 @@ def register_config(self) -> None:
...
-
-        def work_dir_post_process(value: Path) -> Path:
-            return self._conf.work_dir if self._conf.options.work_dir else value
-
         self.add_config(
             keys=["work_dir", "toxworkdir"],
             of_type=Path,
-            default=work_dir_builder,
-            post_process=work_dir_post_process,
+            default=lambda *_: self["tox_root"] / ".tox",  # noqa: U101
```
gaborbernat, line 194: Why are we using here the * operator and not enumerating the argumetns? Why is this now a lambda and not a method as before? I prefered the previous way.

src/tox/config/sets.py
```diff
@@ -188,23 +188,17 @@ def register_config(self) -> None:
...
+            post_process=lambda dir: self._conf.work_dir if self._conf.options.work_dir else dir,
             desc="working directory",
         )
         self.add_config(
             keys=["temp_dir"],
             of_type=Path,
-            default=lambda conf, _: cast(Path, self["work_dir"]) / ".tmp",  # noqa: U100, U101
+            default=lambda *_: self["work_dir"] / ".tmp",  # noqa: U101
```
gaborbernat, line 201: Same.

tests/tox_env/python/test_python_runner.py
```diff
@@ -95,7 +95,16 @@ def test_journal_package_dir(tmp_path: Path) -> None:
 
 def test_package_temp_dir_view(tox_project: ToxProjectCreator, demo_pkg_inline: Path) -> None:
     project = tox_project({"tox.ini": "[testenv]\npackage=wheel"})
-    result = project.run("r", "-vv", "-e", "py", "--root", str(demo_pkg_inline))
+    result = project.run(
```
gaborbernat, line 98: This seems an unrelated test, why did we change this?

gaborbernat (changes requested):

src/tox/config/sets.py
Tbruno25: The args have no impact on what the function will return. I opted for the star as it allowed a single `noqa` code instead of multiple. 

And a lambda was utilized in the  `temp_dir` config already. I have no strong opinion on lambda/funcs but I think the implementation should be consistent either way.

Will revert!

tests/tox_env/python/test_python_runner.py
Tbruno25: @kdestin may be able to speak to this better than I

kdestin: This test seems to have depended on the bug we're fixing. Adding `"--workdir", str(project.path / ".tox")` preserves the expected state of the tox invocation.

src/tox/config/sets.py
```diff
@@ -179,6 +179,15 @@ def __init__(self, conf: Config, section: Section, root: Path, src_path: Path) -
         desc = "define environments to automatically run"
         self.add_config(keys=["env_list", "envlist"], of_type=EnvList, default=EnvList([]), desc=desc)
 
+    def _default_work_dir(self, conf: Config, env_name: str | None) -> Path:  # noqa: U100
+        return cast(Path, self["tox_root"] / ".tox")
+
+    def _default_temp_dir(self, conf: Config, env_name: str | None) -> Path:  # noqa: U100
+        return cast(Path, self["work_dir"] / ".tmp")
+
+    def _work_dir_post_process(self, dir: Path) -> Path:
```
gaborbernat, line 188: Why move these to class methods from before being inlined methods?

docs/changelog/2933.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix issue where ``work_dir`` was not correctly including ``tox_root`` for test runs
```
gaborbernat, line 1: trailing period, to finish your sentence and be grammatically correct

src/tox/config/sets.py
Tbruno25: I tend to try and avoid nested functions for a variety of reasons. By defining them as methods it improves testability while simultaneously making `register_config` shorter and easier to digest at first glance.

It's more idiomatic imo -- of course I'm happy to revert if you feel differently!

gaborbernat (approved):

PR #2970, opened 2023-04-03
Avoid UnicodeDecodeError from command output

masenf:
Instead of raising ``UnicodeDecodeError`` when command output includes non-utf-8 bytes,
``tox`` will now use ``surrogateescape`` error handling to convert the unrecognized bytes
to escape sequences according to PEP-383.

Tox has no way of knowing that the bytestream emitted by a command will be valid utf-8, even if utf-8 is ostensibly the "correct" encoding for the stream or system. It's always possible for an arbitrary command to return non-utf-8 bytes, and this situation should not break tox.

# Thanks for contribution


masenf:
@gaborbernat @jugmac00 do either of you have an opinion on this change?

gaborbernat (approved):

PR #2988, opened 2023-04-19
git: Ignore the .lock file for demo_pkg_inline

hroncok:
I hope this can go in without a changelog entry.


gaborbernat (approved):

gaborbernat:
> I hope this can go in without a changelog entry.

Ok but release will need to wait for something that does generates a changelog.

hroncok:
IMHO there's no need for a release with just this anyway

PR #2995, opened 2023-04-26
Specify an environment in the release workflow

webknjaz:
$sbj

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,9 @@ on:
 jobs:
   release:
     runs-on: ubuntu-22.04
+    environment:
+      name: release
+      url: https://pypi.org/p/tox
     permissions:
       id-token: write
     steps:


.github/workflows/release.yml
```diff
@@ -6,6 +6,9 @@ on:
 jobs:
   release:
     runs-on: ubuntu-22.04
+    environment:
+      name: release
+      url: https://pypi.org/p/tox
```
gaborbernat, line 11: Can you explain this line? Why we need it?

webknjaz: This will show up in UI, also you can apply protections in the repo settings, once it appears. I used to scope secrets to environments, instead of global settings. You can also require human approvals. All's optional.

webknjaz: It's optional but is required when you fill out the field on the PyPI side.

webknjaz: If you don't want the `url:` line, you can also just have `environment: release`.

gaborbernat: It's fine, but I'll need to create my own PR tomorrow because it doesn't let me merge this. Apparently you don't have the necessary rights to make this change. Or so GitHub says. ...

webknjaz: Oh, are you using the GitHub (mobile?) app? That thing has a limited token IIRC — using a web UI should help in such a case.

PR #3019, opened 2023-05-27
Fix legacy devenv

gaborbernat:


gaborbernat (commented):
@asottile upon reflection you were right 👍

gaborbernat:
https://pypi.org/project/tox/4.5.2/ should be good

PR #3024, opened 2023-06-03
Add --list-dependencies options

### LINKED ISSUE #3023: Allow CI runs to disable listing dependencies

Right now CI runs always list dependencies for each environment. While this is often nice, it can result in excessive output, especially when using tox to run linters and other tools.

If `list_dependencies_command` is explicitly set as an empty string for an environment, skip the listing dependencies command.

## Alternative Solutions

One alternative would be a separate configuration option for an environment to specifically disable listing dependencies.

## Additional context

snapcraft and several other projects use tox to run our linters. However, when running in CI, it can be difficult to find specific linting failures between all the `pip freeze` lines:

image

lengau:
This pair of options allows a user to either explicitly list dependencies after setup or explicitly silence listing of dependencies if it would be automatically done.

# Thanks for contribution

- [ ] updated/extended the documentation (N/A: documentation changes are automatically generated)

Fixes #3023

### DIFF
--- a/src/tox/session/cmd/run/common.py
+++ b/src/tox/session/cmd/run/common.py
@@ -70,6 +70,19 @@ def env_run_create_flags(parser: ArgumentParser, mode: str) -> None:
             default=None,
             help="write a JSON file with detailed information about all commands and results involved",
         )
+        list_deps = parser.add_mutually_exclusive_group()
+        list_deps.add_argument(
+            "--list-dependencies",
+            action="store_true",
+            default=is_ci(),
+            help="list the dependencies installed during environment setup",
+        )
+        list_deps.add_argument(
+            "--no-list-dependencies",
+            action="store_false",
+            dest="list_dependencies",
+            help="never list the dependencies installed during environment setup",
+        )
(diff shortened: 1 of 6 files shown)


src/tox/session/cmd/run/common.py
```diff
@@ -163,6 +163,19 @@ def __call__(
             help="skip package installation for this run",
             action="store_true",
         )
+        list_deps = parser.add_mutually_exclusive_group()
```
gaborbernat, line 166: I'd like the is_ci to be the default selection and be moved here.

gaborbernat (changes requested):

src/tox/session/cmd/run/common.py
lengau: That seems a bit counter-intuitive to me, but I've done so :smiley:

gaborbernat: I just want to avoid the surprise that the user set --no-list-dependencies and then CI=1, or other way around and things don't work as expected 😊

gaborbernat (approved):

Issue #2942, opened 2023-03-13
--help output embeds randomly generated default, making docs non-reproducible


paravoid:
## Issue

The `tox legacy --help` output changes on every invocation because of the `--hashseed` default:
```python
parser.add_argument(
    "--hashseed",
    metavar="SEED",
    help="set PYTHONHASHSEED to SEED before running commands. Defaults to a random integer in the range "
    "[1, 4294967295] ([1, 1024] on Windows). Passing 'noset' suppresses this behavior.",    
    action=SeedAction,
    of_type=Optional[int],
    default=random.randint(1, 1024 if sys.platform == "win32" else 4294967295),
    dest="hash_seed",
)
```

Besides the lack of usefulness (or lie, depending on your viewpoint ;) on what the default for this parameter is, the real-world impact of this is that this gets included in the generated Sphinx documentation (by the way of argparse-cli), which makes the build non-reproducible: multiple runs of documentation building generate slightly different output.

## Environment

This is with tox 4.4.6, but I've verified the same applies to current master. This is on a Debian sid system, with (my) Debian packages of tox, but I don't think that makes a difference here.

## Minimal example

```console
$ tox legacy --help | sed -nr 's/(.+--hashseed SEED).+\. (.+)/\1 ... \2/p'
  --hashseed SEED ... (default: 499015295)
$ tox legacy --help | sed -nr 's/(.+--hashseed SEED).+\. (.+)/\1 ... \2/p'
  --hashseed SEED ... (default: 1099173378)
```

paravoid:
I can and that would fix my use case.

My thinking for filing this as an upstream bug though was that this the help output is somewhat misleading: the default isn't "1099173378", it's "<some random number>". Not sure if anyone is not going to be confused in practice though.

gaborbernat:
But that's the default for that run 🤔 so it is accurate, it just changes per run. Not sure what would be a better solution.

paravoid:
Maybe I'm misunderstanding something... I expect what I see in the `--help` output to be useful for the subsequent "run" :) Basically the `--help` output is not a "run", it's just help output.

Or to put it another way: https://tox.wiki/en/latest/cli_interface.html right now says: "default: 17441603". That's not the default! The default is a random value (as the same line says earlier).

[...]

PR #3076, opened 2023-08-03
Make --hashseed default to PYTHONHASHSEED, if defined

paravoid:
The main motivation for this is to able to set the hash seed when building the documentation with "tox -e docs", and thus avoid embedding a random value in the tox documentation for --help. This caused documentation builds to fail to
build reproducibly.

Fixes issue #2942.


paravoid:
> Tests please +1

Done now. Thanks for the review!

tests/tox_env/python/test_python_api.py
```diff
@@ -220,6 +221,38 @@ def test_python_set_hash_seed_incorrect(tox_project: ToxProjectCreator) -> None:
     assert "tox run: error: argument --hashseed: invalid literal for int() with base 10: 'ok'" in result.err
 
 
+def test_python_use_hash_seed_from_env(tox_project: ToxProjectCreator) -> None:
+    ini = "[testenv]\npackage=skip\ncommands=python -c 'import os; print(os.environ[\"PYTHONHASHSEED\"])'"
+    with patch.dict("os.environ", {"PYTHONHASHSEED": "13"}):
+        result = tox_project({"tox.ini": ini}).run("r", "-e", "py")
```
gaborbernat, line 227: To test this change you don't need to run, please use `tox c -e py -k set_env` as a better proxy 👍 and faster tests.

gaborbernat (changes requested):

tests/tox_env/python/test_python_api.py
paravoid: I tried to follow the existing practice, and all 6 tests above it do it in this way. Am I missing something?

gaborbernat: Those validate that the environment variables are forwarded correctly, with these extra tests is enough to validate that set_env is set correctly.

gaborbernat:
@paravoid how would this differ from:

```
env TOX_HASH_SEED=1 tox c -e py -k set_env
```

Would do the same, no?

gaborbernat (approved):

paravoid:
> @paravoid how would this differ from:
> 
> ```
> env TOX_HASH_SEED=1 tox c -e py -k set_env
> ```
> 
> Would do the same, no?

Apparently yes, and indeed can resolve the issue with reproducible docs generation... I didn't know that, and it didn't come up on #2942, so I went down this path instead.

The PYTHONHASHSEED change is kind of valid on its own, but it's also new code that may not be that useful given this information. Leaving it up to you on whether you want to merge it or not. No hard feelings if not!

Issue #3087, opened 2023-08-11
Allow extending lists in --override


stefanor:
## What's the problem this feature will solve?

In Debian package building, we like to run automated test suites when they are available. Many projects use tox, so we try our best to support it.

One of the environment variables we export during many package builds (`_PYTHON_HOST_PLATFORM`) affects the installation tooling. If a wheel is built with one `_PYTHON_HOST_PLATFORM`, it won't install with a different / missing value. Tox filters environment variables that are passed to the build. So our wheel that we install into the virtualenv will inherit this environment variable. But when it tries to install the package, the environment variables are filtered, and it fails.

E.g.:
```
ERROR: foo-0.0.0-cp311-cp311-linux_amd64.whl is not a supported wheel on this platform.
```

The most obvious solution here is to add `_PYTHON_HOST_PLATFORM` to `passenv`. But that means editing the `tox.ini`/`pyproject.toml`/`setup.cfg` from the tool that's driving tox. That's far from ideal. So, I'd like to override it from the command line. But all I can do is blow away any existing `passenv`, I can't append to it. Most of the time that's good enough, but some packages *need* specific environment variables to be passed through tox.

In some ways tox isn't a great fit for Debian. tox works with virtualenvs and downloads things from PyPI where Debian maintains a self-contained distribution and requires no Internet access from package builds. So, we drive tox in some unusual ways, but it works. Currently this is `--sitepackages` and `--installpkg` / `--skip-pkg-install` (as appropriate). It's very convenient to be able to use the same tool that the upstream does, to run tests.

PS: Why do we export `_PYTHON_HOST_PLATFORM`? Because on some architectures it varies between otherwise-compatible machines. So, to make the builds (and debugging symbols) reproducible we set it to what we think the correct value for the platform is. This used to often be rather weird values, but after running into this bug, I made it be right most of the time, I think.

## Describe the solution you'd like

Something like: `--override testenv.passenv+=_PYTHON_HOST_PLATFORM`

## Alternative Solutions

I could possibly imagine doing this via a tox plugin. But I can't see an obvious way to have the plugin execute in package builds and not with any other use of tox.

gaborbernat:
PR welcome 👍

gaborbernat:
> In some ways tox isn't a great fit for Debian. tox works with virtualenvs and downloads things from PyPI where Debian maintains a self-contained distribution and requires no Internet access from package builds.

[...]

PR #3088, opened 2023-08-11
Allow extending lists with --override foo+=bar

stefanor:
Allow appending to a list with `+=` syntax, instead of replacing the existing value.

Fixes: #3087 

### DIFF
--- a/tests/config/test_main.py
+++ b/tests/config/test_main.py
@@ -66,7 +66,7 @@ def test_config_override_wins_memory_loader(tox_ini_conf: ToxIniCreator) -> None
     assert conf["c"] == "ok"
 
 
-def test_config_override_appends(tox_ini_conf: ToxIniCreator) -> None:
+def test_config_override_appends_to_list(tox_ini_conf: ToxIniCreator) -> None:
     example = """
     [testenv]
     passenv = foo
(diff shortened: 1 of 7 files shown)


gaborbernat (changes requested):
Please add documentation in here https://tox.wiki/en/4.7.0/user_guide.html

stefanor:
> Please add documentation in here https://tox.wiki/en/4.7.0/user_guide.html

There isn't currently any documentation for most command line arguments in there. I went for the config page instead.

stefanor:
That test failure looks unrelated.

stefanor:
> That test failure looks unrelated.

Yep, a noop change passed.

tests/config/test_main.py
```diff
@@ -64,6 +66,27 @@ def test_config_override_wins_memory_loader(tox_ini_conf: ToxIniCreator) -> None
...
+
+
+def test_config_override_cannot_append(tox_ini_conf: ToxIniCreator) -> None:
+    example = """
+    [testenv]
+    foo = 1
+    """
+    conf = tox_ini_conf(example, override=[Override("testenv.foo+=2")]).get_env("testenv")
+    conf.add_config("foo", of_type=int, default=0, desc="desc")
+    with pytest.raises(ValueError, match="Only able to append to lists"):
```
gaborbernat, line 86: Shouldn't we be able to also append to dictionaries? This would allow, for example to add to the setenv.

stefanor: Ah, yep, we can handle that.

gaborbernat (approved):

Issue #2858, opened 2023-01-13
Using non-existent environment does not fail


Mr-Pepe:
## Issue

Using an environment that does not exist still executes without failure but does nothing useful.

## Environment

Provide at least:

- OS: Linux
- `pip list` of the host Python where `tox` is installed:

```console
Package       Version
------------- -------
cachetools    5.2.1
chardet       5.1.0
colorama      0.4.6
distlib       0.3.6
filelock      3.9.0
packaging     23.0
pip           22.3.1
platformdirs  2.6.2
pluggy        1.0.0
pyproject_api 1.4.0
setuptools    65.7.0
tomli         2.0.1
tox           4.2.8
virtualenv    20.17.1
wheel         0.38.4

```

## Minimal example

If possible, provide a minimal reproducer for the issue:

```console
tox run -e abcdefg
```

jugmac00:
Looks like this is a regression as with tox v3...
```
❯ tox -e abcdefgh
ERROR: unknown environment 'abcdefgh'
```

gaborbernat:
This is expected 😊 and feature with tox 4.

gaborbernat:
This is expected 😊 and feature with tox 4.

jugmac00:
I was not specific enough. Above output is from tox version 3.

tox version 4 falsly runs tests for an unknown env.

So...
tox3 -> shows error message that env is unknown
tox4 -> runs tests for the version tox was installed with

This is not intended, right?

gaborbernat:
It is intended.

jugmac00:
While I am happy to allow this for e.g. `tox -e py313`, but this is dangerous for e.g. `tox -e lynt` (ie with a typo).

gaborbernat:
There's no easy way to allow just one and not the other. I am mainly considering factors like a-py313-ok. But feel free to put in a PR to try.

Mr-Pepe:
I bumped into this issue when I removed an environment from the `tox.ini` but did not remove the call from my pipeline script. I would have expected the pipeline to fail and was really confused when it didn't. Is this behavior documented?

gaborbernat:
It's undocumented in either direction.

bmw:
This may not matter unless this issue is worked on, but I'm even seeing some strange behavior here with tox 3.28.0. For this tox.ini file:
```
[tox]
skipsdist = true

[testenv]
allowlist_externals = echo
commands =
    echo do pytest things

[testenv:coverage]
commands =
[...]

PR #3089, opened 2023-08-13
Disallow command line environments which are not explicitly specified in the config file

tjsmart:
Checks that all command line provided environments correspond with an environment in the configuration file.

This does not apply to configurations starting with `py*` or `.pkg` (I don't know what the latter is perhaps it shouldn't be special cased).

I'd love to update/extend the documentation but could use some help finding a home for this.

Fixes: #2858. 

### DIFF
--- a/tests/session/cmd/test_parallel.py
+++ b/tests/session/cmd/test_parallel.py
@@ -141,7 +141,7 @@ def test_keyboard_interrupt(tox_project: ToxProjectCreator, demo_pkg_inline: Pat
     )
     cmd = ["-c", str(proj.path / "tox.ini"), "p", "-p", "1", "-e", f"py,py{sys.version_info[0]},dep"]
     process = Popen([sys.executable, "-m", "tox", *cmd], stdout=PIPE, stderr=PIPE, universal_newlines=True)
-    while not marker.exists():
+    while not marker.exists() and (process.poll() is None):
         sleep(0.05)
     process.send_signal(SIGINT)
     out, err = process.communicate()
--- a/tests/session/test_env_select.py
+++ b/tests/session/test_env_select.py
@@ -132,3 +132,35 @@ def test_env_select_lazily_looks_at_envs() -> None:
     # late-assigning env should be reflected in env_selector
     state.conf.options.env = CliEnv("py")
     assert set(env_selector.iter()) == {"py"}
+
+
+def test_cli_env_can_be_specified_in_default(tox_project: ToxProjectCreator) -> None:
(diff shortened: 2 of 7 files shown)


tests/session/cmd/test_parallel.py
```diff
@@ -141,7 +141,7 @@ def test_keyboard_interrupt(tox_project: ToxProjectCreator, demo_pkg_inline: Pat
     )
     cmd = ["-c", str(proj.path / "tox.ini"), "p", "-p", "1", "-e", f"py,py{sys.version_info[0]},dep"]
     process = Popen([sys.executable, "-m", "tox", *cmd], stdout=PIPE, stderr=PIPE, universal_newlines=True)
-    while not marker.exists():
+    while not marker.exists() and (process.poll() is None):
```
tjsmart, line 144: While in a bit of broken state, this test was hanging for me. Suggesting that we use [`Popen.poll`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll) to check if the child process has already terminated (likely with an error if marker does not exist).

gaborbernat (approved):

gaborbernat (changes requested):
Generraly the idea looks good, but you need to fix the CI.

gaborbernat (approved):

atugushev:
@tjsmart FYI, I've reported a bug https://github.com/tox-dev/tox/issues/3096, which might be related to this PR.

tests/session/test_env_select.py
```diff
@@ -132,3 +132,35 @@ def test_env_select_lazily_looks_at_envs() -> None:
...
+
+
+def test_cli_env_not_in_tox_config_fails(tox_project: ToxProjectCreator) -> None:
+    proj = tox_project({"tox.ini": ""})
+    outcome = proj.run("r", "-e", "does_not_exist")
+    outcome.assert_failed(code=-2)
+    assert "provided environments not found in configuration file: ['does_not_exist']" in outcome.out, outcome.out
+
+
+@pytest.mark.parametrize("env_name", ["py", "py310", ".pkg"])
```
hroncok, line 160: `test_allowed_implicit_cli_envs[py310]` now fails when Python 3.10 is not installed. Is that expected, or should I open an issue?

gaborbernat: PR welcome, likely should be pyx.y where x.y is the current Python interpreter.

hroncok: https://github.com/tox-dev/tox/pull/3108

Issue #3080, opened 2023-08-04
PEP 517 config settings support


nschloe:
In https://github.com/tox-dev/tox/issues/573, support for [PEP 517](https://peps.python.org/pep-0517/) was added. From what I can tell, [config settings](https://peps.python.org/pep-0517/#config-settings) support is still missing. Config settings are a dict that is passed to the build backend, e.g., setuptools. Pip implements this as [`--config-settings`/`-C`](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-C), [pypa-build](https://github.com/pypa/build) as [`--config-setting`/`-C`](https://pypa-build.readthedocs.io/en/stable/#python--m-build---config-setting).

It'd be great if tox had some way of specifying config settings that a passed to the build backend, e.g.,
```ini
[testenv:.pkg]
config-settings = 
  foo = bar
  alpha = omega
  run = 256
```

gaborbernat:
PR welcome 👍

nschloe:
In tox's output, I see that `'config_settings'` exist as an entity:
```
[...]
.pkg: 685 W build_sdist> python /home/nschloe/venv/lib/python3.11/site-packages/pyproject_api/_backend.py True foobar.build_meta [tox/tox_env/api.py:427]
Backend: run command build_sdist with args {'sdist_directory': '/path/to/.tox/.pkg/dist', 'config_settings': None}
```
I suppose this comes from [pyproject-api](https://github.com/tox-dev/pyproject-api). I can't quite see right now how this is called.

nschloe:
I've found where `config_settings` need to be added, but I have difficulties understanding how tox settings are defined and passed around. Any help?

gaborbernat:
https://tox.wiki/en/4.6.4/plugins.html#tox.plugin.spec.tox_add_env_config, you should have plenty examples in the code base, check for example https://tox.wiki/en/4.6.4/config.html#package_root-env

gaborbernat:
Done.

PR #3090, opened 2023-08-14
Add config_settings support for build backend calls

nschloe:
This is a first (non-working) stab at https://github.com/tox-dev/tox/issues/3080. I tried to add a config setting just like https://github.com/tox-dev/tox/pull/1891, but it doesn't seem to get passed on to `self.core`. It's supposed to be used like

```ini
[testenv]
config_settings = 
  abc = xyz
```

Any hints on what's missing?


src/tox/tox_env/python/virtual_env/package/pyproject.py
```diff
@@ -214,7 +214,10 @@ def perform_packaging(self, for_env: EnvConfigSet) -> list[Package]:
         elif of_type == "sdist":
             self.setup()
             with self._pkg_lock:
-                sdist = self._frontend.build_sdist(sdist_directory=self.pkg_dir).sdist
+                print("cc", self.core["config_settings"])
```
gaborbernat, line 217: I don't think this is part of the core settings, and I don't think it should be. This is the packaging environment setting in my view.

nschloe: I'm not really sure what this means. What part of the code is the _packaging env setting_?

nschloe: Got it.

gaborbernat:
You likely want to have a base settings and then allow overriding for wheel/sdist/editable.

nschloe:
Do you mean something like

```ini
[testenv:.pkg]
pep517_config_settings_base = 
  a = b
  x = yz
pep517_config_settings_sdist = 
  a = c
  x = z
pep517_config_settings_wheel = 
  a = W
  x = W
```
? In the interest of keeping things simple, I'm not convinced that's necessary. After all, there's a separate setting for sdist/wheel/editable builds, and when the user wants to change that, she has to touch the config _anyway_. E.g., from

```ini
[testenv:.pkg]
package = editable
pep517_config_settings = 
  a = b
  x = yz
```

to

```ini
[testenv:.pkg]
;package = editable
;pep517_config_settings = 
;  a = b
;  x = yz
package = sdist
pep517_config_settings = 
  a = S
  x = S
```
What do you think?

gaborbernat:
Multiple different tox environments can reuse the same packaging environment, so it's important they are configurred differently:

```ini
[testenv]
wheel_build_env = .pkg

[testenv:wheel]
package = wheel


[testenv:sdist]
package = sdist

[testenv:editable]
package = editable
```

All these 3 envs will use the same `.pkg` and its settings even though they build differently.

nschloe:
Ah, so what should the config look like exactly?

gaborbernat:
```
[testenv]
wheel_build_env = .pkg

[testenv:wheel]
package = wheel


[testenv:sdist]
package = sdist

[testenv:editable]
package = editable

[testenv:.pkg]
config_settings = 
  a = S
wheel_config_settings = 
  a = S
sdist_config_settings = 
  a = S
editable_config_settings = 
  a = S
```

If `wheel_config_settings` is not set defaults to `config_settings` and so on.

nschloe:
If
```
config_settings = 
  a = S
  b = T
config_settings_wheel = 
  a = U
```
should the complete list of config settings by
```
  a = U
  b = T
```
or
```
  a = U
```
?

gaborbernat:
Don't merge. Instead override.

src/tox/tox_env/python/virtual_env/package/pyproject.py
```diff
@@ -424,3 +467,10 @@ def _wheel_directory(self) -> Iterator[Path]:
 @impl
 def tox_register_tox_env(register: ToxEnvRegister) -> None:
     register.add_package_env(Pep517VirtualEnvPackager)
+
+
+def merge_dicts(*dicts):
```
gaborbernat, line 472: Don't merge dicts.

gaborbernat (changes requested):
Add tests too please.

nschloe:
> What about the `self._wheel_config_settings`?

`self._wheel_config_settings` is now overridden my either of the other `config_settings` (if present).

nschloe:
> Add tests too please.

Where would they go? `tests/conftest.py`?

nschloe:
I'm having substantial problems setting up a test. There seems to be no similar test prototype I could take inspiration from. For example `meta_dir` and `pkg_dir` don't appear in the tests at all. That's not mentioning that I have no idea how to test if the arguments are actually passed through.

nschloe:
@gaborbernat I'm gonna need some guidance here.

gaborbernat:
Will need https://github.com/tox-dev/pyproject-api/pull/99 first released.

docs/changelog/3090.feature.rst
```diff
@@ -0,0 +1 @@
+Add support for setting build backend ``config_settings`` in the configuration file - by :user:`gaborbernat`.
```
webknjaz, line 1: @nschloe did you forget to use your own username here?

nschloe: Not really. I made _some_ contributions here, but Gabor finished it all up.

docs/config.rst
webknjaz, line 1: @nschloe I don't see any mentions of just `config_settings` that is not hook-specific but shared. Did that end up not being implemented?

gaborbernat: We did not add that, but you can do that with config reference in your tox file.

webknjaz: @gaborbernat do you mean compute dynamically in a `toxfile.py`?

gaborbernat: No, I meant just set it once in `tox.ini` and use config reference to set it for every hook you care about.

webknjaz: @gaborbernat when you say “config reference”, what exactly do you mean? I'm guessing that's what I don't understand. Could you help me bridge the gap in understanding?

gaborbernat: https://tox.wiki/en/4.21.2/config.html#configuration-reference

gaborbernat: refering to another config in the config file

webknjaz: Ah, I didn't know about that new syntax. The old one was substitutions. Thanks!

PR #3094, opened 2023-08-17
Document release process

jugmac00:


jugmac00:
Looks like we have an issue with our `furo` theme

```
Extension error (furo):
Handler <function _html_page_context at 0x7fbb82980ae0> for event 'html-page-context' threw an exception (exception: This documentation is not using `furo.css` as the stylesheet. If you have set `html_style` in your conf.py file, remove it.)
```

FWIW, we do not use `html_style` in our `conf.py`.

I have no time to debug the issue right now.

gaborbernat (approved):

Issue #3096, opened 2023-08-17
Getting `environments not found` error when running a partial set of factors


atugushev:
## Issue

Since `4.9.0` having `env_list=a-b` in `tox.ini` and running `top -e a` raises a error:
```
HandledError| provided environments not found in configuration file: ['a']
```

This error was caught today on on pip-tools' CI (https://github.com/jazzband/pip-tools/pull/1963), see the [log](https://github.com/jazzband/pip-tools/actions/runs/5895206825/job/15990457732?pr=1963#step:9:23). It's likely a regression introduced in https://github.com/tox-dev/tox/pull/3089.

## Environment

Provide at least:

- OS: Ubuntu

<details open>
<summary>Output of <code>pip list</code> of the host Python, where <code>tox</code> is installed</summary>

```console
Package       Version
------------- -------
cachetools    5.3.1
chardet       5.2.0
colorama      0.4.6
distlib       0.3.7
filelock      3.12.2
packaging     23.1
pip           22.3.1
platformdirs  3.10.0
pluggy        1.2.0
pyproject-api 1.5.4
setuptools    65.5.1
tox           4.9.0
virtualenv    20.24.3
wheel         0.40.0
```

</details>

## Output of running tox

<details open>
<summary>Output of <code>tox -rvv</code></summary>

```console
a-b: 266 W remove tox env folder /.tox/a-b [tox/tox_env/api.py:322]
a-b: 307 I find interpreter for spec PythonSpec(path=/usr/local/bin/python) [virtualenv/discovery/builtin.py:58]
a-b: 307 I proposed PythonInfo(spec=CPython3.11.3.final.0-64, exe=/usr/local/bin/python, platform=linux, version='3.11.3 (main, May  3 2023, 08:13:29) [GCC 10.2.1 20210110]', encoding_fs_io=utf-8-utf-8) [virtualenv/discovery/builtin.py:65]
a-b: 308 D accepted PythonInfo(spec=CPython3.11.3.final.0-64, exe=/usr/local/bin/python, platform=linux, version='3.11.3 (main, May  3 2023, 08:13:29) [GCC 10.2.1 20210110]', encoding_fs_io=utf-8-utf-8) [virtualenv/discovery/builtin.py:67]
a-b: 310 D filesystem is case-sensitive [virtualenv/info.py:26]
[...]

PR #3099, opened 2023-08-19
Accept environments with defined factors or of python selector form - suggest closest

BeyondEvil:
Fixes: #3096 

The `4.9.0` introduced a regression to valid environments:

Hyphenated environments:

```ini
[tox]
env_list=a-b
...
```

Combined environments:

```ini
[tox]
env_list=a
commands = 
    with: python -c 'print("with")'
   !with: python -c 'print("without")'
```

### DIFF
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -4,7 +4,8 @@ import logging
 import re
 from collections import Counter
 from dataclasses import dataclass
-from itertools import chain, permutations
+from difflib import get_close_matches
+from itertools import chain
 from typing import TYPE_CHECKING, Dict, Iterable, Iterator, List, cast
 
 from tox.config.loader.str_convert import StrConvert
@@ -117,6 +118,10 @@ class _ToxEnvInfo:
     package_skip: tuple[str, Skip] | None = None  #: if set the creation of the packaging environment failed
 
 
+_DYNAMIC_ENV_FACTORS = re.compile(r"(pypy|py|cython|)((\d(\.\d+(\.\d+)?)?)|\d+)?")
+_PY_PRE_RELEASE_FACTOR = re.compile(r"alpha|beta|rc\.\d+")
+
+
 class EnvSelector:
     def __init__(self, state: State) -> None:
         # needs core to load the default tox environment list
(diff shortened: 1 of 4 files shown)


src/tox/session/env_select.py
```diff
@@ -389,6 +391,16 @@ def _mark_provision(self, on: bool, provision_tox_env: str) -> None:  # noqa: FB
...
 
 
+def _find_env_match(value: str, state_conf: set) -> bool:
+    return any(value in conf.split("-") for conf in state_conf) or any(
+        value == "-".join(combo) for combo in permutations(state_conf, 2)
+    )
+
+
+def _is_valid_exception(env: str) -> bool:
+    return env.startswith("py") or env in (".pkg",)
```
gaborbernat, line 401: Can you ensure that 3.9/3.11 is also handled as an exception, we can use a Regen for that 👍

BeyondEvil: Mind clarifying? What use-case is currently not working?

gaborbernat: `tox r -e 3.11` and `tox r -e 3.11-magic` should still work 🤔

src/tox/session/env_select.py
```diff
@@ -389,6 +391,16 @@ def _mark_provision(self, on: bool, provision_tox_env: str) -> None:  # noqa: FB
         self._provision = on, provision_tox_env
 
 
+def _find_env_match(value: str, state_conf: set[str]) -> bool:
+    return any(value in conf.split("-") for conf in state_conf) or any(
```
BeyondEvil, line 395: I'd like to add some comments to explain what is happening here, but I'm not familiar with the correct nomenclature.

gaborbernat: tox environment names are made up of factors (so `py39-magic-test` is a tox environment made up of three factors `py39`, `magic`, `test`).

src/tox/session/env_select.py
```diff
@@ -389,6 +391,16 @@ def _mark_provision(self, on: bool, provision_tox_env: str) -> None:  # noqa: FB
...
+        value == "-".join(combo) for combo in permutations(state_conf, 2)
```
gaborbernat, line 396: Would this work for permutations of 3? Should we just check that any factor matches the regex:

```
(py)|(\d+([.]\d+([.]\d+)?)?)|(py\d+([.]\d+([.]\d+)?)?)
```

I feel would generalize better 🤔

src/tox/session/env_select.py
BeyondEvil: Something like this: https://regex101.com/r/u2Kwuy/1 ? @gaborbernat

gaborbernat: https://www.debuggex.com/r/8ZDSfICtm-vkvTjs was my approach

BeyondEvil: Seems more complex and don't cover the "-magic" case?

Is my approach missing a case?

src/tox/session/env_select.py
gaborbernat: There's some more of these here https://github.com/tox-dev/tox/blob/main/src/tox/config/loader/ini/factor.py and explanation how those are split/generated 👍

src/tox/session/env_select.py
gaborbernat: Well the -magic doesn't matter. If any of the factors match this regex then we should accept the whole env name.

gaborbernat: The idea here is to still allow dynamic environments for envs having my regex, plus any other factors (and their combinations) the user defined.

src/tox/session/env_select.py
BeyondEvil: What's the maximum number of factors? Given that you don't create the "manually" by specifying an env in `env_list` that has a hyphen in it?

BeyondEvil: I'm sorry. I'm really struggling to understand the heuristics at play here.

gaborbernat: There is no limit.

src/tox/session/env_select.py
BeyondEvil: Ok, but none of our regexs would match `test-cov` or `cov-test`, right? 🤔 

It's almost 4AM where I'm at. I think I need to look at this with fresh eyes.

However, I think generally explicit > implicit. So maybe this is just introducing complexity that is unnecessary. 🤷‍♂️

BeyondEvil: Can you show an example tox.ini that creates a environment with three or more factors _without_ having `env_list` contain any factors.

e.g.:

```ini
[tox]
env_list=3.9
commands =
  cov: python -c 'print("with cov")'
 !cov: python -c 'print("without cov")'
```

Example above will create a maximum of two possible factors, right?

`3.9` and `3.9-cov`

How can that be extended to three or more factors?

gaborbernat (approved):

PR #3109, opened 2023-08-31
Set the --parallel default to "auto", not CPU count

paravoid:
There is no functional reason to have the ArgumentParser default be `auto_detect_cpus()`, as the option goes through a parser, `parse_num_processes()`, that replaces "auto" with `auto_detect_cpus()`, i.e. the CPU count.

Making the default "auto" makes the documentation clearer and reproducible. Currently the [tox docs](https://tox.wiki/en/latest/cli_interface.html#tox-run-parallel-(p)) say:
>   -p VAL, --parallel VAL - run tox environments in parallel, the
>   argument controls limit: all, auto - cpu count, some positive number,
>   zero is turn off (default: 2)

That is misleading to end users: the default is not `2`, but `auto`; 2 just happens to be the CPU count of the machine that generated these docs.

Note that there was a DEFAULT_PARALLEL constant, that was however unused, as a result of an earlier refactoring. Reuse that constant for "auto" here.

- [ ] added news fragment in `docs/changelog` folder (too minor)
- [ ] updated/extended the documentation (too minor)

### DIFF
--- a/src/tox/session/cmd/run/parallel.py
+++ b/src/tox/session/cmd/run/parallel.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)
 
 ENV_VAR_KEY = "TOX_PARALLEL_ENV"
 OFF_VALUE = 0
-DEFAULT_PARALLEL = OFF_VALUE
+DEFAULT_PARALLEL = "auto"
 
 
 @impl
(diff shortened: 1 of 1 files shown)


gaborbernat (approved):

paravoid:
Thanks for the quick review & merge!

PR #3111, opened 2023-09-01
Fix , being used as value parser for env var configs

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/3112

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1025,12 +1025,51 @@ For example, given this config:
 
 You could enable ``ignore_errors`` by running::
 
-    tox --override testenv.ignore_errors=True
+.. code-block:: bash
+
+   tox --override testenv.ignore_errors=True
 
 You could add additional dependencies by running::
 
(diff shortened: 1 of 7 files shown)


docs/config.rst
```diff
@@ -1025,12 +1025,51 @@ For example, given this config:
 
 You could enable ``ignore_errors`` by running::
```
posita, line 1026: Probably shouldn't keep the `::` here.

gaborbernat: No, the code block should handle that. `::` is alternative to `code-block` without specifying the syntax.

posita: Right, but you can't have both. You can have something like this:

``` rst
This is some code (note the double colon)::

    Here is my code!
```

_**Or**_ you can have something like this:

``` rst

This is some code (note the single colon):

.. code-block:: lang

    Here is my code!
```

But this won't work (notice the double colon after the text _**and**_ the use of `code-block`:

``` rst
This is some code (note the double colon)::

.. code-block:: lang

    Here is my code!
```

PR #3120, opened 2023-09-07
Remove stray colons in `config.rst` left over from #3111

### LINKED ISSUE #3118: Naked `.. code-block:: bash` in some places throughout docs
## Issue

From [**Overriding configuration from the command line**](https://tox.wiki/en/latest/config.html#overriding-configuration-from-the-command-line):

Screenshot from 2023-09-07 13-47-46

Apparent source: #3111

This page appears to be generated from [`config.rst`](https://github.com/tox-dev/tox/blob/main/docs/config.rst#overriding-configuration-from-the-command-line). I'm not quite sure what the problem is here, since *some* code blocks work, but others (perhaps those with only a single line) don't. It's interesting to note that GH's renderer suffers from the same problem.

So this _**may**_ be a bug in Sphinx, but if a work-around is available (or maybe selecting an alternative Sphinx version fixes this), it is probably worth exploring for the sake of a polished experience for doc consumers.

posita:
Fixes #3118. (See screenshot below.)

Also addresses small (ab)use of `.. code-block:: console` in `installation.rst`, which is meant for use with interactive sessions.

[screenshot](https://github.com/tox-dev/tox/assets/222581/b4f24c23-a3d8-4ffb-aeca-00d4414f9dac)

Also viewable in the [revised `config.rst`](https://github.com/posita/tox/blob/posita/0/revert-breaking-doc-changes/docs/config.rst#overriding-configuration-from-the-command-line) (compared to the [original `config.rst`](https://github.com/tox-dev/tox/blob/main/docs/config.rst#overriding-configuration-from-the-command-line)).

### DIFF
--- a/docs/changelog/3118.bugfix.rst
+++ b/docs/changelog/3118.bugfix.rst
@@ -1,1 +1,1 @@
-Fix bug in ``config.rst`` by removing stray colons left over from (:issue:`3111`) - by :user:`posita`. (:issue:`3118`)
+Fix bug in ``config.rst`` by removing stray colons left over from (:issue:`3111`) - by :user:`posita`.
--- a/docs/changelog/3118.bugfix.rst
+++ b/docs/changelog/3118.bugfix.rst
@@ -1,0 +1,2 @@
+- Fix bug in ``config.rst`` by removing stray colons left over from (:issue:`3111`) - by
+  :user:`posita`. (:issue:`3118`)
--- a/docs/changelog/3118.bugfix.rst
+++ b/docs/changelog/3118.bugfix.rst
@@ -1,2 +1,1 @@
-- Fix bug in ``config.rst`` by removing stray colons left over from (:issue:`3111`) - by
-  :user:`posita`. (:issue:`3118`)
+Fix bug in ``config.rst`` by removing stray colons left over from (:issue:`3111`) - by :user:`posita`. (:issue:`3118`)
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):

docs/changelog/3118.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+- Fix bug in ``config.rst`` by removing stray colons left over from (:issue:`3111`) - by
```
gaborbernat, line 1: Remove leading `-`.

gaborbernat (changes requested):

docs/changelog/3118.bugfix.rst
posita: aye aye!

docs/changelog/3118.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix bug in ``config.rst`` by removing stray colons left over from (:issue:`3111`) - by :user:`posita`. (:issue:`3118`)
```
gaborbernat, line 1: Remove `(:issue:`3118`)` that's generated.

gaborbernat (changes requested):

docs/changelog/3118.bugfix.rst
gaborbernat: https://tox--3120.org.readthedocs.build/en/3120/changelog.html#vnext-2023-09-07 see you're duplicating data

posita: Ah! Whoops! I was imitating without understanding. :blush:

fixed

gaborbernat (approved):

Issue #2999, opened 2023-05-01
tox breaks ipython input/output when


albertodonato:
## Issue

When running ipython as a command in a tox env, the input and output are messed up. Terminal gets reset at every enter from the ipython shell. This seems to be due some conflict interacting with the console
A workaround is to pass `--simple-prompt` to ipython, but this means ipython doesn't provide things like autocompletion.

Tested with `tox 4.5.1` and `ipython 8.13.1`.

## Environment

Provide at least:

- OS: Ubuntu 23.04 amd64
- `pip list` of the host Python where `tox` is installed:

```console
$ pip list
Package       Version
------------- -------
cachetools    5.3.0
chardet       5.1.0
colorama      0.4.6
distlib       0.3.6
filelock      3.12.0
packaging     23.1
pip           23.1.2
platformdirs  3.5.0
pluggy        1.0.0
pyproject_api 1.5.1
setuptools    66.1.1
tox           4.5.1
virtualenv    20.23.0
```

## Output of running tox

Provide the output of `tox -rvv`:

```console
hpy: 70 W remove tox env folder /home/ack/Desktop/.tox/py [tox/tox_env/api.py:322]
py: 131 I find interpreter for spec PythonSpec(path=/home/ack/Desktop/test/bin/python3) [virtualenv/discovery/builtin.py:58]
py: 132 D got python info of /usr/bin/python3.11 from /home/ack/.local/share/virtualenv/py_info/1/ca3ed784184f1b3bb7c3539bfb45e71710cd27667424f92c2d5bb4df9c107c23.json [virtualenv/app_data/via_disk_folder.py:131]
py: 132 I proposed PythonInfo(spec=CPython3.11.2.final.0-64, system=/usr/bin/python3.11, exe=/home/ack/Desktop/test/bin/python3, platform=linux, version='3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0]', encoding_fs_io=utf-8-utf-8) [virtualenv/discovery/builtin.py:65]
py: 132 D accepted PythonInfo(spec=CPython3.11.2.final.0-64, system=/usr/bin/python3.11, exe=/home/ack/Desktop/test/bin/python3, platform=linux, version='3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0]', encoding_fs_io=utf-8-utf-8) [virtualenv/discovery/builtin.py:67]
py: 133 D filesystem is case-sensitive [virtualenv/info.py:26]
[...]

PR #3139, opened 2023-10-13
Fix terminal size in tox commands (#2999)

ziima:
Found it! And because it took me 6 hours to debug it, I'm going to describe how the debugging went.

At first, I started debugging the tox itself. Quickly I managed to discover, the issue can be triggered by `python3 -c "from prompt_toolkit import prompt; prompt('Give me some input: ')"` which behaved differently in tox and outside. After couple of hours trying to navigate the tox internals I landed at `ReadViaThreadUnix` where I noticed it print a significant number of '\r\n' pairs after the prompt. That led me to the `prompt_toolkit`. I spend some time there, until I discovered the screen it renders within tox has an unusual height. A but later I noticed the screen it tries to render have switched values of columns and rows when compared to `COLUMNS` and `ROWS` pass in env values by tox. Bit more digging lead to `os.get_terminal_size()`. The MWE was now just simply:

```ini
[testenv]
deps = ipython
commands =
    python3 -c "import os; print(os.get_terminal_size())"
```

```console
# tox
py: commands[0]> python3 -c 'import os; print(os.get_terminal_size())'
os.terminal_size(columns=54, lines=189)

# .tox/py/bin/python3 -c 'import os; print(os.get_terminal_size())'
os.terminal_size(columns=189, lines=54)
```

I had no idea what to do now, but I didn't want to give up. I realized I could just grep `tox` source code for columns and I finally found match in `_pty` function. I tried to switch the two arguments and voila, works like a charm.

This PR is not yet finished, I haven't found how to design a test for this and I'm too tired at the moment. I hope to add it soon.

- [ ] ~~updated/extended the documentation~~

### DIFF
--- a/docs/changelog/2999.bugfix.rst
+++ b/docs/changelog/2999.bugfix.rst
@@ -1,0 +1,1 @@
+Fix terminal size of tox subcommands (fixes ipython, ipdb, prompt_toolkit, ...).
--- a/src/tox/execute/local_sub_process/__init__.py
+++ b/src/tox/execute/local_sub_process/__init__.py
@@ -314,7 +314,7 @@ def _pty(key: str) -> tuple[int, int] | None:
     # adjust sub-process terminal size
     columns, lines = shutil.get_terminal_size(fallback=(-1, -1))
     if columns != -1 and lines != -1:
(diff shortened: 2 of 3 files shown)


ziima:
PR is ready.

PR #3145, opened 2023-10-30
Create a `.gitignore` in `work_dir` and info file directories

### LINKED ISSUE #2530: Automatic `.gitignore`
Some testing tools add a `.gitignore` with a `*` rule into the temporary directories they create. This saves projects needing to add a pattern to their Git ignore file, and prevents accidents where such a pattern is missing. For example, [Coverage.py does so](https://github.com/nedbat/coveragepy/blob/a4da2c62c45132e1a7d1e1f78e32522be5ad405f/coverage/html.py#L269-L274), since version 6.1.

Perhaps tox could do the same when creating a `.tox` directory?

kurtmckee:
> ## NOTE:
>
> Tests, and possibly documentation, are still needed for this work. Please confirm the implementation is acceptable (for example, perhaps a helper function is desirable).

This PR introduces blanket `.gitignore` files that are created in the following locations:

* In the `work_dir` (such as `.tox/`)
* Wherever an info JSON file is written

This addresses several issues I found when running tox under the following circumstances:

```bash
# Create a completely empty directory, then initialize git.
mkdir test-tox
cd test-tox
git init

# Run `tox` with an environment for an interpreter that IS NOT installed.
# RESULT: `.tox/py40/.tox-info.json` visible to git
tox -e py40

# Run `tox devenv` with an environment for an interpreter that IS NOT installed.
# RESULT: `venv/.tox-info.json` visible to git
tox devenv -e py40

# Run `tox devenv` with an environment for an interpreter that IS installed.
# RESULT: `.tox/py/log/1-install_package.log` is visible to git
tox devenv -e py
```

The changes introduced in this PR address the visibility of these files in git.

Closes #2530

Still to do:

### DIFF
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -413,6 +413,9 @@ class EnvSelector:
             yield name
 
     def ensure_only_run_env_is_active(self) -> None:
+        if not self._state.conf.core["work_dir"].exists():
+            self._state.conf.core["work_dir"].mkdir()
+            (self._state.conf.core["work_dir"] / ".gitignore").write_text("*", encoding="utf-8")
         envs, active = self._defined_envs, self._env_name_to_active()
         invalid = [n for n, a in active.items() if a and isinstance(envs[n].env, PackageToxEnv)]
         if invalid:
--- a/src/tox/tox_env/info.py
+++ b/src/tox/tox_env/info.py
@@ -63,6 +63,8 @@ class Info:
 
     def _write(self) -> None:
         self._path.parent.mkdir(parents=True, exist_ok=True)
+        if not (self._path.parent / ".gitignore").exists():
+            (self._path.parent / ".gitignore").write_text("*", encoding="utf-8")
         self._path.write_text(json.dumps(self._content, indent=2))
 
 


src/tox/session/env_select.py
```diff
@@ -413,6 +413,9 @@ def iter(  # noqa: A003
             yield name
 
     def ensure_only_run_env_is_active(self) -> None:
+        if not self._state.conf.core["work_dir"].exists():
+            self._state.conf.core["work_dir"].mkdir()
+            (self._state.conf.core["work_dir"] / ".gitignore").write_text("*", encoding="utf-8")
```
gaborbernat, line 418: This doesn't seem right.

src/tox/tox_env/info.py
```diff
@@ -63,6 +63,8 @@ def reset(self) -> None:
 
     def _write(self) -> None:
         self._path.parent.mkdir(parents=True, exist_ok=True)
+        if not (self._path.parent / ".gitignore").exists():
+            (self._path.parent / ".gitignore").write_text("*", encoding="utf-8")
```
gaborbernat, line 67: This looks ok-ish 🤔

gaborbernat (changes requested):
Test, changelog, etc.

src/tox/session/env_select.py
kurtmckee: Ah, I see.

I put the code here because the very next line contains a call to `self._env_name_to_active()`, which ends up creating the `work_dir` if it doesn't already exist. I considered putting the code in `execute()`, but that function wasn't consistently called in the tox subcommands I tested. `ensure_only_run_env_is_active()` was called consistently for the tox subcommands I tested.

The code is duplicated because `info.py` creates a directory and a file that isn't necessarily in the `work_dir` (see the `.tox/` vs `venv/` files listed in the text I posted above). Duplicating the code ensured that a `.gitignore` file existed when the info file wasn't put in a subdirectory of `work_dir`.

gaborbernat:
Stalled.

kurtmckee:
Oh! I thought I was waiting on additional feedback.

PR #3156, opened 2023-11-15
docs(config): Fix base_python default reference

rpatterson:
The previous description of `{package_root}` doesn't make much sense because it doesn't have anything to do with Python versions.  I'm guessing it was a copy and paste that wasn't updated.

Change it to a value that tries to be both short and accurate matching the conventions observed in other options.

- ~~[ ] ensured there are test(s) validating the fix~~

### DIFF
--- a/docs/changelog/3156.doc.rst
+++ b/docs/changelog/3156.doc.rst
@@ -1,0 +1,1 @@
+Fix default value for ``base_python`` - by :user:`rpatterson`.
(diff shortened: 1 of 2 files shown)


rpatterson:
I don't have much time to shepherd this through so I understand if you need to close as `wontfix` but I thought it was better to submit than not.

gaborbernat (approved):

PR #3159, opened 2023-11-16
Imply `--parallel` when `--parallel-no-spinner` is passed 

### LINKED ISSUE #3158: `--parallel-no-spinner` should imply `--parallel`
## Issue

The flag `--parallel-no-spinner` reads like "the tests should run parallel, but don't show spinner". But if you just pass `--parallel-no-spinner` without also passing `--parallel`, it does nothing. I think the presence of this flag implies that we want to run the tests parallelly.

tusharsadhwani:
Resolves #3158


gaborbernat (changes requested):
Please add test validating the change.

tests/session/cmd/test_parallel.py
```diff
@@ -169,3 +171,27 @@ def test_parallel_requires_arg(tox_project: ToxProjectCreator) -> None:
...
+
+    mocked.assert_called_once_with(
+        mock.ANY,
+        max_workers=None,
+        has_spinner=False,
+        live=False,
+    )
+
+    # Test legacy parser
+    with mock.patch.object(parallel, "execute") as mocked:
```
gaborbernat, line 189: Looks like this should be a separate test. One test should test one thing only most of the time.

gaborbernat (changes requested):

tests/session/cmd/test_parallel.py
tusharsadhwani: Alright, made the two into their own tests.

src/tox/session/cmd/legacy.py
```diff
@@ -110,7 +110,7 @@ def legacy(state: State) -> int:
             option.env = CliEnv(["py"])
         option.devenv_path = Path(option.devenv_path)
         return devenv(state)
-    if option.parallel != 0:  # only 0 means sequential
+    if option.parallel_no_spinner is True or option.parallel != 0:  # only 0 means sequential
```
tusharsadhwani, line 113:
```suggestion
    if option.parallel != 0:  # only 0 means sequential
```

gaborbernat (approved):

webknjaz:
@gaborbernat looks like this caused a regression when the `TOX_PARALLEL_NO_SPINNER=1` env var is set globally, in envs like CI.

gaborbernat:
@webknjaz can you detail the regression? @@tusharsadhwani  can you take a look?

tusharsadhwani:
I can check. A brief explanation of the issue will help.

Issue #3151, opened 2023-11-08
On Windows APPDATA is missing in os.environ


Stefanhg:
## Issue

os.environ is missing APPDATA which are used in some methods. 
Issue was initially reproduced in setuptools_scm
https://github.com/pypa/setuptools_scm/issues/861

I have attached a .zip file that includes the example used to reproduce the issue.
If i just run pytest alone the test passes.

**It works until version 3.28.0** 

## Environment

Provide at least:


- OS: Windows 1010.0.19045 build 19045

<details open>
<summary>Output of <code>pip list</code> of the host Python, where <code>tox</code> is installed</summary>

```
cachetools    5.3.2
chardet       5.2.0
colorama      0.4.6
distlib       0.3.7
filelock      3.13.1
packaging     23.2
pip           22.3.1
platformdirs  3.11.0
pluggy        1.3.0
pyproject-api 1.6.1
setuptools    65.5.0
tomli         2.0.1
tox           4.11.3
virtualenv    20.24.6
```

</details>

## Output of running tox

<details open>
<summary>Output of <code>tox -rvv</code></summary>

```
base: 238 W remove tox env folder D:\temp\test_tox\.tox\base [tox\tox_env\api.py:322]
base: 300 I find interpreter for spec PythonSpec(path=C:\Users\XXXXXXXX\OneDrive - XXXXXXXX A S\D_DRIVE\Python\test_tox\venv\Scripts\python.exe) [virtualenv\discovery\builtin.py:58]
base: 300 D got python info of %s from (WindowsPath('C:/Python310/python.exe'), WindowsPath('C:/Users/XXXXXXXX/AppData/Local/pypa/virtualenv/py_info/1/2be5faf79681da6f2a61fdfdd5405d65d042280f7fba6178067603e3a2925119.json')) [virtualenv\app_data\via_disk_folder.py:131]
base: 316 I proposed PythonInfo(spec=CPython3.10.10.final.0-64, system=C:\Python310\python.exe, exe=C:\Users\XXXXXXXX\OneDrive - XXXXXXXX A S\D_DRIVE\Python\test_tox\venv\Scripts\python.exe, platform=win32, version='3.10.10 (tags/v3.10.10:aad5f6a, Feb  7 2023, 17:20:36) [MSC v.1929 64 bit (AMD64)]', encoding_fs_io=utf-8-cp1252) [virtualenv\discovery\builtin.py:65]
[...]

PR #3160, opened 2023-11-18
Fixed 'AppData' not passed to env by default (#3151)

Stefanhg:
Fixes AppData missing in environment variables by default.
It was found in setuptools_scm that on some machines it was not able to find the version in HG as AppData is required. I assume it is related to where Pythons packages are placed per default.

I have not updated documentation as I did not see any documentation about what environment variables are passed per default. 

Issue where problem was found:
https://github.com/pypa/setuptools_scm/issues/861

### DIFF
--- a/tests/session/cmd/test_show_config.py
+++ b/tests/session/cmd/test_show_config.py
@@ -264,7 +264,7 @@ def test_show_config_matching_env_section(tox_project: ToxProjectCreator) -> Non
 
 
 def test_package_env_inherits_from_pkgenv(tox_project: ToxProjectCreator, demo_pkg_inline: Path) -> None:
-    project = tox_project({"tox.ini": "[pkgenv]\npass_env = A, B\ndeps=C\n D"})
+    project = tox_project({"tox.ini": "[pkgenv]\npass_env = A, AA\ndeps=C\n D"})
     outcome = project.run("c", "--root", str(demo_pkg_inline), "-k", "deps", "pass_env", "-e", "py,.pkg")
     outcome.assert_success()
     exp = """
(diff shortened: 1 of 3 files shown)


Stefanhg:
@gaborbernat The test_result_json_sequential failing testcase seems to be failing randomly. 
I managed to make it fail once and then 200 retests later none failed, waited 1 hour and ran it 3 more times and then caught the error again.
Have you faced this before?

tests/session/cmd/test_show_config.py
```diff
@@ -262,7 +264,7 @@ def test_show_config_matching_env_section(tox_project: ToxProjectCreator) -> Non
 
 
 def test_package_env_inherits_from_pkgenv(tox_project: ToxProjectCreator, demo_pkg_inline: Path) -> None:
-    project = tox_project({"tox.ini": "[pkgenv]\npass_env = A, B\ndeps=C\n D"})
+    project = tox_project({"tox.ini": "[pkgenv]\npass_env = A, AA\ndeps=C\n D"})
```
jugmac00, line 267: Why do you change this test?

Stefanhg: I did include a comment about it in my commit:

Because AppData is now included on Windows, it will include A, AppData, B when testing on Windows
therefore by using A, AA instead it will have the right flow/order.

Stefanhg: It may be that my comment is not sufficient enough. I will try to explain it a bit better.
Because AppData is included in the passed environment variables per default on Windows since the pass_env is [A, B, ...] in the test, AppData will be included between A and B causing the test to fail: [A, AppData, B, ...]
So therefore changing it to [A, AA] solved it. It was not the most beautiful solution so It could be considered if I should try to find a different value like [K, L] if that is preferred? 
I was considering using [1, A] but that also looked messy. 

Does that make sense?

gaborbernat (approved):

Issue #163, opened 2016-09-17
Document PY_COLORS to enforce color output


pytoxbot:
- Bitbucket: https://bitbucket.org/hpk42/tox/issue/163
- Originally reported by: @hashar
- Originally created at: 2014-03-21T18:30:40.458

While integrating a tox based job in Jenkins (which has a dumb terminal) I was looking for a way to force color output.
Looking at _py/terminalwriter.py , it looks up the environment variable PY_COLORS. A value of 1 would enforce color, a value of 0 would disable color. The global override the TERM value.

It would be nice to have PY_COLORS documented in the tox documentation.

py-trunk doc update request is: https://bitbucket.org/hpk42/py-trunk/issue/138/py_colors-env-is-undocumented

hashar:
Thank you @gaborbernat  I had completely forgot/missed this issue :]

PR #3172, opened 2023-12-14
Always pass FORCE_COLOR & NO_COLOR to the environment

hashar:
The environment variables `FORCE_COLOR` and `NO_COLOR`are a popular way to force or disable color output. An example usage is pytest being run under a CI system.

### DIFF
--- a/src/tox/tox_env/api.py
+++ b/src/tox/tox_env/api.py
@@ -219,6 +219,8 @@ class ToxEnv(ABC):
             "LD_LIBRARY_PATH",  # location of libs
             "LDFLAGS",  # linker flags
             "HOME",  # needed for `os.path.expanduser()` on non-Windows systems
+            "FORCE_COLOR",  # force color output
+            "NO_COLOR",  # disable color output
         ]
         if sys.stdout.isatty():  # if we're on a interactive shell pass on the TERM
             env.append("TERM")
--- a/docs/changelog/3171.feature.rst
+++ b/docs/changelog/3171.feature.rst
@@ -1,0 +1,1 @@
+Always pass ``FORCE_COLOR`` and ``NO_COLOR`` to the environment
--- a/tests/session/cmd/test_show_config.py
+++ b/tests/session/cmd/test_show_config.py
@@ -123,8 +123,11 @@ def test_pass_env_config_default(tox_project: ToxProjectCreator, stdout_is_atty:
         + (["APPDATA"] if is_win else [])
         + ["CC", "CCSHARED", "CFLAGS"]
         + (["COMSPEC"] if is_win else [])
-        + ["CPPFLAGS", "CURL_CA_BUNDLE", "CXX", "HOME", "LANG", "LANGUAGE", "LDFLAGS", "LD_LIBRARY_PATH"]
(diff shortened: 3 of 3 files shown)


gaborbernat (changes requested):
https://no-color.org/ and FORCE_COLOR are the standard env vars. These app-specific ones don't get to be in the core.

hashar:
I went with `PY_COLORS` a while ago which was even used by tox at some point ( #163 ).  The other use case was for `pytest` but it has learned about `NO_COLOR` `FORCE_COLOR` since version 6.0.0 ( https://github.com/pytest-dev/pytest/pull/7466 ).
```python
def should_do_markup(file: TextIO) -> bool:
    if os.environ.get("PY_COLORS") == "1":
        return True
    if os.environ.get("PY_COLORS") == "0":
        return False
    if "NO_COLOR" in os.environ:
        return False
    if "FORCE_COLOR" in os.environ:
        return True
    return (
        hasattr(file, "isatty") and file.isatty() and os.environ.get("TERM") != "dumb"
    )
```

May I repurpose this pull request to add `NO_COLOR` and `FORCE_COLOR` to the environment variables passed by default? I find them handy.

gaborbernat:
I don't think you need to add support for them. They should work today. Did you find the opposite?

hashar:
`NO_COLOR` and `FORCE_COLOR` are only handled by the  config cli parser to set the `color` flag. For the environments, they are not passed:
```ini
[testenv]
commands = python3 -c 'import os; print(os.environ.get("FORCE_COLOR"))'
```
```
$ FORCE_COLOR=1 tox -q
None
```

(same for `NO_COLOR`)

gaborbernat:
If that doesn't work, a pull request is welcome. Just make sure to add the test and the changelog.

hashar:
May you please reopen the pull request or should I send a new one? :)

src/tox/tox_env/api.py
```diff
@@ -219,6 +219,7 @@ def _default_pass_env(self) -> list[str]:
             "LD_LIBRARY_PATH",  # location of libs
             "LDFLAGS",  # linker flags
             "HOME",  # needed for `os.path.expanduser()` on non-Windows systems
+            "PY_COLORS",  # Popular way to force color output in CI
```
gaborbernat, line 222: Use FORCE_COLOR, NO_COLOR instead.

gaborbernat (changes requested):
Don't forget to add test and changelog.

src/tox/tox_env/api.py
hashar: Done! Replaced by `FORCE_COLOR` and `NO_COLOR`.

hashar:
@gaborbernat I have repurposed the pull request to pass pass `FORCE_COLOR` and `NO_COLOR` instead of `PY_COLORS`.

gaborbernat:
Yeah, but the CI is failing now that's blocking progress.

hashar:
A test was failing under Windows OS. 

I also fixed `docs/changelog/3171.feature.rst` which I forgot to update.

gaborbernat (approved):

PR #3173, opened 2023-12-14
Apply overrides from the prefix / generic testenv

hashar:
An override to `testenv` is not applied to more specific testenv when they define the same setting, even if one asks to append.

Given a `tox.ini`:

    [testenv]
    set_env = WHERE=top

    [testenv:test]
    set_env = WHERE=test

If one uses `tox -x testenv.set_env+=CI=1`, the `CI=1` is appended to the generic testenv:

    $ tox config -k set_env
    set_env =
	WHERE=top
	CI=1

But is not seen in the more specific `testenv:test`:

    $ tox config -k set_env -e test
    set_env =
	WHERE=test

That is due to the override being set to `testenv` and never looked up by `testenv:test`.

Change the load overides from its prefix section.

This let one forces appending values to list regardless of what is set in the users `tox.ini`.

Ref: https://phabricator.wikimedia.org/T348871

### DIFF
--- a/src/tox/config/source/ini.py
+++ b/src/tox/config/source/ini.py
@@ -54,7 +54,11 @@ class IniSource(Source):
             return IniLoader(
                 section=section,
                 parser=self._parser,
-                overrides=override_map.get(section.key, []),
+                overrides=override_map.get(
+                    section.key,
+                    # Fallback to overrides from the prefix
+                    override_map.get(section.prefix, []),
+                ),
                 core_section=self.CORE_SECTION,
                 section_key=key,
             )


hashar:
I need to add unit tests to verify and cover the behavior.

gaborbernat:
This doesn't sound to me like a bug. Sounds like it's working as expected, perhaps you just misunderstood the tool. The base environment is only inherited if you did not set it in the specific environment or if you're explicitly requested it.

hashar:
> The base environment is only inherited if you did not set it in the specific environment or if you're explicitly requested it.

I got that part. What I am trying to achieve is to port tox v3 `TOX_TESTENV_PASSENV`. On the discussion at #2681, it is proposed to use `TOX_OVERRIDE=testenv.passenv=A,B,C` but  if any testenv has a `passenv` already, A, B and C will not be passed.

As indicated yesterday, I wrote tests covering the behavior I am expecting https://github.com/tox-dev/tox/compare/main...hashar:tox:get_override_from_parent_section#diff-da9ba02e176d428e326d3f0bc316309d58d1607b2d42ad0c694fc24b1d456fc2

gaborbernat:
> A, B and C will not be passed.

Intentionally. I don't think that happened on tox 3 either, and if it did, it was a bug. What you're proposing would be a bug.

hashar:
> >    A, B and C will not be passed.
> 
> Intentionally. I don't think that happened on tox 3 either, and if it did, it was a bug

It definitely worked with tox v3 and that was the purpose of `TOX_TESTENV_PASSENV` which would let us pass environment **in addition** of whatever tox.ini test environments defines.

Given a `tox.ini` defining a `BRA` in `testenv` and in a more specific `py311`:
```ini
[tox]
skipsdist = true

[testenv]
setenv = BRA=42

[testenv:py311]
setenv = BRA=311zzzz

commands = /usr/bin/env
allowlist_externals = /usr/bin/env
```

If I invoke tox 3, I get `BRA` defined as expected:
```
$ tox3
py311 run-test: commands[0] | /usr/bin/env
BRA=311zzzz
```

And if I defined a `WIKIMEDIA` environment variable and pass it to `TOX_TESTENV_PASSENV` to let it through:
```
$ WIKIMEDIA=1 TOX_TESTENV_PASSENV=WIKIMEDIA tox3
py311 run-test: commands[0] | /usr/bin/env
WIKIMEDIA=1
BRA=311zzzz
```

The variable got added next to the built-in default and `BRA`, which is fined in the testenv is still added.

The code in the legacy branch is https://github.com/tox-dev/tox/blob/b26fd5d/src/tox/config/__init__.py#L782-L809  which is that the built-in default are set and the variables from `TOX_TESTENV_PASSENV` get added to them and thus can pass from the host to the test environment.


> Did you try using TOX_OVERRIDE=testenv.test.passenv=A,B,C instead to override the test directly?

That would work for the `test` testenv. The issue is when you want to pass those environment variable to ALL testenv, given they are typically defined by developers and either out of one control or easy to not keep in sync.

gaborbernat:
Be that so, it's not supported in tox 4. I recommend using an override that sets it not the base but the actual run env you want to override. You are welcome to open a PR to https://tox.wiki/en/4.11.4/upgrading.html as a breaking change. This was unsupported in tox 3, but worked by chance IMHO. Either way, we don't plan to support it in tox 4 as part of the core (you likely can patch it in via a plugin if you must).

PR #3192, opened 2024-01-18
Document usage of Pytest with Tox as a FAQ entry

### LINKED ISSUE #3187: Document that temp dir tweaks should be applied when running Pytest in parallel with Tox

Parallel runs of Pytest are supported only when the base temporary directory is unique for each concurrent run. For Tox 3, this is documented by way of an example here: <https://tox.wiki/en/3.27.0/example/pytest.html#extended-example-change-dir-before-test-and-use-per-virtualenv-tempdir>

I appreciate that this really is a feature of Pytest and not Tox. However, since Tox provides a unique temporary directory per virtualenv out of the box, it may be helpful to document how this can be used to run Pytest with Tox's `--parallel` option.

I don't know **where** within the Tox 4 documentation pages this should be included though...

## Alternative Solutions

Do nothing: keep the documentation as is.

## Additional context

None.

faph:
Fixes #3187 

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -399,7 +399,7 @@ environment as provided by ``tox``:
     [testenv]
     commands = pytest --basetemp="{env_tmp_dir}"
 
-Setting the ``pytest`` ``--basetemp`` arguments also causes all temporary ``pytest`` files to be deleted immediately
+Setting the ``pytest`` ``--basetemp`` argument also causes all temporary ``pytest`` files to be deleted immediately
 after the tests are completed. To restore the default ``pytest`` behavior to retain temporary files for the most recent
 ``pytest`` invocations, the system's temporary directory location could be configured like this instead:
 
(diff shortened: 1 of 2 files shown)


faph:
@gaborbernat Let me know what you think. This is loosely based on the Tox 3 docs with some additional detail in places.

docs/faq.rst
```diff
@@ -366,3 +366,48 @@ If you need to test against e.g. Python 2.7, 3.5 or 3.6, you need to add the fol
...
+    commands = pytest --basetemp="{env_tmp_dir}"
+
+Setting the ``pytest`` ``--basetemp`` argument also causes all temporary ``pytest`` files to be deleted immediately
+after the tests are completed. To restore the default ``pytest`` behavior to retain temporary files for the most recent
+``pytest`` invocations, the system's temporary directory location could be configured like this instead:
+
+.. code-block:: ini
+
+    [tox]
+    setenv =
```
gaborbernat, line 409: setenv -> set_env

gaborbernat (changes requested):
Otherwise looks good.

docs/faq.rst
faph: Done

faph:
Thanks!

PR #3194, opened 2024-01-19
docs/config.rst: State in config directive sections their ini file sections

0cjs:
[Full description in the commit message(s).]

I did not copy the commit messages into the pull request because [Say everything once and only once][once].

The tests validating the fix are `tox -e docs`; these pass. (And these are useful; they caught an RST syntax error.)

I did not add a new fragment to the changelog because this seems too small to bother readers of that file with.

The full set of tests (`tox`) has at least one failure; this occurs on the `main` branch also.

[once]: https://www.oreilly.com/library/view/extreme-programming-installed/0201708426/0201708426_art03lev1sec3.html

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -102,6 +102,8 @@ examples first and use this page as a reference.
 Core
 ----
 
+The following options are set in the ``[tox]`` section of ``tox.ini`` or the ``[tox:tox]`` section of ``setup.cfg``.
+
 .. conf::
    :keys: requires
    :default: <empty list>
@@ -236,6 +238,8 @@ Python language core options
 
 tox environment
 ---------------
+
+The following options are set in the ``[testenv]`` or ``[testenv:*]`` sections of ``tox.ini`` or ``setup.cfg``.
 
 Base options
 ~~~~~~~~~~~~
--- a/docs/changelog/3194.doc.rst
+++ b/docs/changelog/3194.doc.rst
@@ -1,0 +1,1 @@
+Configuration: state in config directive sections their ini file sections - by :user:`0cjs`.


docs/config.rst
```diff
@@ -102,6 +102,9 @@ examples first and use this page as a reference.
 Core
 ----
 
+The following options are set in the ``[tox]`` section of ``tox.ini`` or
```
gaborbernat, line 105: Can you please align to 120 characters?

gaborbernat (changes requested):

docs/config.rst
0cjs: Done.

This is not standard for Python, your linter doesn't check for it, but the note about this is a bit buried in your development docs. Do you want a PR to bring this a little more to the fore so that new developers don't miss it?

gaborbernat: What do you have in mind?

gaborbernat (approved):

docs/config.rst
```diff
@@ -102,6 +102,8 @@ examples first and use this page as a reference.
 Core
 ----
 
+The following options are set in the ``[tox]`` section of ``tox.ini`` or the ``[tox:tox]`` section of ``setup.cfg``.
```
gaborbernat, line 105: For what's worth, part of the reason we did not document this here is that people can add new configuration formats via plugins, or via `toxfile.py` when these don't exists but the settings are still valid.

0cjs: Ok, that's good to know. I think it's still worth more clearly documenting the standard mechanisms, even if we don't cover the alternative mechanisms yet, but I will prepare another PR to work on that issue.

0cjs:
Ok, it's not at all clear to me what's going on with the "timeline protection" thing (maybe requiring a changelog entry for every commit? Why not just generate that from the commit message, then?), and your developer documentation doesn't mention anything about it, so I'll just let you sort the rest of that out. Also, the way that `docs/_draft.rst` works is not at all obvious, and is also not mentioned in your documentation.

Regarding the various open conversations here, I'll cons up PRs for them once this one goes through.

docs/config.rst
0cjs: As a starting point for discussion, I've put an initial go at changes on my `dev/cjs/24a20/doc-config-mechanisms` branch. I am far from sure I understand yet what you're saying about configuration mechanisms, but this is what I've got so far.

0cjs: For a start, it should be moved up to the top of the "style guide" section, since your linter does not flag standard Python style as wrong. I'll think about it further after dealing with some of the other docs stuff.

docs/config.rst
0cjs: As a starting point for discussion, I've put an initial go at further changes related to this on my dev/cjs/24a20/doc-config-mechanisms branch. I am far from sure I understand yet what you're saying about configuration mechanisms, but this is what I've got so far.

gaborbernat:
Yes, you do need to create a changelog file for every pull request. The commit messages serve a different purpose so it's not advisable to generate it from it. PR improving the contribution guides is also welcome.

0cjs:
Ok. Small documentation changes are not the sort of thing I'd normally want to see in a changelog, but each to his own.

I do have a concern about the issue number. if I include the one that mentions this, it will not close the issue when the fix is merged, will it? (This PR does not solve the issue.) If it does, should I create a separate issue referencing this first, or leave the issue number out of the changelog entry?

gaborbernat:
You can also use the car PR number as the issue number.

0cjs:
I'm not clear on why merging is blocked on this. The message says only, "Merging can be performed automatically with 0 approving reviews," and the other two sections ("All conversations are resolved" and "All checks have passed") are green. Is there something else I should be doing here?

gaborbernat (approved):

PR #3197, opened 2024-01-23
docs/development: Summarise important points experienced devs need to know

0cjs:
See commit message for details.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -1,6 +1,35 @@
 Development
 ===========
 
+Key points
+----------
+
+If you're already experienced with submitting GitHub PRs to open-source Python projects, the following are the key
+points you need to know about this project. (If you're not, you should carefully read all the documentation after this
+section. This section contains only highlights; it's not a substitute for reading this entire file.)
+
+- Check the `style guide <#style-guide>`_ below. Particularly note that lines should be wrapped at 120 characters, not
+  the PEP-8 standard of 79. The linter will not catch short lines, but most other formatting details should be caught by
+  the tests and linting below.
+- Documentation is in `RST <https://docutils.sourceforge.io/docs/user/rst/quickref.html>`_ format; beware the
+  differences from GitHub Markdown. The tox ``docs`` and ``fix`` targets will catch only some RST errors; documentation
+  changes *must* be checked visually (see below).
+- All PRs require a changelog entry. This should reference an issue if it closes that issue, otherwise reference the PR.
+  Create one or more (if there's more than one issue) ``docs/changelog/####.{feature,bugfix,doc,removal,misc}.rst`` per
+  the `changelog entry <#changelog-entries>`_ section below.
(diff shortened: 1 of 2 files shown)


docs/development.rst
```diff
@@ -1,6 +1,37 @@
...
+points you need to know about this project. (If you're not, you should carefully read all the documentation after this
+section. This section contains only highlights; it's not a substitute for reading this entire file.)
+
+- Check the `style guide <#style-guide>`_ below. Particularly note that lines should be wrapped at 120 characters, not
+  the PEP-8 standard of 79. The linter will not catch short lines, but most other formatting details should be caught by
+  the tests and linting below.
+- Documentation is in `RST <https://docutils.sourceforge.io/docs/user/rst/quickref.html>`_ format; beware the
+  differences from GitHub Markdown. The tox ``docs`` and ``fix`` targets will catch only some RST errors; documentation
+  changes *must* be checked visually (see below).
+- PRs should `use multiple commits <#submitting-pull-requests>`_ to separate independent changes, refactoring,
```
gaborbernat, line 17: This part is not true 👍 We just require one issue/feature per PR.

0cjs: Ah, I think I misread the "Submitting pull requests" section. So, just to confirm, if I have a single feature change to a piece of code that involves fixing its formatting, refactoring it slightly, and then implementing the change, you want three separate PRs for that?

And I guess, similarly here, you want me to split this into two separate PRs, one for the summary and one for the sample changelog entry and `:user:` doc. (Or even three, splitting that second one into two?)

gaborbernat: That all should be a single PR.

0cjs: Not sure which of the two questions above you're answering, but I'll assume for both cases you want a single PR. And I agree. But you still want the formatting change, refactoring and feature change as separate commits within that PR for easier review and testing, right? (That was what I was trying to say above; I'm open to suggestions for better wording.) Or should each PR contain only a single commit?

gaborbernat: - [ ] I personally don't care about separate commits. The authors preference.

0cjs: Ok, I've just taken out that bullet point, then.

But looking again at what you've written above ("That all should be a single PR"), that seems directly contradictory to your current documentation:

> Additionally, avoid including "cosmetic" changes to code that is unrelated to your change, as these make reviewing the PR more difficult. Examples include re-flowing text in comments or documentation, or addition or removal of blank lines or whitespace within lines. Such changes can be made separately, as a "formatting cleanup" PR, if needed.

That reads to me as if a reformat of code I'm about to modify, a refactoring and then a feature addition should be three separate commits. If that's not what you mean, you might consider clarifying that. (I don't understnand what you're looking for well enough to propose a change, so I won't be doing so.)

gaborbernat:
I think you're running into https://github.com/nedbat/coveragepy/issues/1733, not much can do about it myself though.

gaborbernat (approved):

PR #3203, opened 2024-01-25
Fix `main` branch used by diff-cover to work in more situations

0cjs:
__WARNING:__ to make manual checks easier, there is a temporary "test" commit on this dev branch that must be removed before merging. (@0cjs will do this on request, or you can do it yourself if you don't want to wait.)

See the commit message for a detailed description of the change and its rationale.

To see that the change is working, you can run `tox -e py11 -- tests/config/test_of_types.py` or similar; you should see that diff-cover finds that the to-be-removed test commit has 100% coverage of its new code. There were previously no automated tests to show that diff-cover is working correctly, and I have not added any because that seems very difficult to do and not worth the cost.

Checklist:
- linter: not needed; this code is not covered by it
- descriptive PR text: not copied from the commit message because [DRY]
- tests validating fix: see above
- news fragment: per commit message, not needed or wanted because it's not a user-visible change (or change to the tox program at all)
- updated/extended documentation: there's a suggestion in the commit message, but as it requires further discussion, I am leaving that for a separate PR.

[DRY]: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,7 +34,7 @@ commands =
       --cov-report xml:{toxworkdir}{/}coverage.{envname}.xml \
       -n={env:PYTEST_XDIST_AUTO_NUM_WORKERS:auto} \
       tests --durations 5 --run-integration}
-    diff-cover --compare-branch {env:DIFF_AGAINST:origin/main} {toxworkdir}{/}coverage.{envname}.xml
+    diff-cover --compare-branch {env:DIFF_AGAINST:main@{upstream}} {toxworkdir}{/}coverage.{envname}.xml
 
 [testenv:fix]
 description = format the code base to adhere to our styles, and complain about what we cannot do automatically
--- a/src/tox/config/of_type.py
+++ b/src/tox/config/of_type.py
@@ -16,6 +16,10 @@ if TYPE_CHECKING:
 T = TypeVar("T")
 V = TypeVar("V")
 
+def unused_function(x: int, y: int) -> int:
+    ''' Temporary addition to test coverage. '''
+    z = x + y
+    return z
 
 class ConfigDefinition(ABC, Generic[T]):  # noqa: PLW1641
     """Abstract base class for configuration definitions."""
--- a/src/tox/config/of_type.py
+++ b/src/tox/config/of_type.py
@@ -16,10 +16,11 @@ if TYPE_CHECKING:
 T = TypeVar("T")
 V = TypeVar("V")
 
+
 def unused_function(x: int, y: int) -> int:
-    ''' Temporary addition to test coverage. '''
-    z = x + y
-    return z
+    """Temporary addition to test coverage."""
+    return x + y
+
 
 class ConfigDefinition(ABC, Generic[T]):  # noqa: PLW1641
     """Abstract base class for configuration definitions."""
--- a/tests/config/test_of_types.py
+++ b/tests/config/test_of_types.py
@@ -3,6 +3,12 @@ from __future__ import annotations
 from tox.config.of_type import ConfigConstantDefinition, ConfigDynamicDefinition
 
 
+def test_unused_function():
+    ''' Temporary addition to test coverage. '''
+    from tox.config.of_type import unused_function
+    assert 5 == unused_function(2,3)
+
+
 def test_config_constant_eq() -> None:
     val_1 = ConfigConstantDefinition(("key",), "description", "value")
     val_2 = ConfigConstantDefinition(("key",), "description", "value")
--- a/tests/config/test_of_types.py
+++ b/tests/config/test_of_types.py
@@ -3,10 +3,11 @@ from __future__ import annotations
 from tox.config.of_type import ConfigConstantDefinition, ConfigDynamicDefinition
 
 
-def test_unused_function():
-    ''' Temporary addition to test coverage. '''
+def test_unused_function() -> None:
+    """Temporary addition to test coverage."""
     from tox.config.of_type import unused_function
-    assert 5 == unused_function(2,3)
+
+    assert unused_function(2, 3) == 5
 
 
 def test_config_constant_eq() -> None:


tox.ini
```diff
@@ -34,7 +34,7 @@ commands =
       --cov-report xml:{toxworkdir}{/}coverage.{envname}.xml \
       -n={env:PYTEST_XDIST_AUTO_NUM_WORKERS:auto} \
       tests --durations 5 --run-integration}
-    diff-cover --compare-branch {env:DIFF_AGAINST:origin/main} {toxworkdir}{/}coverage.{envname}.xml
```
gaborbernat, line 37: This worked for me without forking 🤔

0cjs:
Checks notes:
- The timeline protection needs to be removed unless you really want a changelog entry.
- The pre-commit.ci checks will pass when the test commit is removed; they're formatted the way they are to make it easier to see what they're doing while they're there.
- Tried to fix the type check problem, but I'm running into issues with a robot heavily modifying my code. I'll look into this.

tox.ini
0cjs: Exactly! You're not creating fork repos for your development work here, so you don't see the issue. (Additionally, you choose the name `origin` for your one remote, or you would see the problem anyway.)

Note that because of the above, this fix will neither help or harm you. I did explain, in the commit message, the situation that other developers are facing; I'm happy to try to clarify that explanation if it doesn't make sense.

gaborbernat: I do create it. But then I set this environment variable to upstream/main and all works. I imagine most people don't create forks, therefore, the default caters to them.

0cjs: Perhaps a better way of pointing out the effect of the fixes is the following, which I've added to the commit message:

```
    These changes affect developers in the following way:
    - Those with push access to the upstream rep (tox-dev/tox.git) see no
      difference.
    - Those who fork and pay attention to which `main` branch they're using in
      their local repos have an increased chance of diff-cover working.
    - Those who fork and don't pay attention to that may see an improvement or
      may see things break, but if it does break they have a bigger problem to
      fix anyway.
```

0cjs: Most people don't have push access to your repo, so most people _do_ create forks, because a forked repo is the only place they can push their change to create a PR. Note that this fork is not one intended to maintain changes separate from the upstream; it's merely a staging area for their development branches.

I'm not clear what you mean by "I do create it." You said in your first message that you're not forking. What's the output of `git remote -v` in your development repos? Does it show two remotes, or just one? Mine is:

```
$ git remote -v
r	https://github.com/0cjs/tox.git (fetch)
r	git@github.com.:0cjs/tox.git (push)
upstream	https://github.com/tox-dev/tox.git (fetch)
upstream	git@github.com.:tox-dev/tox.git (push)
```

gaborbernat: Yes they do create forks. However, they don't tend to add the upstream repository as a remote. Instead for most people they were just cloning their fork. So the fork is the origin for most.

0cjs: Right. Those people will be unaffected by this change; they'll still be using an outdated `main` after new changes come on to the upstream `main`. This change helps those who _do_ either have their local main instead track the upstream `main` or who manually update their fork's `main` from the upstream `main`.

gaborbernat: Not if they create a branch after forking and work off that branch. Then their main will be true diff compared to their branch...

0cjs: As a side note, the commit message is intended to cover everything you're talking about here; I am open to suggestions for changes to it if you didn't find it clear enough.

0cjs: Yes, "if they create a branch after forking and work off that branch" (which is the situation I'm assuming here). They will be diffing not against the latest upstream main in `tox-dev/tox.git`, but against the older version of `main` in their fork repo, which dates from the time of the fork and is not updated after that. (Unless they manually update it, in which case this commit works for them anyway.)

gaborbernat: In my mind that's fine and desirable. The intent is not to diff upstream main, but more like their new changes 🤔

0cjs: Well first, that's not how it works for you in your no-fork-repo situation: you're diffing against upstream main, not the commit where you branched from main. This change does not affect that. It does fix breakage for developers in your position who wish to use a remote name other than `origin`.

For the situation you said is the common one, where people create a fork repo and use only that as a remote, this also doesn't change things; they will be diffing against their remote's `main` (which is presumably where they branched from). This is the behaviour you say you want, and this change maintains that. However, it also provides an improvement for them, again fixing the current breakage when someone uses a remote name other than `origin`.

So to summarise:
- Non-fork developers who name the remote `origin`: no change
- Non-fork developers who use a different remote name: fix currently broken behaviour.
- Fork-only developers who name the remote `origin`: no change
- Fork-only developers who use a different remote name: fix currently broken behaviour.
- Multi-remote developers: often fix currently broken behaviour; make using multiple remotes easier and more reliable.

0cjs: Note that if you really want everybody to be diffing against where the dev branch diverged from `main`, that can be done, but requires some extra code to determine where that point is. I have a shell function, `mbase()` that can find this point, though it's many years old and probably wants to be rewritten to use modern Git facilities. (And probably have tests, for that matter.) I'm not interested in a PR to do that myself, but if we can resolve our current situation here I'd be happy to open an issue describing the problem and offering what code I have.

gaborbernat: I don't see how your change is a no change to my last use case.

gaborbernat: That being said I'm -0 on this, so perhaps other maintainers can chime in what they prefer.

0cjs: > I don't see how your change is a no change to my last use case.

I don't know which use case is your "last use case." Can you explain what the case is, and what change this makes to it?

0cjs: It also occurs to me, if you want to be diffing against where the dev branch split from main, rather than against changes since the split, wouldn't you be more likely to get that with `main` than `origin/main`?

0cjs: @gaborbernat Is my "last use case" comment above something that you're still thinking about? Should I be poking you if I don't get a reply to queries like this within a day or two?

gaborbernat: cc @jugmac00

gaborbernat:
Closing as seems it stalled, we can reopen if you get time again for it.

Issue #3199, opened 2024-01-23
Multiple -e options behave differently in tox 4.x vs tox 3.x


rhertzog:
## Issue

`tox -e env1 -e env2` runs the two environments in tox 3.x but runs only env2 in tox 4.x.

I don't know if that change is intentional but it's sure to break quite a few CI that rely on `-e` being additive rather than overriding a former value. I would suggest to either fix the behaviour to be consistent with 3.x or put some strong warning in the changelog and documentation that one needs to use the comma separated syntax: `tox -e env1,env2`

## Environment

Provide at least:

- OS: Debian Linux (unstable)

```console
$ tox --version
4.12.0 from /usr/lib/python3/dist-packages/tox/__init__.py
```

Thank you for maintaining tox! It's a great tool.

webknjaz:
v4 was a huge multi-year rewrite with a lot of breaking changes. I'm pretty sure most are expected. Probably worth mentioning @ https://tox.wiki/en/4.12.1/upgrading.html#updating-usage-with-e though.

jugmac00:
I can confirm that this has changed. To me it looks like a regression and I see no obvious reason why we should not re-introduce it for tox 4.

gaborbernat:
A PR fixing this would be accepted.

0cjs:
I've not dug into the old v3 code, but I'm assuming that it did something along the lines of `add_argument('-e', ..., action='append', default=[])`. I.e., it collected all of the `-e` arguments into a list, and that list was used directly as the list of environment names. This would not allow a comma-separated list such as `-e foo,bar`; that would produce a single envname `foo,bar`. 

The current code I suppose introduced the `CliEnv` class (`session/env_select.py`), used as `add_argument("-e", ..., type=CliEnv)`. There are no direct tests or documentation for `CliEnv`, but I am guessing that `CliEnv(s)` converts the comma-separated envnames to a list[str] of the envnames.

The issue here is that with the argument value being a single `CliEnv` value, using `-e` will wipe out any previous values supplied with `-e`.

argparse doesn't offer any way to have a `type=` parameter work on both the current value and any independent previous values for that argument, as far as I can tell, so the only option to fix this seems to be to go back to having the value be a list (or really, a mutable sequence) so that we can use `action='append'`.

[...]

PR #3204, opened 2024-01-26
Add tests for CliEnv

0cjs:
This is a start on issue #3199. The plan for that involves changing CliEnv; this starts documenting its current behaviour by adding tests for it.

The final reformatting commit may need to be squashed into the previous commit; see that commit message for details.

No news fragment because this isn't a user-visible change. No documentation yet because that will be informed by review of this commit.

### DIFF
--- a/tests/session/test_env_select.py
+++ b/tests/session/test_env_select.py
@@ -17,25 +17,26 @@ CURRENT_PY_ENV = f"py{sys.version_info[0]}{sys.version_info[1]}"  # e.g. py310
 
 
 @pytest.mark.parametrize(
-    ("user_input",      "env_names",            "is_all",   "is_default" ), [
-    (None,              (),                     False,      True         ),
-    ("",                (),                     False,      True         ),
-    ("a1",              ("a1",),                False,      False        ),
-    ("a1,b2,c3",        ("a1", "b2", "c3"),     False,      False        ),
-    #   If the user gives "ALL" as any envname, this becomes an "is_all" and other envnames are ignored.
-    ("ALL",             (),                     True,       False        ),
-    ("a1,ALL,b2",       (),                     True,       False        ),
-    #   Zero-length envnames are ignored as being not present. This is not intentional.
-    (",,a1,,,b2,,",     ("a1", "b2"),           False,      False        ),
-    (",,",              (),                     False,      True         ),
-    #   Environment names with "invalid" characters are accepted here; the client is expected to deal with this.
-    ("\x01.-@\x02,xxx", ("\x01.-@\x02", "xxx"), False,      False        ),
+    ("user_input", "env_names", "is_all", "is_default"),
+    [
+        (None, (), False, True),
+        ("", (), False, True),
+        ("a1", ("a1",), False, False),
+        ("a1,b2,c3", ("a1", "b2", "c3"), False, False),
+        #   If the user gives "ALL" as any envname, this becomes an "is_all" and other envnames are ignored.
+        ("ALL", (), True, False),
+        ("a1,ALL,b2", (), True, False),
+        #   Zero-length envnames are ignored as being not present. This is not intentional.
+        (",,a1,,,b2,,", ("a1", "b2"), False, False),
+        (",,", (), False, True),
+        #   Environment names with "invalid" characters are accepted here; the client is expected to deal with this.
+        ("\x01.-@\x02,xxx", ("\x01.-@\x02", "xxx"), False, False),
     ],
 )
 def test_clienv(user_input: str, env_names: tuple[str], is_all: bool, is_default: bool) -> None:
     ce = CliEnv(user_input)
-    assert (ce.is_all, ce.is_default_list, tuple(ce)) \
-        == (is_all,    is_default,         tuple(env_names))
+    assert (ce.is_all, ce.is_default_list, tuple(ce)) == (is_all, is_default, tuple(env_names))
+
 
 @pytest.mark.parametrize(
     ("user_input", "expected"),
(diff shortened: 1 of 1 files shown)


tests/session/test_env_select.py
```diff
@@ -15,6 +15,57 @@
...
+        ("ALL", (), True, False),
+        ("a1,ALL,b2", (), True, False),
+        #   Zero-length envnames are ignored as being not present. This is not intentional.
+        (",,a1,,,b2,,", ("a1", "b2"), False, False),
+        (",,", (), False, True),
+        #   Environment names with "invalid" characters are accepted here; the client is expected to deal with this.
+        ("\x01.-@\x02,xxx", ("\x01.-@\x02", "xxx"), False, False),
+    ],
+)
+def test_clienv(userinput: str, envnames: tuple[str], isall: bool, isdefault: bool) -> None:
```
gaborbernat, line 39: Please use the snake case for variable names.

0cjs: Done.

tests/session/test_env_select.py
```diff
@@ -15,6 +15,57 @@
 
 CURRENT_PY_ENV = f"py{sys.version_info[0]}{sys.version_info[1]}"  # e.g. py310
 
+####################################################################
```
gaborbernat, line 18: I'm -1 on adding these markers 🤔

tests/session/test_env_select.py
```diff
@@ -15,6 +15,57 @@
...
+def test_env_select_lazily_looks_at_envs() -> None:
+    state = State(get_options(), [])
+    env_selector = EnvSelector(state)
+    # late-assigning env should be reflected in env_selector
+    state.conf.options.env = CliEnv("py")
+    assert set(env_selector.iter()) == {"py"}
+
+
+####################################################################
+#   Tests of env_select functionality as used through other code.
```
gaborbernat, line 67: I'm -1 on adding these markers 🤔

tests/session/test_env_select.py
```diff
@@ -15,6 +15,57 @@
...
+        ("a1,ALL,b2", (), True, False),
+        #   Zero-length envnames are ignored as being not present. This is not intentional.
+        (",,a1,,,b2,,", ("a1", "b2"), False, False),
+        (",,", (), False, True),
+        #   Environment names with "invalid" characters are accepted here; the client is expected to deal with this.
+        ("\x01.-@\x02,xxx", ("\x01.-@\x02", "xxx"), False, False),
+    ],
+)
+def test_clienv(user_input: str, env_names: tuple[str], is_all: bool, is_default: bool) -> None:
+    c = CliEnv(user_input)
```
gaborbernat, line 40: please don't use one character variable names 👍

gaborbernat (changes requested):

tests/session/test_env_select.py
0cjs: Given that it's the _only_ variable in that entire function, which is only three lines long, making it a longer variable name adds only verbosity. (There's a reason that, even though we now can have very long variable names, we say `for i in ...` rather than `for index_of_list ...`.) But I have changed it as you requested.

tests/session/test_env_select.py
0cjs: Well, I'm focused here on making this code quicker to understand for people not as familiar with it as I guess you are. If you insist, I'll remove them, but having this when I started working on this file would have saved me some time and made my life easier.

tests/session/test_env_select.py
0cjs: Will be resolved with resolution of lines 18-19.

tests/session/test_env_select.py
gaborbernat: I insist 😊 If we need some extra information, let's add it to the docstring of the test IMHO.

0cjs: Ok. Please next time say if you have to have something changed, rather than just saying you're "-1". Those lines are removed.

Putting info in the docstrings doesn't serve the purpose. The point of grouping things is that you can quickly see the groupings and understand something common to all the elements of the group without having to look at each individual element.

gaborbernat: A better solution in pytest to achieve that is https://stackoverflow.com/questions/50016862/grouping-tests-in-pytest-classes-vs-plain-functions, if that's your intent.

0cjs: That, again, is (IMHO, of course) extra complexity and extra verbosity to do something a simple comment can do. You'll notice that neither of the "two compelling use cases" mentioned in the top-voted answer is related to my purpose here.

That said, thanks for pointing this out: though it doesn't solve the problem I'm trying to solve here, the ability to parametrize a group of tests that particular way is interesting. (I am not sure if it's any better than just assigning the parameters to a variable and re-using that, but it's a technique worth considering next time I run into a need for something along those lines.)

0cjs:
I'm not clear on why this PR is still saying that there are changes requested. I clicked "See review"  and it just takes me to requests that are from before I pushed up the fixes and that are marked outdated and resolved.

tests/session/test_env_select.py
gaborbernat: I'd argue that comments get stale easily, but adding a new indentation or moving it to a different file makes the intent much more explicit and hard to miss. IMHO, comments should be only used within methods to explain why we're doing something (not the how, for that, the code should be self-documenting).

gaborbernat:
> I'm not clear on why this PR is still saying that there are changes requested.

Until I approve, will stay in that status. That's how GitHub works.

gaborbernat (approved):

0cjs:
> Until I approve, will stay in that status. That's how GitHub works.

Ah, ok! I've seen something else similar; GitHub keeps saying things that make me wonder if I missed fixing something. But I'll just stop worrying about it. Please do poke me if you see a PR where I've responded to things but there are still things that I need to fix, thoug.

PR #3216, opened 2024-02-06
pyproject.toml: Move certain tool.ruff settings to tool.ruff.lint

0cjs:
tool.ruff.{select,isort,ignore,per-file-ignores} will soon be deprecated in favour of putting these under tool.ruff.lint. (See reverted PR #3214 for the complaint that appears in a more recent version of ruff.) Change these now to avoid complaints later.

- [n/a] ensured there are test(s) validating the fix
- [n/a] added news fragment in `docs/changelog` folder
- [n/a] updated/extended the documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,11 +103,11 @@ build.targets.sdist.include = ["/src", "/tests", "/tox.ini"]
 version.source = "vcs"
 
 [tool.ruff]
-select = ["ALL"]
+lint.select = ["ALL"]
 line-length = 120
 target-version = "py38"
-isort = { known-first-party = ["tox", "tests"], required-imports = ["from __future__ import annotations"] }
-ignore = [
+lint.isort = { known-first-party = ["tox", "tests"], required-imports = ["from __future__ import annotations"] }
+lint.ignore = [
   "CPY",     # No copyright header
   "INP001",  # no implicit namespaces here
   "D",       # ignore documentation for now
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -132,7 +132,7 @@ ignore = [
 ]
 format.preview = true
 lint.preview = true
-[tool.ruff.per-file-ignores]
+[tool.ruff.lint.per-file-ignores]
```
gaborbernat, line 135: Please use the previous format, don't create new section.

gaborbernat (changes requested):

pyproject.toml
0cjs: I'm not clear if I'm supposed to be addressing this; the conversation was marked as "resolved" before I even saw it.

In case I am, I'm not clear on what you're talking about above. I didn't create a new section; I just renamed an existing section.

pyproject.toml
```diff
@@ -103,11 +103,11 @@ build.targets.sdist.include = ["/src", "/tests", "/tox.ini"]
 version.source = "vcs"
 
 [tool.ruff]
-select = ["ALL"]
+lint.select = ["ALL"]
```
gaborbernat, line 106: move this to the lint group below

gaborbernat:
Superseeded by https://github.com/tox-dev/tox/pull/3226

Issue #3219, opened 2024-02-15
Partial match on testenv results in use of base testenv


stephenfin:
## Issue

Consider the following toy `tox.ini`:

```ini
[testenv]
skip_install = true
allowlist_externals =
  echo
commands =
  echo "base testenv"

[testenv:functional{-py310}]
commands =
  echo "functional testenv"
```

This _should_ result in a single testenv `functional-py310` (there's only one factor) plus the default `pyNN` testenvs. Attempts to use a testenv with the `functional` factor but a different `pyNN` factor (e.g. `tox -e functional-py312`) or without a `pyNN` factor (e.g. `tox -e functional`) should fail since it doesn't match an env making it no different to e.g. `tox -e doesnotexist`.

In tox 3, this was the case:

```console-session
❯ virtualenv .venv --python=python3.8
❯ source .venv/bin/activate
❯ pip install 'tox<3'

❯ tox -e functional-py310 -q
functional testenv
__________________________________________________________________ summary __________________________________________________________________
  functional-py310: commands succeeded
  congratulations :)

❯ tox -e functional-py312 -q
ERROR: unknown environment 'functional-py312

❯ tox -e functional -q
ERROR: unknown environment 'functional'

❯ tox -e doesnotexist -q
ERROR: unknown environment 'doesnotexist'
```

Once we switch to tox 4, this is no longer the case:

```console-session
❯ tox -e functional-py310 -q
functional testenv
  functional-py310: OK (0.22 seconds)
  congratulations :) (0.29 seconds)

❯ tox -e functional-py312 -q
base testenv
  functional-py312: OK (0.13 seconds)
  congratulations :) (0.18 seconds)

❯ tox -e functional -q
base testenv
  functional: OK (0.14 seconds)
  congratulations :) (0.20 seconds)
```

Totally undefined envs fail as expected (though with an uglier error than previously):

```console-session
❯ tox -e doesnotexist -q
ROOT: HandledError| provided environments not found in configuration file:
doesnotexist
```

This can result in us running a wholly different set of tests/commands if e.g. a user uses a `pyNN` factor that is not defined inline or does not use a factor (with the idea being you'd use the Python version tox is installed under).

## Environment

Using latest tox 3 and 4 available from `pip` (currently 3.28.0 and 4.12.1, respectively).

## Minimal example

See above.

stephenfin:
I'd like to know what is the expected behaviours in the below:

```
[testenv]
skip_install = true
allowlist_externals =
[...]

PR #3220, opened 2024-02-15
Consider complete envs, not factors

stephenfin:
To determine whether we have been given a valid *env*, we compare the *factors* of the *env* to our known set of *factors*. If a testenv contains a series of valid factors, we assume the testenv is valid. This is an incorrect assumption: a testenv is only valid if it defined in the configuration file *or* is one of the "special" auto-generated testenvs like e.g. 'py310'. Correct this mistake.

This is a draft PR while I wait for feedback on my general idea here. I'd like to confirm that my understanding of expected behavior is correct.

Signed-off-by: Stephen Finucane <stephen@that.guru>
Closes: #3219

### DIFF
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -136,8 +136,20 @@ class _ToxEnvInfo:
     package_skip: tuple[str, Skip] | None = None  #: if set the creation of the packaging environment failed
 
 
-_DYNAMIC_ENV_FACTORS = re.compile(r"(pypy|py|cython|)((\d(\.\d+(\.\d+)?)?)|\d+)?")
-_PY_PRE_RELEASE_FACTOR = re.compile(r"alpha|beta|rc\.\d+")
+_DYNAMIC_ENV_FACTORS = re.compile(
+    r"""
+    (pypy|py|cython|)             # interpreter
+    (
+        (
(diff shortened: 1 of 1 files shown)


stephenfin:
This is currently failing a few tests because it does not allow for autogenerated testenvs based on factors defined in e.g. `[testenv] commands`. For example:

```
[testenv]
commands =
    foo: echo "got foo"
    bar: echo "got bar"
```

That this worked suprised me: I'd normally have defined this in my `tox.ini` like so:

```
[testenv:py{310,311,312}{,-cov}]
```

but that is definitely duplication. I can work to fix this but again, I want to confirm that the bug I'm seeing is indeed a bug (it really feels like one)

stephenfin:
Converting to non-draft, even though it's clearly not complete, so that folks see it.

gaborbernat:
> Converting to non-draft, even though it's clearly not complete, so that folks see it.

We see, without it too, was waiting for you to pass the CI.

> This is an incorrect assumption: a testenv is only valid if it defined in the configuration file _or_ is one of the "special" auto-generated testenvs like e.g. 'py310'. Correct this mistake.

I'd say, in general, yes. But note, I want to allow `py313-magic` if `py312-magic` is defined. This will make testing a new Python interpreter version not require defining that in the configuration.

gaborbernat:
Closing as seems it stalled, we can reopen if you get time again for it.

PR #3260, opened 2024-04-01
Fix type-checking

stefanor:
Update test typing for build 1.2.0 which has an explicit Distribution type.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,6 +19,7 @@ from tox.tox_env.python.api import PythonInfo, VersionInfo
 from tox.tox_env.python.virtual_env.api import VirtualEnv
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -19,6 +19,7 @@
 from tox.tox_env.python.virtual_env.api import VirtualEnv
 
 if TYPE_CHECKING:
+    from build._types import Distribution
```
stefanor, line 22: @layday: This is a private module, but its describing the API for a public interface. Maybe it should be more public?

layday: It is public: `from build import DistributionType`.

stefanor: Sorry, missed that because I was looking for the same name.

gaborbernat (approved):

Issue #3287, opened 2024-05-29
TypeError: isinstance() argument 2 cannot be a parameterized generic


ssbarnea:
## Issue

Stacktrace with tox when loading plugin configuration. Based on the error displayed I believe that is not the plugin fault. The loaded configuration option is a list, like below:

```
[ansible]
skip =
    py3.7
    py3.8
    2.9
    2.10
    2.11
    2.12
    2.13
```

I am still trying to fully understand this python typing issue in order to know how to avoid it.

## Environment

Provide at least:

- OS:

<details open>
<summary>Output of <code>pip list</code> of the host Python, where <code>tox</code> is installed</summary>

```console
ROOT: Using a default tox.ini file with tox-ansible plugin is not recommended. Consider using a tox-ansible.ini file and specify it on the command line (`tox --ansible -c tox-ansible.ini`) to avoid unintentionally overriding the tox-ansible environment configurations.
Traceback (most recent call last):
  File "/Users/ssbarnea/.asdf/installs/python/3.12.3/bin/tox", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/Users/ssbarnea/.asdf/installs/python/3.12.3/lib/python3.12/site-packages/tox/run.py", line 20, in run
    result = main(sys.argv[1:] if args is None else args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ssbarnea/.asdf/installs/python/3.12.3/lib/python3.12/site-packages/tox/run.py", line 42, in main
    result = provision(state)
             ^^^^^^^^^^^^^^^^
  File "/Users/ssbarnea/.asdf/installs/python/3.12.3/lib/python3.12/site-packages/tox/provision.py", line 86, in provision
    MANAGER.tox_add_core_config(state.conf.core, state)
  File "/Users/ssbarnea/.asdf/installs/python/3.12.3/lib/python3.12/site-packages/tox/plugin/manager.py", line 79, in tox_add_core_config
    self.manager.hook.tox_add_core_config(core_conf=core_conf, state=state)
  File "/Users/ssbarnea/.asdf/installs/python/3.12.3/lib/python3.12/site-packages/pluggy/_hooks.py", line 513, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
[...]

PR #3288, opened 2024-05-29
Allow `ConfigSet.add_config` to receive parameterized generics for `of_type`.

ssbarnea:
Allow `ConfigSet.add_config` to receive parameterized generics for `of_type`.

Fixes: #3287


gaborbernat (changes requested):
I cannot accept anything without test...

ssbarnea:
> I cannot accept anything without test...

But of course! Note that I was not able to use pytest skip marks to avoid py38 because interpreter failed to load the test file itself. Still, using an if before the test definition worked fine.

gaborbernat (approved):

PR #3303, opened 2024-07-01
Add `windir` to the default list of `pass_env` variables on Windows

### LINKED ISSUE #3302: Include `windir` by default in `pass_env`

When running tools in a tox environment, some applications require that the `WINDIR` environment variable exists. The existence of that environment variables is guaranteed since Windows 95, and no application ever documents this requirement.

I've learned this through years of past experience, but bumped into this again today with this minimal reproducer `tox.ini` on Windows (assuming a proper project exists):

```
[testenv:build]
deps = poetry
commands = poetry build
```

Running this shows the following output (note the "system cannot find the path specified" messages):

```
build: commands[0]> poetry build
Building sqliteimport (0.2.0)
  - Building sdist
The system cannot find the path specified.
  - Built sqliteimport-0.2.0.tar.gz
  - Building wheel
The system cannot find the path specified.
  - Built sqliteimport-0.2.0-py3-none-any.whl
```

This is ultimately happening because [Poetry assumes that it can call `%WINDIR%\System32\where.exe` to find the git executable](https://github.com/python-poetry/poetry-core/blob/06d72bc0f1f05086a1cc0e760075c519af33db5b/src/poetry/core/vcs/git.py#L161-L163):

```python
if WINDOWS:
    # Finding git via where.exe
    where = "%WINDIR%\\System32\\where.exe"
```

While I can hunt this down in an open source Python project, this type of failure is very difficult to identify in compiled executables.

I want to add the `windir` environment variable to the list of default environment variables included in `pass_env`, because it is just as critical as `TMP` and `USERPROFILE`.

## Alternative Solutions

There's no alternative solution. Tox can add `windir` to the default list of `pass_env`, or developers will individually have to figure out for themselves how to resolve failures with some Windows applications run by tox.

kurtmckee:
This PR adds `windir` to the default list of Windows `pass_env` environment variables.

`windir` is a system environment variable that has existed since Windows 95. Like `TMP` and `USERPROFILE`, it is assumed to exist and no software documents that it should be set. It is often difficult to figure out that crashes and errors are caused by a missing `windir` environment variable.

This change allows Windows software that relies on `windir` to work out-of-the-box.

> [!NOTE]
>
> The canonical spelling of `windir` is lowercase. This PR uses an uppercase variant based on PR feedback.

Fixes #3302

### DIFF
--- a/tests/session/cmd/test_show_config.py
+++ b/tests/session/cmd/test_show_config.py
@@ -137,7 +137,9 @@ def test_pass_env_config_default(tox_project: ToxProjectCreator, stdout_is_atty:
         + (["SYSTEMDRIVE", "SYSTEMROOT", "TEMP"] if is_win else [])
         + (["TERM"] if stdout_is_atty else [])
         + (["TMP", "USERPROFILE"] if is_win else ["TMPDIR"])
-        + ["VIRTUALENV_*", "http_proxy", "https_proxy", "no_proxy"]
+        + ["VIRTUALENV_*"]
+        + (["WINDIR"] if is_win else [])
+        + ["http_proxy", "https_proxy", "no_proxy"]
     )
     assert pass_env == expected
 
--- a/docs/changelog/3302.bugfix.rst
+++ b/docs/changelog/3302.bugfix.rst
@@ -1,0 +1,1 @@
+- Add ``windir`` to the default list of Windows ``pass_env`` environment variables. - by :user:`kurtmckee`
--- a/src/tox/tox_env/api.py
+++ b/src/tox/tox_env/api.py
@@ -231,6 +231,7 @@ class ToxEnv(ABC):
                     "USERPROFILE",  # needed for `os.path.expanduser()`
                     "PATHEXT",  # needed for discovering executables
                     "MSYSTEM",  # controls paths printed format
+                    "WINDIR",  # base path to system executables and DLLs
                 ],
             )
         else:  # pragma: win32 no cover


gaborbernat (approved):

tests/session/cmd/test_show_config.py
```diff
@@ -138,6 +138,7 @@ def test_pass_env_config_default(tox_project: ToxProjectCreator, stdout_is_atty:
         + (["TERM"] if stdout_is_atty else [])
         + (["TMP", "USERPROFILE"] if is_win else ["TMPDIR"])
         + ["VIRTUALENV_*", "http_proxy", "https_proxy", "no_proxy"]
+        + (["windir"] if is_win else [])
```
gaborbernat, line 141: should be all capital

gaborbernat (changes requested):

tests/session/cmd/test_show_config.py
kurtmckee: I think I've addressed this; amended and force-pushed for a clean history.

gaborbernat (approved):

gaborbernat (changes requested):
The CI is failing...

kurtmckee:
Oof, this is harder when I'm not in front of my Windows PC to test locally. I'll update this again!

kurtmckee:
It looks like the `docs` environment is failing on Windows due to a timeout with GitHub, but if I'm reading that incorrectly and it's related to this PR please let me know. :+1:

gaborbernat (approved):

Issue #999, opened 2018-09-18
native pyproject.toml config


gaborbernat:
This is a ticket to track and implement a native ``pyproject.toml`` configuration. Will update with the proposed plan and format.

Core goals:
- explicitly inheritance based (a lot of substitution at the moment is used to replicate inheritance), 
- substitution able,
- reasonably familiar (e.g. share most concepts with ini format),
- succinct (some form of factors needs to have).

We'll use this opportunity to rename some keys (e.g. ``skipsdist`` should be ``skip_packaging``, etc).

obestwalter:
Learning from the old format, I would add:

* easy to debug (good visualization of the settings - `--showconfig`` 2.0)
* conservative error handling (err on the side of crashing early with a helpful error, rather than being too allowing for "creative" uses)

obestwalter:
... and most importantly: a clear, versioned specification that can then evolve as we for sure won't get it right at first try :)

obestwalter:
I think, It would be great if we could organize a sprint around this whole topic at some point this year.

gaborbernat:
Well, most importantly we need someone try to tackle this and implement it. Personally, in my priority queue, this is quite low. I've tackled it inside https://github.com/gaborbernat/toxn/blob/master/pyproject.toml#L5 only to prove it's quite hard to get this right. I'll only have some bandwidth for this after the #1102, #998 and #338. I also want to first split up configuration module from a mega file to sub-modules.

obestwalter:
That's exactly why I think a sprint with a few more people might be a good way to get this off the ground. Maybe around a conference where a few of as will go anyway - e.g. EuroPython.

gaborbernat:
I don't consider this a good candidate for a sprint. That creates an artificial deadline (e.g. end of sprint). Instead, we should probably propose each and every one of us our own proposal and iterate on that until we find something that can do *at least* what the ``ini`` does at the moment. With ``toxn`` I had to start over a few times just because halfway through I figured well this path makes feature x/y really hard to support/implement.

ionelmc:
Why not a `tox.toml`? `pyproject.toml` is going to be a mess (just like `setup.cfg` was when people decided to make tools that rewrite/change stuff in it).

obestwalter:
@gaborbernat 
> That creates an artificial deadline (e.g. end of sprint). 

[...]

PR #3309, opened 2024-07-14
[WIP] Native TOML config

ziima:
This is a prototype solution for #999, see discussion there.

This branch will be rebased a lot.


src/tox/config/source/toml.py
```diff
@@ -0,0 +1,142 @@
...
+                raise ValueError
+            content = path.read_text()
+        self._raw = tomllib.loads(content)
+        self._section_mapping: defaultdict[str, list[str]] = defaultdict(list)
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}(path={self.path})"
+
+    def transform_section(self, section: Section) -> Section:  # noqa: PLR6301
+        return IniSection(section.prefix, section.name)
```
gaborbernat, line 49: We should not use the ini system at all.

src/tox/config/source/toml.py
```diff
@@ -0,0 +1,142 @@
...
+        return None
+
+    def get_base_sections(self, base: list[str], in_section: Section) -> Iterator[Section]:  # noqa: PLR6301
+        for a_base in base:
+            section = IniSection.from_key(a_base)
+            yield section  # the base specifier is explicit
+            if in_section.prefix is not None:  # no prefix specified, so this could imply our own prefix
+                yield IniSection(in_section.prefix, a_base)
+
+    def sections(self) -> Iterator[IniSection]:
```
gaborbernat, line 76: It makes no sense why have anything to do with the ini system here.

src/tox/config/source/toml.py
```diff
@@ -0,0 +1,142 @@
...
+    from tox.config.sets import ConfigSet
+
+
+class TomlSource(Source):
+    """Configuration sourced from a toml file (such as tox.toml).
+
+    This is experimental API! Expect things to be broken.
+    """
+
+    CORE_SECTION = CORE
```
gaborbernat, line 34: I don't see why we should reuse the ini systems prefixes for both the core and testing environment.

tests/config/source/test_discover.py
```diff
@@ -10,7 +10,7 @@
 
 def out_no_src(path: Path) -> str:
     return (
-        f"ROOT: No tox.ini or setup.cfg or pyproject.toml found, assuming empty tox.ini at {path}\n"
+        f"ROOT: No tox.ini or tox.toml or setup.cfg or pyproject.toml found, assuming empty tox.ini at {path}\n"
```
gaborbernat, line 13: I'm not sure I agree with the need for a dedicated TOML file.  I am currently -1 on adding that.

tests/config/source/test_source_toml.py
```diff
@@ -0,0 +1,83 @@
...
+    assert known_envs
+    custom_section = config.get_section_config(
+        section=Section("custom", "foo"),
+        base=[],
+        of_type=CustomConfigSet,
+        for_env=None,
+    )
+    assert custom_section["a"] == "b"
+
+
```
gaborbernat, line 67: I'm not a fan of this long comments.

gaborbernat (changes requested):

ziima:
I've ditched `TomlLoader` and updated `MemoryLoader` instead. They were mostly the same.

src/tox/config/source/toml.py
```diff
@@ -0,0 +1,140 @@
...
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from tox.config.loader.api import OverrideMap
+    from tox.config.sets import ConfigSet
+
+
+class TomlSource(Source):
+    """Configuration sourced from a toml file (such as tox.toml).
```
gaborbernat, line 29: I do not want a dedicated toml, but should be inside pyproject.toml.

gaborbernat (changes requested):

src/tox/config/source/toml.py
ziima: I duplicated `IniSection` to `TomlSection` for the time being. I plan to do more clean up later on.

src/tox/config/source/toml.py
stinovlas: I'd prefer to have an option of separate TOML config file as well. If you don't like a dedicated file such as `tox.toml`, what about allowing to specify configuration file in env variable, such as `TOX_CONFIG`?

gaborbernat: I don't like the environment idea 😃

ziima:
I changed testenv in TOML to `tool.tox.env` as suggested in https://github.com/tox-dev/tox/issues/999#issuecomment-2231261265

Currently supported pyproject.toml looks like 
```toml
[tool.tox]
env_list = ['py311', 'py312', 'type']

# The default testenv is `testenv`.
[tool.tox.env.testenv]
deps = 'pytest'
commands = ['pytest tests']

[tool.tox.env.type]
deps = 'mypy'
commands = ['mypy src']
```

I also added support for native `pyproject.toml`. It broke ability to load legacy `pyproject.toml` using `--conf` though. To be fixed later.

I split `BaseSection` from `Section` to handle different structure of data within TOML. I'm not quite sure what to do with that at the moment.

I keep `tox.toml` support at least for tests during development. It's easier that way.

gaborbernat:
I think deps needs to be a list of strings. Also, I would like to rename the base environment name to something like testenv -> base.

tests/config/source/test_discover.py
ssbarnea: I am in favour of defaulting to `pyproject.toml` but still allowing user to define an alternative config file to load. Some might find a need for it (like CI/CD, or custom user modified one when they cannot merge their changes in the default one).

I would say that even with alternative filename, I would still expect it to follow the same sections as the `pyproject.toml`, just to keep code simple.

ziima: I went through the docs for several non-build tools I know (excluded those which do not support `pyproject.toml` at all) to see if the support custom TOML file. Here are results:

| Tool | Custom TOML file | Custom TOML root |
| ---- | --------------------- | ------------------ |
| mypy | none | |
| ruff | `ruff.toml` | none |
| bump-my-version | `.bumpversion.toml` | `tool.bumpversion` or `bumpversion` |
| isort | none | |
| black | none | |
| codespell | none | |
| coverage | none | |
| towncrier | `towncrier.toml` | `tool.towncrier` |

I'd say custom TOML file is not that rare. Custom root in TOML file seems to vary.

ziima:
Ad options: `deps` just needs a tweak to work, but there is an underlying issue regarding factors. I'm not yet sure how to handle factors in TOML. For `deps`, I could just do
```toml
deps = ["quality: ruff"]
```
but that is somewhat clumsy and it really doesn't work with non-string options, such as `ignore_errors`.

I'm starting to think the best approach would be to provide a subkey to explicitly define factors in TOML, such as

```toml
[[tox.deps]]
name = "ruff"
when = ["quality"]

[[tox.deps]]
name = "pip"
when = ["py312", "py311"]

[[tox.deps]]
name = "setuptools"

# or a shortcut of
[tox]
deps = ["ruff", "pip", "setuptools"]
# when none of the deps has custom factors.
```

Ad base environment name: I tried to respect the `base` option, which has `testenv` as default value. I haven't polish this part, but this is the goal for the moment.

gaborbernat:
IMHO we should do something similar to what hatch does for environment generation and factors.

gaborbernat:
@ziima do you plan to finalize this?

ziima:
I'm back from holiday, so I should be able to find some time to work on this again.

gaborbernat:
Great to hear so. And thank you for doing all the hard work.

gaborbernat:
Replaced by https://github.com/tox-dev/tox/pull/3353

PR #3321, opened 2024-08-11
Fix #3318 - Suppress spinner in parallel runs in CI

### LINKED ISSUE #3318: Set --parallel-no-spinner if in CI

`--parallel-no-spinner` is fairly required for CI runs to prevent the spinner from messing the output. It seems reasonable to set the flag on by default if CI is detected.

Set the `--parallel-no-spinner` default based on `is_ci()` utility.

## Alternative Solutions

No alternatives.

## Caveats
There's no `--parallel-spinner` which would force the spinner if it was disabled by the CI detection. `TOX_PARALLEL_NO_SPINNER=0` should still work in those (hopefully rare) cases.

ziima:
Suppress spinner in parallel run in CI (not legacy though).

### DIFF
--- a/tests/session/cmd/test_parallel.py
+++ b/tests/session/cmd/test_parallel.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import sys
 from argparse import ArgumentTypeError
 from signal import SIGINT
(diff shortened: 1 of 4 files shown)


gaborbernat (approved):

tests/session/cmd/test_parallel.py
```diff
@@ -186,6 +187,19 @@ def test_parallel_no_spinner(tox_project: ToxProjectCreator) -> None:
     )
 
 
+def test_parallel_no_spinner_ci(tox_project: ToxProjectCreator) -> None:
+    """Ensure spinner is disabled by default in CI."""
+    with mock.patch.object(parallel, "execute") as mocked, mock.patch.dict(os.environ, {"CI": "1"}):
```
gaborbernat, line 192: Use pytest-mock instead

gaborbernat (changes requested):

tests/session/cmd/test_parallel.py
ziima: Fixed

ziima: OK, I had to rolled back. It seems pytest-mock has some issue with `os.environ`. Also `unittest.mock` is used to modify `os.environ` throughout tox tests.

gaborbernat: I strongly urge you to reconsider. For patching the operating system environment variables, you should use the built-in pytest monkeypatch fixture.

gaborbernat: I will not accept any pull request that doesn't use the already practiced tools and introduces a new way of doing the same thing.

ziima:
Thanks for fixing the PR, I didn't get back to that yet.

PR #3325, opened 2024-08-13
Fix absolute base python paths conflicting

gaborbernat:

### DIFF
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -168,9 +168,14 @@ class Python(ToxEnv, ABC):
         if env_base_python is not None:
             spec_name = PythonSpec.from_string_spec(env_base_python)
             for base_python in base_pythons:
-                if Path(base_python).is_absolute():
-                    return [base_python]
                 spec_base = PythonSpec.from_string_spec(base_python)
+                if spec_base.path is not None:
+                    path = Path(spec_base.path).absolute()
+                    if str(spec_base.path) == sys.executable:
+                        ver, is_64 = sys.version_info, sys.maxsize != 2**32
+                        spec_base = PythonSpec.from_string_spec(f"{sys.implementation}{ver.major}{ver.minor}-{is_64}")
+                    else:
+                        spec_base = cls.python_spec_for_path(path)
                 if any(
                     getattr(spec_base, key) != getattr(spec_name, key)
                     for key in ("implementation", "major", "minor", "micro", "architecture")
@@ -182,6 +187,17 @@ class Python(ToxEnv, ABC):
                         return [env_base_python]
                     raise Fail(msg)
         return base_pythons
+
+    @classmethod
+    @abstractmethod
+    def python_spec_for_path(cls, path: Path) -> PythonSpec:
+        """
+        Get the spec for an absolute path to a Python executable.
+
+        :param path: the path investigated
+        :return: the found spec
+        """
+        raise NotImplementedError
 
     @abstractmethod
     def env_site_package_dir(self) -> Path:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
   push:
     branches: ["main"]
-    tags-ignore: [ "**" ]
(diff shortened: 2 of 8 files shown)


hroncok:
Hello. Is it expected that this breaks extensions who implement this abstract class?

See https://github.com/fedora-python/tox-current-env/issues/77

gaborbernat:
yes, sadly I don't have a better option to add new abstract methods here

src/tox/tox_env/python/api.py
```diff
@@ -168,9 +168,14 @@ def _validate_base_python(
         if env_base_python is not None:
             spec_name = PythonSpec.from_string_spec(env_base_python)
             for base_python in base_pythons:
-                if Path(base_python).is_absolute():
-                    return [base_python]
                 spec_base = PythonSpec.from_string_spec(base_python)
+                if spec_base.path is not None:
+                    path = Path(spec_base.path).absolute()
+                    if str(spec_base.path) == sys.executable:
+                        ver, is_64 = sys.version_info, sys.maxsize != 2**32
```
hroncok, line 175: When reviewing https://github.com/fedora-python/tox-current-env/pull/78 @frenzymadness found out that `is_64` is always True:

### Fedora Rawhide i686

```pycon
>>> sys.maxsize
2147483647
>>> 2**32
4294967296
>>> sys.maxsize != 2**32
True
```

### Fedora Rawhide x86_64


```pycon
>>> sys.maxsize
9223372036854775807
>>> 2**32
4294967296
>>> sys.maxsize != 2**32
True
```

https://docs.python.org/3/library/platform.html#platform.architecture says:

> To get at the “64-bitness” of the current interpreter, it is more reliable to query the [sys.maxsize](https://docs.python.org/3/library/sys.html#sys.maxsize) attribute:
> 
> ```python
> is_64bits = sys.maxsize > 2**32
> ```

### Fedora Rawhide i686

```pycon
>>> sys.maxsize > 2**32
False
```

### Fedora Rawhide x86_64


```pycon
>>> sys.maxsize > 2**32
True
```

src/tox/tox_env/python/api.py
```diff
@@ -168,9 +168,14 @@ def _validate_base_python(
...
+                        spec_base = PythonSpec.from_string_spec(f"{sys.implementation}{ver.major}{ver.minor}-{is_64}")
```
hroncok, line 176: Moreover, @frenzymadness pointed out that the string ends with a boolean, like `...-313-True`, while according to a PATTERN in virtualenv:

https://github.com/pypa/virtualenv/blob/95c5eed96c225faff01ac2f108368e0c4303989b/src/virtualenv/discovery/py_spec.py#L8

```python
PATTERN = re.compile(r"^(?P<impl>[a-zA-Z]+)?(?P<version>[0-9.]+)?(?:-(?P<arch>32|64))?$")
```

It should probably end with `-32` or `-64` if we care about the 32/64bitness.

Moreover, when testing this assumption, I found out that the entire string is probably completely bogus:

```pycon
>>> ver, is_64 = sys.version_info, sys.maxsize != 2**32
>>> f"{sys.implementation}{ver.major}{ver.minor}-{is_64}"
"namespace(name='cpython', cache_tag='cpython-313', version=sys.version_info(major=3, minor=13, micro=0, releaselevel='candidate', serial=1), hexversion=51183809, _multiarch='x86_64-linux-gnu')313-True"
```

We probably want to replace `sys.implementation` with `sys.implementation.name` if `cpython313-64` is what you intended.

src/tox/tox_env/python/api.py
gaborbernat: Ah good catch both. can you put in a PR to fix it 😅

hroncok: I am working on it. I assume you intended a string like `cpython3.13-64`.

gaborbernat: That is correct, thank you.

hroncok: Drafted https://github.com/tox-dev/tox/pull/3327 -- waiting for the CI to validate it.

PR #3327, opened 2024-08-15
Fixup the spec string for sys.executable

hroncok:
The previous spec string was:

    namespace(name='cpython', cache_tag='cpython-313', version=sys.version_info(major=3, minor=13, micro=0, releaselevel='candidate', serial=1), hexversion=51183809, _multiarch='x86_64-linux-gnu')313-True

When it was supposed to be:

    cpython313-64

Fixes https://github.com/tox-dev/tox/pull/3325/files#r1718230254
Fixes https://github.com/tox-dev/tox/pull/3325/files#r1718246292

Adds tests for a new method.

### DIFF
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -158,6 +158,14 @@ class Python(ToxEnv, ABC):
         return None
 
     @classmethod
+    def python_spec_for_sys_executable(cls) -> PythonSpec:
+        implementation = sys.implementation.name
+        version = sys.version_info
+        bits = "64" if sys.maxsize > 2**32 else "32"
+        string_spec = f"{implementation}{version.major}{version.minor}-{bits}"
+        return PythonSpec.from_string_spec(string_spec)
+
+    @classmethod
     def _validate_base_python(
         cls,
         env_name: str,
@@ -172,8 +180,7 @@ class Python(ToxEnv, ABC):
                 if spec_base.path is not None:
                     path = Path(spec_base.path).absolute()
                     if str(spec_base.path) == sys.executable:
-                        ver, is_64 = sys.version_info, sys.maxsize != 2**32
-                        spec_base = PythonSpec.from_string_spec(f"{sys.implementation}{ver.major}{ver.minor}-{is_64}")
+                        spec_base = cls.python_spec_for_sys_executable()
                     else:
                         spec_base = cls.python_spec_for_path(path)
                 if any(
--- a/src/tox/tox_env/python/api.py
+++ b/src/tox/tox_env/python/api.py
@@ -158,7 +158,7 @@ class Python(ToxEnv, ABC):
         return None
 
     @classmethod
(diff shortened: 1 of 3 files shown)


src/tox/tox_env/python/api.py
```diff
@@ -157,6 +157,14 @@ def extract_base_python(cls, env_name: str) -> str | None:
             return next(iter(candidates))
         return None
 
+    @classmethod
+    def python_spec_for_sys_executable(cls) -> PythonSpec:
+        implementation = sys.implementation.name
+        version = sys.version_info
+        bits = "64" if sys.maxsize > 2**32 else "32"
```
gaborbernat, line 164: bits can be int here

src/tox/tox_env/python/api.py
```diff
@@ -157,6 +157,14 @@ def extract_base_python(cls, env_name: str) -> str | None:
             return next(iter(candidates))
         return None
 
+    @classmethod
```
gaborbernat, line 160: make it private please

src/tox/tox_env/python/api.py
hroncok: Yes, they can but they will be inserted to a string anyway, so I thought why go int to str, when I can use a str literal.

src/tox/tox_env/python/api.py
hroncok: I will definitively take your lead here, but I wanted to use this in tox-current-env.

hroncok: Making it private ruff-screams at me in tests:

```
tests/tox_env/python/test_python_api.py:314:12: SLF001 Private member accessed: `_python_spec_for_sys_executable`
    |
312 |     mocker.patch.object(sys, "implementation", implementation)
313 |     mocker.patch.object(sys, "maxsize", 2**arch // 2 - 1)
314 |     spec = Python._python_spec_for_sys_executable()
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ SLF001
315 |     assert spec.implementation == impl
316 |     assert spec.major == major
    |

Found 1 error.
```

gaborbernat: Just silence it. 😅

gaborbernat (commented):
You will need to add change log entry as the bug fix but otherwise it looks good to go.

src/tox/tox_env/python/api.py
hroncok: I am unfamiliar with your linter setup (it feels like an obstacle rather than something that would help me as a drive-by contributor). Could you please tell me what exactly is necessary or better yet push a commit here that does that? I could go and learn how to configure ruff to exclude tests from "Private member accessed" checked, but I assume you know this better than me.

gaborbernat: # noqa: SLF001

gaborbernat (approved):

PR #3342, opened 2024-09-10
Ensure tox-created directories contain CACHEDIR.TAG

### LINKED ISSUE #3334: `.tox` folder should maybe contain a `CACHEDIR.TAG`

Since `.tox/` is an ephemeral folder, it could be useful if it contained a [`CACHEDIR.TAG` file](https://bford.info/cachedir/).

These files are useful markers that [e.g. GNU `tar`](https://www.gnu.org/software/tar/manual/html_node/exclude.html#index-exclude_002dcaches) supports to exclude archiving/backing up/... recreatable data.

When Tox creates `.tox`, it adds a spec-compliant `CACHEDIR.TAG` file there.

akx:
This PR ensures that Tox-created work directories contain a `CACHEDIR.TAG` file, which can be used to exclude the directories from e.g. backups.

Fixes #3334

### DIFF
--- a/docs/changelog/3342.feature.rst
+++ b/docs/changelog/3342.feature.rst
@@ -1,0 +1,7 @@
+Tox now creates ``CACHEDIR.TAG`` files in work directories it creates,
+so that tools like ``tar`` can exclude them from e.g. backups where
+ephemeral directories are not desired.
+
+Tag files are not created in directories Tox does not itself create.
+
+- by :user:`akx`
--- a/src/tox/tox_env/api.py
+++ b/src/tox/tox_env/api.py
@@ -200,6 +200,11 @@ class ToxEnv(ABC):  # noqa: PLR0904
         return cast(Path, self.core["work_dir"])
 
     @property
+    def temp_dir(self) -> Path:
+        """:return: the tox work dir folder"""
+        return cast(Path, self.core["temp_dir"])
+
+    @property
     def name(self) -> str:
         return cast(str, self.conf["env_name"])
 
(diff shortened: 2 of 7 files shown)


gaborbernat (changes requested):
Why create these in sub folders? Why not just created it within the `.tox` root directory and be done with it?

akx:
> Why not just created it within the `.tox` root directory and be done with it?

I felt there was an idea behind the decision to not have an explicit call to create that directory – maybe that the workdir wouldn't be eagerly created before it's needed (just to hold a subdirectory).

> Just because there isn't does not mean we can't add it.

Sure, I'll do that. Would you prefer eagerly creating the workdir with the cache tag, or having the places that create those workdir subdirs ensure the workdir has the cache tag?

gaborbernat:
> having the places that create those workdir subdirs ensure the workdir has the cache tag

Probably this 🤔

akx:
> > having the places that create those workdir subdirs ensure the workdir has the cache tag
> 
> Probably this 🤔

It turned out a little trickier like that, since an user could point Tox to a pre-created workdir, which we then certainly do not want to litter with a `CACHEDIR.TAG` file.

I still didn't write tests for this incarnation, just in case you'd like to take a review look first.

gaborbernat (changes requested):
The CI is fialing

gaborbernat (changes requested):
Needs tests too 👍

gaborbernat:
@akx you still plan to finish this or should I close it?

akx:
> @akx you still plan to finish this or should I close it?

Yeah, it's still under construction – just been busy with work and figuring out how to best test-cover this. Rebased & added a commit that has some WIP tests :)

gaborbernat:
No worries, but note the CI is still failing 😊

gaborbernat:
@akx friendly bump here 👍

PR #3344, opened 2024-09-15
Prepare Loader and MemoryLoader for TOML support

ziima:
- [ ] ~~updated/extended the documentation~~

### DIFF
--- a/src/tox/config/loader/memory.py
+++ b/src/tox/config/loader/memory.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Iterator
+from typing import TYPE_CHECKING, Any, Iterator, Sequence
 
 from tox.config.types import Command, EnvList
 
@@ -62,4 +62,6 @@ class MemoryLoader(Loader[Any]):
             return value
         if isinstance(value, str):
             return StrConvert.to_env_list(value)
+        if isinstance(value, Sequence):
+            return EnvList(value)
         raise TypeError(value)
--- a/src/tox/config/loader/memory.py
+++ b/src/tox/config/loader/memory.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, Any, Iterator, Sequence
 
 from tox.config.types import Command, EnvList
 
-from .api import Loader
+from .api import Loader, Override
 from .section import Section
 from .str_convert import StrConvert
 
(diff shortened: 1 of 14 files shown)


src/tox/config/loader/memory.py
```diff
@@ -14,9 +14,16 @@
 
 
 class MemoryLoader(Loader[Any]):
-    def __init__(self, **kwargs: Any) -> None:
-        super().__init__(Section(prefix="<memory>", name=str(id(self))), [])
-        self.raw: dict[str, Any] = {**kwargs}
+    def __init__(
```
gaborbernat, line 17: This is a backwards incompatible change and cannot happen.

gaborbernat (changes requested):

src/tox/config/loader/memory.py
ziima: OK, we could have cleared this in #3309. Anyway, he are reasons why I did it:
* `MemoryLoader` is not listed in API docs: https://tox.wiki/en/latest/plugins_api.html Hence, I assumed it may have incompatible changes.
* `MemoryLoader.__init__` isn't currently compatible with the documented API for `Loader.__init__`.

I can add some compatible layer and/or deprecation warnings, but I need to know how to handle it correctly. More incompatible changes will be needed for TOML support.

ziima: The compatibility makes the child class more flexible, especially when some of the omitted arguments become needed such as in this case.

Now I'm not quite sure which way use to make it compatible. It seems to me it would be best to create new `MemoryLoader` class with modified constructor and deprecate current one. Is that OK?

gaborbernat: I disagree. Why does a memory loader need to know about its section? Or its overrides. These are concepts that do not make sense in this context, so I am not sure why you are trying to force it into?

ssbarnea: I hope you reach an agreement here. I see the reasoning for having to access these bits, we use it in tox-ansible plugin. At the same time, I would try to avoid breaking changes (changing signature does not always count as breaking in my eyes).

gaborbernat: I feel like overrides should be their own loader 🤔 and not depend on the config source. 🤔

gaborbernat: Ok, I looked into this, overrides are intended to be not enabled for a memory loader. And I guess what I am not understanding is why you want to change that?

gaborbernat: Here is an implementation demonstrating there is no need for this https://github.com/tox-dev/tox/pull/3353, will continue over the next week to get it in a ready to ship state 👍

gaborbernat:
Replaced by https://github.com/tox-dev/tox/pull/3353

PR #3404, opened 2024-10-14
Expose type checking dependencies into an extra

ssbarnea:
This will allow to add tox to [mypy_primer](https://github.com/hauntsaninja/mypy_primer) so new versions of mypy will be tested against it (used directly from their [GHA pipelines](https://github.com/python/mypy/blob/master/.github/workflows/mypy_primer.yml)). It should also make it easier to do type checking for plugin developers.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,11 @@ optional-dependencies.testing = [
   "time-machine>=2.15; implementation_name!='pypy'",
   "wheel>=0.44",
 ]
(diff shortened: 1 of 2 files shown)


gaborbernat:
`mypre_primer` ?

ssbarnea:
> `mypre_primer` ?

Fixed the typo and included links to both primer project and also how it is used by mypy itself. I will be adding a PR to the project to add tox to it.

PR #3410, opened 2024-10-16
Add NETRC to the default_pass_env list

chipot:
Add NETRC to the list of default_pass_env.
This is used by pip when downloading package from an index needing authentication.

My usecase is a CI process needing to obtain authentication tokens and pass them to pip for it to be able to use the index. Installing the credentials in `~/.netrc` would work but isn't always possible when using temporary tokens, or running CI jobs in parallel on the same host. The other supported way is to write a temporary `netrc` file, and point `pip` to it using `NETRC` environment variable. After the CI run is done, this `netrc` can be removed.

- [] ran the linter to address style issues (`tox -e fix`)

### DIFF
--- a/docs/changelog/3410.feature.rst
+++ b/docs/changelog/3410.feature.rst
@@ -1,0 +1,1 @@
+Add `NETRC` to the list of environment variables always passed through.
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
CI failing.

gaborbernat (commented):
Changelog please.

gaborbernat (approved):

PR #3426, opened 2024-10-28
Adds ability to configure stderr output color

ssbarnea:
This feature will allow users to control the coloring of stderr with tox, including ability to avoiding adding colors to it, while still using colors as many tools are already producing colored output.

[development documentation](http://tox.readthedocs.org/en/latest/development.html#development))! -->

### DIFF
--- a/src/tox/config/cli/parser.py
+++ b/src/tox/config/cli/parser.py
@@ -11,6 +11,8 @@ from argparse import SUPPRESS, Action, ArgumentDefaultsHelpFormatter, ArgumentEr
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple, Type, TypeVar, cast
 
+from colorama import Fore
+
 from tox.config.loader.str_convert import StrConvert
 from tox.plugin import NAME
 from tox.util.ci import is_ci
@@ -365,6 +367,12 @@ def add_color_flags(parser: ArgumentParser) -> None:
         default=color,
         choices=["yes", "no"],
         help="should output be enriched with colors, default is yes unless TERM=dumb or NO_COLOR is defined.",
+    )
+    parser.add_argument(
+        "--stderr-color",
+        default="RED",
+        choices=[*Fore.__dict__.keys()],
+        help="color for stderr output, use RESET for terminal defaults.",
     )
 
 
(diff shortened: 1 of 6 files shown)


src/tox/config/cli/parser.py
```diff
@@ -366,6 +368,12 @@ def add_color_flags(parser: ArgumentParser) -> None:
         choices=["yes", "no"],
         help="should output be enriched with colors, default is yes unless TERM=dumb or NO_COLOR is defined.",
     )
+    parser.add_argument(
+        "--stderr-color",
+        default="RED",
+        choices=["0", *Fore.__dict__.keys()],
+        help="color for stderr output, use 0 to disable coloring.",
```
gaborbernat, line 375: We already have NO_COLOR env-var users should use if they don't want color, let's not add 0.

ssbarnea: Here is the deal: I want color, I just don't want any specific color! Clearly using NO_COLOR is not suitable here.

Any other ideas? -- I an easily look for other string values like "NONE" instead of "0".

ssbarnea: Sorted.

ssbarnea:
@gaborbernat Can you have another look please? Thanks.

tests/execute/local_subprocess/test_local_subprocess.py
```diff
@@ -77,7 +87,7 @@ def test_local_execute_basic_pass(  # noqa: PLR0913
     out_got, err_got = out_err.read_out_err()
     if show:
         assert out_got == out
-        expected = (f"{Fore.RED}{err}{Fore.RESET}" if color else err) if err else ""
+        expected = f"{Fore.__dict__[stderr_color]}{err}{Fore.RESET}" if color and err else err
```
gaborbernat, line 90: Use getattr instead of this dictionary lookup.

gaborbernat (changes requested):
The build is failing so there's nothing to review yet that much

tests/execute/local_subprocess/test_local_subprocess.py
ssbarnea: Indeed, much nicer than the __dict__ approach.

gaborbernat, line 90: Getattr

gaborbernat (changes requested):

gaborbernat (approved):

PR #3446, opened 2024-11-15
feat: adding a schema command

henryiii:
Started on a `schema` command that produces a JSONSchema.


henryiii:
Adding a commit that would support plugins setup like tox-gh. To actually fix that one, though, it needs to be modified a bit something like this:

```python
    bail_reason = None
    if state.conf.options.command == "schema":
        pass
    elif not core_conf["is_on_gh_action"]:
        bail_reason = "tox is not running in GitHub Actions"
    elif getattr(getattr(state.conf.options, "env", None), "is_default_list", False) is False:
        bail_reason = f"envlist is explicitly given via {'TOXENV' if os.environ.get('TOXENV') else '-e flag'}"
    else:
        logging.warning("running tox-gh")
```

henryiii:
I've been using `uvx check-jsonschema --schemafile src/tox/tox.schema.json tox.toml` to check it locally so I just need to look up a Python API for it. The biggest question, though; does it make sense to keep a (continually) rendered version in the repo? Or should it only be generated?

gaborbernat (approved):

henryiii:
> does it make sense to keep a (continually) rendered version in the repo? Or should it only be generated?

Should I add a job to rerender the schema in the repo?

gaborbernat:
That would be great 👍

PR #3461, opened 2024-12-10
Use ubuntu-24.04 during build

ssbarnea:
This change will ensure that we are in control of the ubuntu version
using build and avoid surprises from GitHub as the date when they
make the switch is not predictable.

This needs to change only once every two years, so not really a
problem. Also sorts the noise we currently get due to the upcoming
switch:

ubuntu-latest pipelines will use ubuntu-24.04 soon. For more details, see https://github.com/actions/runner-images/issues/10636

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -27,7 +27,7 @@ jobs:
           - "3.9"
           - "3.8"
         os:
-          - ubuntu-latest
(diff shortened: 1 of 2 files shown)


gaborbernat:
I don't like this. I want to receive notifications when things change rather than silently keep working.

PR #3466, opened 2025-01-16
💅 Make SVG image compatible with Firefox

### LINKED ISSUE #3465: SVG logo in the docs does not render in Firefox
## Issue

$sbj. It works in Chrome, but in FF (Gentoo Linux, but I don't think this is related — it's just as broken in Firefox on Android) it renders blank. Other SVGs work for me. Like I can, see https://github.com/aio-libs/aiohttp/blob/8844385/docs/aiohttp-icon.svg. But https://github.com/tox-dev/tox/blob/e3e77a6/docs/_static/img/tox.svg is just blank everywhere — on RTD, in the GH preview and when accessing by the direct raw URL (https://tox.wiki/en/latest/_static/tox.svg). The browser gets the file and DevTools show that the nodes are being parsed.

There must be something wrong with that XML in SVG that is incompatible with FF. This needs to be investigated.

webknjaz:
It appears that Firefox has bugs around interpreting `<feBlend>` filters in SVG documents. This makes it render the logo as a transparent image in the docs, both in the mobile and the desktop versions.

This patch applies a workaround [[1]] found on the internet to make it work without waiting for Firefox to fix their bug.

Resolves #3465.

[1]: https://github.com/svg/svgo/issues/732

### DIFF
--- a/docs/_static/img/tox.svg
+++ b/docs/_static/img/tox.svg
@@ -14,6 +14,11 @@
   style="enable-background:new"
 >
   <defs id="defs2">
+    <style>
+    .backgroundImageFilterWithLuminocityMode {
+      filter: url(.#backgroundImageFilterWithLuminocityModeBlend);
+    }
+    </style>
     <linearGradient id="linearGradient6236">
       <stop id="stop6232" offset="0" style="stop-color:#ffff00;stop-opacity:1;" />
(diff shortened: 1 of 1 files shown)


webknjaz:
@gaborbernat the failure is unrelated (`setup-uv`'s broken?). Plz also apply the change note skip label since I don't think it's needed.

gaborbernat (approved):

Issue #3474, opened 2025-01-29
TOML configuration of `set_env` should also support loading from environment files


brianhelba:
## What's the problem this feature will solve?

Loading environment files in `set_env` was implemented by @gaborbernat in #1668 and is currently documented as being generally usable. However, it appears that the `file|` syntax and custom parsing code is incompatible with a TOML configuration of `set_env`.

## Describe the solution you'd like

TOML configuration of `set_env` should support the ability to load environment files. This would probably be most elegant with some new syntax, as the `file|` syntax doesn't map cleanly onto the expected type of `set_env` in TOML.

## Alternative Solutions

Alternatively, the documentation should be updated to clarify that the environment file loading / `file|` syntax is only supported in the legacy INI configuration file format.

gaborbernat:
PR welcome 👍

juditnovak:
I made an attempt to address the issue in https://github.com/tox-dev/tox/pull/3478 . Please let me know what you think.

PR #3478, opened 2025-02-05
TOML set_env file support

juditnovak:
This code change is addressing https://github.com/tox-dev/tox/issues/3474, making sure that TOML config has the same support to apply `set_env` from file, as INI configuration does.

Resolves https://github.com/tox-dev/tox/issues/3474

### DIFF
--- a/src/tox/config/loader/toml/__init__.py
+++ b/src/tox/config/loader/toml/__init__.py
@@ -58,31 +58,18 @@ class TomlLoader(Loader[TomlTypes]):
 
     def build(  # noqa: PLR0913
         self,
-        key: str,
+        key: str,  # noqa: ARG002
         of_type: type[_T],
         factory: Factory[_T],
         conf: Config | None,
         raw: TomlTypes,
         args: ConfigLoadArgs,
     ) -> _T:
-        delay_replace = inspect.isclass(of_type) and issubclass(of_type, SetEnv)
-
-        def replacer(raw_: str, args_: ConfigLoadArgs) -> str:
-            reference_replacer = Unroll(conf, self, args)
-            try:
-                replaced = str(reference_replacer(raw_))  # do replacements
-            except Exception as exception:
-                if isinstance(exception, HandledError):
-                    raise
-                msg = f"replace failed in {args_.env_name}.{key} with {exception!r}"
-                raise HandledError(msg) from exception
-            return replaced
-
         exploded = Unroll(conf=conf, loader=self, args=args)(raw)
-        refactoried = self.to(exploded, of_type, factory)
-        if delay_replace:
-            refactoried.use_replacer(replacer, args=args)  # type: ignore[attr-defined] # issubclass(to_type, SetEnv)
-        return refactoried
+        result = self.to(exploded, of_type, factory)
+        if inspect.isclass(of_type) and issubclass(of_type, SetEnv):
+            result.use_replacer(lambda c, s: c, args=args)  # type: ignore[attr-defined] # noqa: ARG005
+        return result
 
     def found_keys(self) -> set[str]:
         return set(self.content.keys()) - self._unused_exclude
@@ -126,5 +113,6 @@ class TomlLoader(Loader[TomlTypes]):
 
 
 __all__ = [
+    "HandledError",
     "TomlLoader",
 ]
--- a/src/tox/config/loader/toml/__init__.py
+++ b/src/tox/config/loader/toml/__init__.py
@@ -1,11 +1,14 @@
 from __future__ import annotations
 
+import inspect
 import logging
 from pathlib import Path
 from typing import TYPE_CHECKING, Dict, Iterator, List, Mapping, TypeVar, cast
 
 from tox.config.loader.api import ConfigLoadArgs, Loader, Override
+from tox.config.set_env import SetEnv
 from tox.config.types import Command, EnvList
+from tox.report import HandledError
 
 from ._api import TomlTypes
 from ._replace import Unroll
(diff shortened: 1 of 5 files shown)


src/tox/config/loader/toml/__init__.py
```diff
@@ -55,15 +58,34 @@ def load_raw_from_root(self, path: str) -> TomlTypes:
...
         factory: Factory[_T],
         conf: Config | None,
         raw: TomlTypes,
         args: ConfigLoadArgs,
     ) -> _T:
+        delay_replace = inspect.isclass(of_type) and issubclass(of_type, SetEnv)
+
+        def replacer(raw_: str, args_: ConfigLoadArgs) -> str:
+            if conf is None:
+                replaced = raw_  # no replacement supported in the core section
```
juditnovak, line 72: TBH this condition I've just taken from the INI correspondant... Pls double-check if it makes sense in TOML parsing context. Thanks.

tests/config/test_set_env.py
```diff
@@ -149,7 +161,19 @@ def test_set_env_honor_override(eval_set_env: EvalSetEnv) -> None:
     assert set_env.load("PIP_DISABLE_PIP_VERSION_CHECK") == "0"
 
 
-def test_set_env_environment_file(eval_set_env: EvalSetEnv) -> None:
+@pytest.mark.parametrize(
+    ("conf_type", "config"),
+    [
+        ("ini", "[testenv]\npackage=skip\nset_env=file|A{/}a.txt\nchange_dir=C"),
+        ("toml", '[env_run_base]\npackage="skip"\nset_env="file|A{/}a.txt"\nchange_dir="C"'),
```
gaborbernat, line 168: It would make more sense that this to be a dictionary key rather than we just unroll a string. Something like 'set_env = [ { file= "a.txt"}] `.

gaborbernat (changes requested):

tests/config/test_set_env.py
juditnovak: Just to make sure that I'm following correctly. Could you pls confirm that the following is what you suggest?

### INI:
The current syntax supported for `tox.ini` for `set_env`:
```
[testenv]
set_env=file|<path>
```

### TOML:
Moving away from the `file|` syntax introduced for `tox.ini`, benefitting from embedded data structures instead such as:
```
[env_run_base]
set_env=[{file=<path>}]
```
While this would raise an ERROR in TOML:
```
[env_run_base]
set_env="file|<path>"
```

Is this the direction you would like to take? Not to support `file|` terminology when TOML configuration is used?

(Sorry, I understood that a PR was welcome for the issue as raised -- which my attempt was aiming to respond. Of course, I'm glad to modify the code alongside a different design, if preferred. I agree that the above solution is more elegant.)

gaborbernat: Yeah, with the TOML format we generally are trying to move away from embedding information into strings, and instead use the rich type system and structure of the TOML language instead.

gaborbernat:
You still plan to tackle this?

juditnovak:
@gaborbernat Yes I do. Sorry for the silence, I was unavailable for a couple of days. Updated PR should be available early next week.

gaborbernat:
@juditnovak tag me when ready 👍

juditnovak:
@gaborbernat Sorry for the delay, the PR is ready. I keep thinking that I'm kind unhappy with the solution in terms of separation of concerns -- as part of the INI/TOML processing is happening on the level of each loader (`__init__.py`), while format-specific logic now is also hitting the shared (ideally generic) `set_env` module...
(Yet, otherwise `set_env` logic would need to "cascade up" to each loader's level :-/ )

In case I'm overlooking a different, more elegant solution, feel free to let me know, and I'm glad to adjust.

src/tox/config/loader/toml/__init__.py
gaborbernat: This does not make sense here. Please revert changes to this.

gaborbernat (changes requested):

juditnovak:
@gaborbernat The condition in question is removed (assuming that 'revert' meant removal of newly added code).

src/tox/config/loader/toml/__init__.py
```diff
@@ -55,15 +58,31 @@ def load_raw_from_root(self, path: str) -> TomlTypes:
...
+                msg = f"replace failed in {args_.env_name}.{key} with {exception!r}"
+                raise HandledError(msg) from exception
+            return replaced
+
         exploded = Unroll(conf=conf, loader=self, args=args)(raw)
-        return self.to(exploded, of_type, factory)
+        refactoried = self.to(exploded, of_type, factory)
+        if delay_replace:
+            refactoried.use_replacer(replacer, args=args)  # type: ignore[attr-defined] # issubclass(to_type, SetEnv)
+        return refactoried
```
gaborbernat, line 85: I mean this file should not be changed at all 🤔

juditnovak: Could you please elaborate on your preferred strategy, please?

I was following the INI logic keeping INI and TOML processing side-by-side. If a different logic is preferred, may I ask what you have in mind?

gaborbernat: Why do we need this processing for TOML? This is an ini specific stuff 🤔

juditnovak: No, it's not. The `replacer` function is used within the `set_env` module, to determine the correct path, that may have substitutions inside.

Note that we are using the TOML-specific `Unroll()` call within the code block in question.

This is the reason why TOML-specific tests with `file="{env:myenvfile}`work.

juditnovak: I've added an extension to the documentation within this same PR, highlighting the feature ensured by the code above (commit 7a0971db54e9e02d76ff98b96f62793aac722483).

Please let me know:
 a. if further explanation may be helpful from my side
 b. if this feature is to be preserved for INI files (as in current versions of Tox) but not to be supported for TOML (according to your request above)
 
 In case there may be a potential mis-understanding on my side, please also let me know.

juditnovak: @gaborbernat :arrow_up: 
(No intent to pressurize, but rather to trigger a notification, in case there may have been none before :-) )

gaborbernat:
Yeah, didn't forget about this, have the tab pinned. Just this week has been very busy at work and didn't get time to allocate to this, we should get this out by early next week though 👍

juditnovak:
@gaborbernat Sure, thanks much for getting back. 
No pressure, I just wanted to make sure to keep you in the loop :-)

gaborbernat (approved):

gaborbernat:
Fix released with https://github.com/tox-dev/tox/releases/tag/4.24.2

PR #3528, opened 2025-05-11
Fix using deprecated virtualenv option `--wheel`

robsdedude:
The CI started failing when virtualenv deprecated the `--wheel` and `--no-wheel` options (in [20.31.0](https://virtualenv.pypa.io/en/latest/changelog.html#features-20-31-0)). This PR makes it so that tox doesn't run into this deprecation which caused the tests to fail that checked for some expected stdout which were then unexpectedly containing the deprecation message.

       (Existing test were failing, so if this improves the situation, I consider this box ticked :wink:) 
- [ ] ~~added news fragment in `docs/changelog` folder~~ not user-facing, or it it?
- [ ] ~~updated/extended the documentation~~ API/CLI didn't change

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ dependencies = [
   "pyproject-api>=1.8",
   "tomli>=2.2.1; python_version<'3.11'",
   "typing-extensions>=4.12.2; python_version<'3.11'",
-  "virtualenv>=20.29.1",
+  "virtualenv>=20.31",
 ]
 optional-dependencies.test = [
   "devpi-process>=1.0.2",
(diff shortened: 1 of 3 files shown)


robsdedude:
This turns out to be more involved than I anticipated it to be :innocent: 

I think to be able to get rid of using virtualenv to install wheel, `pyproject-api` needs to be update to not require wheel. Which would be in line with the [pyproject.toml specs](https://packaging.python.org/en/latest/specifications/pyproject-toml/#declaring-build-system-dependencies-the-build-system-table): 

> ```toml
> [build-system]
> # Minimum requirements for the build system to execute.
> requires = ["setuptools"]
> ```
> 
> Build tools are expected to use the example configuration file above as their default semantics when a pyproject.toml file is not present.

I think it might be this line that's causing issues: https://github.com/tox-dev/pyproject-api/blob/46dc443b1a33599fb76f869bd911fd4827eafb0d/src/pyproject_api/_frontend.py#L179

gaborbernat:
https://github.com/tox-dev/pyproject-api/releases/tag/1.9.1

hroncok:
With pyproject-api 1.9.1, this works and fixes the issue for me.

gaborbernat (dismissed):

hroncok (approved):

PR #3529, opened 2025-05-11
Add security contact information

jugmac00:
Tidelift will coordinate the vulnerability and the fix with the tox team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,3 +23,8 @@ tox is a generic virtual environment management and test command line tool you c
 Please read our [user guide](https://tox.wiki/en/latest/user_guide.html#basic-example) for an example and more detailed
 introduction, or watch [this YouTube video](https://www.youtube.com/watch?v=SFqna5ilqig) that presents the problem space
 and how tox solves it.
+
+## Security contact information
+
+To report a security vulnerability, please use the [Tidelift security contact](https://tidelift.com/security). Tidelift
+will coordinate the fix and disclosure.


jugmac00:
@gaborbernat As quickly discussed on discord, Tidelift reached out last week that we need to take care of some tasks, one of them was to add a way to report security issues with Tidelift. As soon this has been merged, I can finalize the last task, where I need to set the link to the file with the security notice.

webknjaz:
@jugmac00 make sure to add a `SECURITY.md` that would show up @ https://github.com/tox-dev/tox/security (it's one item missing from the checklist: https://github.com/tox-dev/tox/community).

Though, personally, I find it more convenient to enable submitting reports through the GH UI. There's a toggle @ https://github.com/tox-dev/tox/settings/security_analysis labeled **`Private vulnerability reporting`** where you can enable this. With that, people can suggest vulnerabilities, which can then be triaged and accepted/rejected, turned into draft, worked on. It also lets you create a private fork where both the reporter and the maintainers can collaborate through pull requests (that don't trigger CI or other automations, obviously) and finally publish, request a CVE, credit various contributing parties, invite more people to collaborate, etc.

So I'd recommend considering doing this (additionally or instead). I think it's possible to have the Tidelift process route incoming reports that way, too.

jugmac00:
@webknjaz thanks for your suggestions. Tide lift mentioned we could either put the info in the readme or in a standalone file. To get started, I went with the former approach. Are you aware of a good example for a security.md which we could use as an inspiration?

webknjaz:
It's important to have that file, since GH links it from various places in the UI. Including https://github.com/tox-dev/tox/issues/new/choose, which currently doesn't have a hint about security issues but should start displaying it once you add that file (example: https://github.com/aio-libs/yarl/issues/new/choose). Additionally to that automatically linked option, I sometimes an explicit link there too: https://github.com/aio-libs/yarl/blob/60f99a1/.github/ISSUE_TEMPLATE/config.yml#L8-L18.


I think you can point to another location from there if you want to host this information in another doc. Like this: https://github.com/ansible/pylibssh/security/policy.

Though, personally, I like including a more verbose guide: https://github.com/aio-libs/aiohttp/security/policy.

If you start by clicking the `Propose` button @ https://github.com/tox-dev/tox/community, it'll also have some initial template. You can place the file itself not only in the repo root, but alternatively under `.github/` or `docs/`. I like using the `docs/` variant since it's easy to then include it into the Sphinx site.

webknjaz:
I've also started linking the vulnerability form from the issue templates to target people trying to post a public issue before they get a chance to do it: https://github.com/aio-libs/yarl/issues/new?template=bug_report.yml. Here's what the form looks like, when you enable it: https://github.com/aio-libs/yarl/security/advisories/new (it's not configurable, same for everybody).

webknjaz:
@gaborbernat did you close the PR by accident? I expect it could be updated to use a different file.

gaborbernat:
Given that we are changing entirely where and what we had, I thought we might as well just open a new PR.

jugmac00:
Thanks a lot, @webknjaz - your input is highly appreciated.

PR #3530, opened 2025-05-12
Prevent Tox from hanging with `--installpkg` sdist due to orphaned build backend

### LINKED ISSUE #3512: Tox hangs on waiting for pyproject_api backend if `--installpkg` is used
## Issue

Tox hangs when using certain PEP 517 backends such as `pdm-backend` together with the `--installpkg` option.
It is the option that, for instance, [`devpi-client`](https://pypi.org/project/devpi-client/) opts to use for the `devpi test` command.

Interesting that it only happens if `tox` is run with at **least two environments**. Only one environment does not trigger the issue.

## Environment

CPython 3.12 on Ubuntu/Debian Linux.
The issue is reproducible with the official `python:3.12` Docker image (see below).

## Output of running tox

The output is indistinguishable from the normal, expected output unless the `--exit-and-dump-after` option is used.

## Minimal example

Consider the following self-contained example, let's assume it is saved under `/tmp/test.sh`:
```sh filename="/tmp/test.sh"
#!/bin/bash

set -e
set -x

python -m venv /tmp/venv
source /tmp/venv/bin/activate
pip install -qU build pip tox

mkdir -p /tmp/skeleton/skeleton
cat << EOF > /tmp/skeleton/pyproject.toml
[build-system]
requires = ["pdm-backend"]
build-backend = "pdm.backend"

[project]
name = "skeleton"
description = "Just a skeleton for reproducing a tox issue."
version = "0.1.1337"
dependencies = [
    "requests",
]

[tool.pdm.build]
includes = [
    "skeleton/",
]
source-includes = [
    "tests/",
    "tox.ini",
]
EOF
cat << EOF > /tmp/skeleton/tox.ini
[tox]
envlist = dummy1,dummy2

[testenv:dummy1]
commands =
    python -c print(1)

[testenv:dummy2]
commands =
    python -c print(42)
EOF

pushd /tmp/skeleton
python -m build --sdist
popd
tox --installpkg /tmp/skeleton/dist/skeleton-0.1.1337.tar.gz -c /tmp/skeleton/tox.ini
```

Now, run this `/tmp/test.sh` inside the `python:3.12` Docker image:
```
docker run --rm -it -v /tmp/test.sh:/tmp/test.sh python:3.12 /bin/bash /tmp/test.sh
```

vytas7:
See the discussion on the issue (#3512) for the detailed explanation of the problem.

Fixes #3512.


gaborbernat (changes requested):
We'll need a test and a changelog here 👍

vytas7:
Thanks for taking a look at this @gaborbernat!

The question remains though what we should do with #3512? Should we leave it open?

What would the correct fix be for the case of the builder's root actually being reconfigured to another path, and in which scenario can this happen?

vytas7:
OK, I'll link the newsfragment to this PR, leaving the issue _as-is_ for now.

tests/tox_env/python/virtual_env/package/test_package_pyproject.py
```diff
@@ -458,3 +458,27 @@ def test_pyproject_config_settings_editable_legacy(
...
+    commands =
+        python -c print(1)
+
+    [testenv:dummy2]
+    commands =
+        python -c print(42)
+    """
+    sdist = pkg_with_pdm_backend / "dist" / "skeleton-0.1.1337.tar.gz"
+    proj = tox_project({"tox.ini": tox_ini}, base=pkg_with_pdm_backend)
+    result = proj.run("--installpkg", str(sdist), "--exit-and-dump-after", "10")
```
vytas7, line 483: This `--exit-and-dump-after` should be increased to something larger (like 120) once I iron out the tests.

vytas7 (commented):
@gaborbernat I need help on the tests.
I tried to transplant my MRE from the issue into the test suite, and it seems to work just fine, however, if I deliberately reintroduce the bug, it still passes. :thinking:

I verified that a new backend executor is created in that case, however, that doesn't seem to cause any issues, unlike in the real run.

gaborbernat:
Seems this stalled 🤔

vytas7:
Hi @gaborbernat!
it has not stalled _per se_, I'm happy to help, but I'm not sure how to write a meaningful test.

The patch itself works for us, and we use it internally at work (hosting a custom post-release of `tox` on our internal Devpi).

There is my reproducer on #3512 that I tried to transplant into the test suite, but for some reason I cannot reproduce it therein, only when running Tox "normally".

Maybe I could try to take a look at writing a non-functional, unit-test like, test, just mocking Python attributes?

gaborbernat:
Can you fix the CI?

vytas7:
> Can you fix the CI?

I don't remember what was wrong there, I can take a fresh look. But IIRC my attempt at running the said "functional" test passed but caused the CI gates to crash or hang for some reason that I didn't understand either.

As said, I can also try to write more a unit-test like case unless you could point me to an example how to write a "functional" test running a real process. I saw that some tests did that but I don't understand the difference or the proper way to implement it.

gaborbernat:
We'd need to get the CI passing as is before we can think of adding more tests 🤔

vytas7:
@gaborbernat to sum up, I _believe_ my fix does what it should, and my fixture `pkg_with_pdm_backend` does recreate the example. What is the canonical way to run that with `tox` as a "functional" test in the suite?

gaborbernat:
We don't have functional tests, just integration tests, there are examples of that in the suite 🤔

vytas7:
Aha, thanks, I'll take a look. Maybe my attempt to `--exit-and-dump-after` as a safeguard eventually kills the suite itself? :thinking:

vytas7:
@gaborbernat that seemed to be the case, I removed that safeguard, and the suite now ostensibly passes just fine.

To be precise in good faith though, IIRC my proposed regression test did not catch the issue even if I undid the fix. Probably there are subtle differences between running `tox` standalone vs via integration tests. I'll recheck.

Or maybe it is good enough for now? Cause I don't know how to reproduce this in tests short of a "functional" run.

src/tox/tox_env/python/virtual_env/package/pyproject.py
```diff
@@ -118,8 +118,15 @@ def root(self) -> Path:
 
     @root.setter
     def root(self, value: Path) -> None:
-        self._root = value
-        self._frontend_ = None  # force recreating the frontend with new root
+        # NOTE(vytas): Recreating the frontend with a new root will orphan the
```
gaborbernat, line 121: Align comments to 120 charachters. Remove ` NOTE(vytas)`

docs/changelog/3530.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Prevent tox from hanging upon exit due to orphaned build threads and
```
gaborbernat, line 1: Align to 120 charachters.

src/tox/tox_env/python/virtual_env/package/pyproject.py
```diff
@@ -118,8 +118,15 @@ def root(self) -> Path:
...
+        #   current frontend.backend_executor, if any, making tox hang upon
```
gaborbernat, line 122: We should instead explicitly close down that frontend.

gaborbernat (changes requested):

docs/changelog/3530.bugfix.rst
vytas7: Alright, although other newsfragments in `main` are not aligned to 120 characters.

src/tox/tox_env/python/virtual_env/package/pyproject.py
vytas7: I've verified this approach also works with my MRE from the issue.

Issue #3542, opened 2025-06-03
logged environment variables are not sorted


ssbarnea:
## Issue

When tox logs environment variables it does not sort them first and this makes it harder to read the logs.

I will make PR to fix it, I just needed a issue number for the changelog entry.

PR #3543, opened 2025-06-03
Log environment variables sorted by key while redacting values of unsafe ones

ssbarnea:
Improves logging of environment variables by sorting them by key and redacting the values for the ones that are likely to contain secrets.

Fixes: #3542

### DIFF
--- a/tests/tox_env/test_api.py
+++ b/tests/tox_env/test_api.py
@@ -2,6 +2,10 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+import pytest
+
+from tox.tox_env.api import redact_value
+
 if TYPE_CHECKING:
     from pathlib import Path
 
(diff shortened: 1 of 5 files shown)


tests/tox_env/test_api.py
```diff
@@ -32,3 +36,24 @@ def test_setenv_section_substitution(tox_project: ToxProjectCreator) -> None:
...
     result.assert_success()
+
+
+@pytest.mark.parametrize(
+    ("key", "do_redact"),
+    [
+        pytest.param("ACCESS_TOKEN", True),
+        pytest.param("API_KEY", True),
+        pytest.param("DB_PASSWORD", True),
+        pytest.param("FOO", False),
```
gaborbernat, line 47: Missing some of the ones included into rege x, let's test all of them, and we should document all of the patterns in our docs (users shouldn't need to read code to find it out).

ssbarnea: I will be trying to get some from gitleak internal rules as they are clearly more mature than a list that I just come up in two days. Still, is a copy/paste as that project is not python based and we also don't want to add new dependencies (i suspect).

ssbarnea:
@gaborbernat Any chance you could look at it again today? I think that I addressed the requests.

This patch is a blocker for improving the security of GHA pipelines as I would not want to disable log collection for tox. Thanks.

ssbarnea:
@gaborbernat Sorry to bother you again. Is there anything else I need to change?

gaborbernat (approved):

gaborbernat (dismissed):

ssbarnea:
@gaborbernat The py314 failure is unrelated and seems to be a bug in coveragepy. I will try to help @nedbat with a patch if I do not run out of time. That also made me realize that I need to fix few bugs in tox-uv in order to run coveragepy tests..., side effect raising https://github.com/tox-dev/tox-uv/pull/210

ssbarnea:
@gaborbernat Any idea about what to do about py314 errors related to coverage?
```
CoverageWarning: Couldn't import C tracer: No module named 'coverage.tracer' (no-ctracer)
```

They started to fail on main last night, so unrelated to this patch. Should we attempt to capture/hide these warnings to avoid our test failures?

nedbat:
I think I will skip that warning on platforms where I don't ship a built wheel, though I'd be curious whether the installation of coverage on 3.14 shows that it tried to compile the extension or not?

nedbat:
Coverage 7.9.1 is released to quiet that warning on 3.14.

gaborbernat (approved):

cidrblock:
TY @gaborbernat

PR #3551, opened 2025-06-26
trivial: Prefer '[testenv] pip_pre'

stephenfin:
We have an option specifically for this use case. Tell people about it.


gaborbernat (changes requested):
We strongly encourage people to use environment variables over configurations nowadays, so this still stands.

stephenfin:
Ack. Should the `pip_pre` option be marked as deprecated (you added it in 4.0)?

gaborbernat:
I don't think we need to deprecate it, but if anyone asks do the environment variables instead.

Issue #3565, opened 2025-07-22
Crash with  module 'tarfile' has no attribute 'data_filter' on CPython <3.9.17 and <3.10.12


nicoddemus:
## Issue

Seeing this internal error across multiple packages that use `tox`:

```
py39-xdist: install_deps> python -I -m pip install pytest-xdist>=2.1.0 -e .
.pkg_external: freeze> python -m pip freeze --all
.pkg_external: pip==25.1.1,setuptools==80.9.0
py39-xdist: internal error
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\session\cmd\run\single.py", line 47, in _evaluate
    tox_env.setup()
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\api.py", line 283, in setup
    self._setup_with_env()
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\runner.py", line 146, in _setup_with_env
    self._setup_pkg()
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\runner.py", line 169, in _setup_pkg
    self._packages = self._build_packages()
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\python\runner.py", line 122, in _build_packages
    packages = package_env.perform_packaging(self.conf)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\package.py", line 46, in _func
    return meth(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\python\virtual_env\package\cmd_builder.py", line 96, in perform_packaging
    return self.extract_install_info(for_env, path)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\package.py", line 46, in _func
    return meth(*args, **kwargs)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\tox\tox_env\python\virtual_env\package\cmd_builder.py", line [11](https://github.com/pytest-dev/pytest/actions/runs/16427355653/job/46421357851?pr=13611#step:6:12)0, in extract_install_info
[...]

PR #3568, opened 2025-07-25
Don't pass in the filter argument to tar.extractall on old Python versions

gaborbernat:
Resolves https://github.com/tox-dev/tox/issues/3565

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,9 +25,9 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.12.4"
     hooks:
-      - id: ruff-format
-      - id: ruff
+      - id: ruff-check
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -25,9 +25,9 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: "v0.12.4"
     hooks:
-      - id: ruff-format
-      - id: ruff
+      - id: ruff-check
```
ssbarnea, line 28: @gaborbernat Why have you moved the format after check on ruff? My impression was that it was far better to do the format first and the check with fix after. Reason for this being that reformatting is likely to also include fixes for stuff that check might only complain about, so more noise.

Am I wrong? ChatGPT seems to agree, not that I would trust it so much anyway.

PR #3578, opened 2025-08-08
fix: provide clear messaging about config file loading

ssbarnea:
Fix bug where tox is silently ignoring loading pyproject.toml files
that failed be loaded by tomllint and falling back to other config.

- if pyproject.toml does not have one of the two expected sections
  it will be logged, so it would be visible with `-v`
- make pyproject.toml failure to load a fatal error with clear
  information about cause
- keeps ignoring missing pyproject.toml files as expected

With these, users of tox will not be surprised when they break the
config file by mistake, like adding a duplicate section.


src/tox/config/source/discover.py
```diff
@@ -59,21 +60,32 @@ def _locate_source() -> Source | None:
...
-                return src_type(candidate)
-            except ValueError:
-                pass
+            if candidate.exists():
+                try:
+                    return src_type(candidate)
+                except MissingRequiredConfigKeyError as e:
+                    msg = f"{src_type.__name__} skipped loading {candidate.resolve()} due to {e}"
+                    logging.info(msg)
+                except ValueError as e:
```
gaborbernat, line 69: Don't use single letter variables outside of comprehensions.

gaborbernat (changes requested):

Daverball:
I believe something about this change caused a regression. If you have a `setup.cfg` but your `tox` configuration resides in `pyproject.toml` then tox will complain about the missing `tox:tox` section in `setup.cfg` instead of using the `pyproject.toml` configuration.

PR #3591, opened 2025-08-29
Expose a new `tox_extend_envs` hook in plugins API

### LINKED ISSUE #3510: [bug] Plugins can only inject envs as default but not additional
Over the course of having discoveries related to #3507, I realized that all the examples with injecting `MemoryLoader(env_list=env_list)` add envs as default.

And so typing in `tox` would attempt running them:
```console
$ tox l
default environments:
pip-compile              -> [tox-lock] Invoke pip-compile of pip-tools
pip-compile-build-lock   -> [tox-lock] Produce a PEP 517/660 build deps lock
pip-compile-tox-env-lock -> [tox-lock] Produce a lock file for the passed tox env using current python

additional environments:
build-dists              -> Build dists and put them into the dist/ folder
pre-commit               -> Run the quality checks under python3; run as `SKIP=check-id1,check-id2 tox r -e pre-commit` to instruct the underlying `pre-commit` invocation avoid running said checks; Use `tox r -e pre-commit -- check-id1 --all-files` to select checks matching IDs aliases: `tox r -e pre-commit -- mypy --all-files` will run 3 MyPy invocations, but `tox r -e pre-commit -- mypy-py313 --all-files` runs one.
build-docs               -> Build The Docs
```

I don't want to influence whatever the end-user configures as `env_list` in their config, only providing additional envs that should be requested via `tox r -e <env>` (or through the env var, or config explicitly).

It seems to me like a gap in the plugin API. Having stared long enough at the source code, it seems that the discovery of additional envs is tightly coupled with the source objects. It's effectively happening @ https://github.com/tox-dev/tox/blob/a985171b574a7f0da45b6f5bcba7b587b3b25ea9/src/tox/session/env_select.py#L180.
And if we unwrap a few layers of indirection, that effectively does `state.conf._src._discover_tox_envs(state.conf.core)`.

It looks like this currently means that adding a memory loader into the core config adds default envs that are processed separately when discovering all envs: https://github.com/tox-dev/tox/blob/a985171b574a7f0da45b6f5bcba7b587b3b25ea9/src/tox/session/env_select.py#L172. And this is what makes it work.
But adding “inactive” / “additional” envs would only be possible by monkey-patching the private `state.conf._src` object or `state.conf.__iter__()`.

@gaborbernat should the plugin API just implement a `tox_discover_envs()` hook or a method on the `state` or `config` objects for declaring additional environments?

webknjaz:
This patch adds a new hook point in the tox lifecycle that lets plugin authors declare additional environment names dynamically. This can be used to provide extra tox environments shipped as installable plugins.

Resolves #3510.

- [x] updated/extended the documentation: https://tox--3591.org.readthedocs.build/en/3591/plugins.html#tox.plugin.spec.tox_extend_envs

### DIFF
--- a/docs/changelog/3510.feature.rst
+++ b/docs/changelog/3510.feature.rst
@@ -1,0 +1,1 @@
+3591.feature.rst
--- a/docs/changelog/3591.feature.rst
+++ b/docs/changelog/3591.feature.rst
@@ -1,0 +1,10 @@
+A new tox life cycle event is now exposed for use via :doc:`Plugins
+API </plugins>` -- by :user:`webknjaz`.
+
+The corresponding hook point is :func:`tox_extend_envs
+<tox.plugin.spec.tox_extend_envs>`. It allows plugin authors to
+declare ephemeral environments that they can then populate through
+the in-memory configuration loader interface.
+
+This patch was made possible thanks to pair programming with
+:user:`gaborbernat` at PyCon US 2025.
--- a/src/tox/config/main.py
+++ b/src/tox/config/main.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import os
 from collections import OrderedDict, defaultdict
+from itertools import chain
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Iterator, Sequence, TypeVar
+from typing import TYPE_CHECKING, Any, Iterable, Iterator, Sequence, TypeVar
 
 from .sets import ConfigSet, CoreConfigSet, EnvConfigSet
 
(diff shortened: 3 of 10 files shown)


gaborbernat (commented):
I'm ok with this if you add a tesdt demonstrating it.

webknjaz:
Yep, that's why it's still a draft. I've been playing with a PoC in a `toxfile.py` of another project locally and confirmed that this is usable. I'll need to look into turning that into test coverage and add a change note before undrafting.

webknjaz:
cc @sirosen @jugmac00 — this is a new API that @gaborbernat and I hacked together at PyCon US in the spring. FYI.

gaborbernat (approved):

webknjaz:
@gaborbernat thanks for merging 🙏

Would you mind releasing so I could use this in a plugin, finally?

gaborbernat:
https://github.com/tox-dev/tox/releases/tag/4.29.0

PR #3598, opened 2025-09-02
Ensure `tox_extend_envs` list can be read twice

webknjaz:
Previously (PR #3591) a new hook point was added. The tests checked that it works well with the `tox config` and `tox list` commands. However, `tox run` turned out to have a problem — it would complain that there's no such env, when invoked:

```console
ROOT: 170 E HandledError| provided environments not found in configuration file:
pip-compile-tox-env-lock [tox/run.py:23]
```

Turned out, this was because the config object is being interated twice in some subcommands. This in turn iterates over the discovered additional ephemeral environments list object. The implementation passes an iterator into it and so when it's first accessed, it's exhausted and the second attempt does not give the same envs, causing inconsistency.

The patch solves this by using `itertools.tee()`, making sure that the underlying iterable is always cached and it's possible to repeat iteration as many times as possible without loosing the data in the process.

### DIFF
--- a/src/tox/config/main.py
+++ b/src/tox/config/main.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 from collections import OrderedDict, defaultdict
-from itertools import chain
+from itertools import chain, tee
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Iterable, Iterator, Sequence, TypeVar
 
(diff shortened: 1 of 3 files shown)


webknjaz:
Urgh.. it looks like the fix doesn't work on older Pythons (3.11 and lower). I'll take a look.

webknjaz:
@gaborbernat this is now ready for inclusion in v4.29.1

webknjaz:
@gaborbernat could you release when you have time?

PR #3601, opened 2025-09-03
Ensure automatically provisioned environment is torn down

### LINKED ISSUE #3600: Automatically provisioned Tox hangs on pyproject_api if `--installpkg` is used
## Issue
Automatically provisioned Tox hangs on pyproject_api backend if `--installpkg` is used.

At first glance, this seems to be very similar to #3512, except that instead of multiple environments the trigger now is recursively provisioning `tox` in a separate _venv_. A single environment is enough to reproduce.

I'll take a quick debugging session to see if the reason could be any similar to #3512.
In any case, it seems that my fix from #3530 does not help in this scenario.

A quick workaround exists, though -- it is enough to install the missing dependencies into the current environment (normally just `setuptools` was enough in the example at hand).

## Environment
CPython 3.12 on Ubuntu/Debian Linux.

## Minimal example

Clone https://github.com/ansible/pylibssh that @webknjaz is maintaining.

Build an sdist:
```
python -m build --sdist
```

Run `tox`:
```
tox --installpkg dist/ansible_pylibssh-*.tar.gz --notest
```

In the case of recursively provisioned environment, `tox` hangs. Otherwise it terminates cleanly.

vytas7:
Fixes #3600.

- [x] wrote descriptive pull request text (_see the newsfragment_)

### DIFF
--- a/src/tox/tox_env/package.py
+++ b/src/tox/tox_env/package.py
@@ -97,7 +97,8 @@ class PackageToxEnv(ToxEnv, ABC):
         self._envs.add(run_env.conf.name)
 
     def teardown_env(self, conf: EnvConfigSet) -> None:
-        self._envs.remove(conf.name)
+        if conf.name in self._envs:
+            self._envs.remove(conf.name)
         if len(self._envs) == 0:
             self._teardown()
 
--- a/src/tox/tox_env/package.py
+++ b/src/tox/tox_env/package.py
@@ -98,6 +98,7 @@ class PackageToxEnv(ToxEnv, ABC):
 
     def teardown_env(self, conf: EnvConfigSet) -> None:
         if conf.name in self._envs:
+            # conf.name (".tox") may be missing in self._envs in the case of an automatically provisioned environment
             self._envs.remove(conf.name)
         if len(self._envs) == 0:
             self._teardown()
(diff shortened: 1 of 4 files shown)


src/tox/tox_env/package.py
```diff
@@ -97,7 +97,8 @@ def mark_active_run_env(self, run_env: RunToxEnv) -> None:
         self._envs.add(run_env.conf.name)
 
     def teardown_env(self, conf: EnvConfigSet) -> None:
-        self._envs.remove(conf.name)
+        if conf.name in self._envs:
```
vytas7, line 100: Is that an acceptable workaround @gaborbernat?

Otherwise this throws an exception since `conf.name` (`".tox"`) is not found in `self._envs` when tearing down the automatically provisioning environment.

vytas7:
I had no energy to write any tests either since the issue doesn't affect me, and there exists an easy workaround.

But if you could provide pointers how such a test could look like, I can try to hack together something...

webknjaz:
> But if you could provide pointers how such a test could look like, I can try to hack together something...

@vytas7 maybe just extend/copy https://github.com/tox-dev/tox/blob/b33a43c270eb87e61833b7ed934b401d23aeded5/tests/tox_env/python/virtual_env/package/test_package_pyproject.py#L469-L487 / https://github.com/tox-dev/tox/blob/b33a43c270eb87e61833b7ed934b401d23aeded5/tests/tox_env/python/virtual_env/package/test_package_cmd_builder.py#L180-L196 adding a lightweight dep?

vytas7:
> adding a lightweight dep?

Right, I hacked something together based on the existing examples. I needed to add a mirrored Devpi index to be able to fetch `setuptools` in the automatically provisioned environment, not sure if there is a better way.

PR #3624, opened 2025-10-08
Add 3.14, drop 3.9 and support | union style

gaborbernat:
Resolves https://github.com/tox-dev/tox/pull/3624.

Adds support for the `|` operator for unions.


gaborbernat:
Resolves https://github.com/tox-dev/tox/pull/3624

PR #3629, opened 2025-10-17
TST: add weekly compatibility checks for CPython 3.15

neutrinoceros:
Python 3.15 is still early in its development and I don't know that there's a precedent for adding "support" quite this early in the cycle, but given there are 0 failing tests at the moment, I figure it would might still be useful so:
- later breaking changes are detected soon-ish
- this would help any downstream package tested with tox knowing they can try out 3.15 pre-releases safely (or at least that issues might already be known here).

Anyway, I think this is worth a shot. Let me know if this is too early.

- [N/A] updated/extended the documentation

### DIFF
--- a/.github/workflows/weekly.yaml
+++ b/.github/workflows/weekly.yaml
@@ -1,0 +1,56 @@
+name: weekly checks
+on:
+  workflow_dispatch:
+  push:
+    branches: ["main"]
+    tags-ignore: ["**"]
+  pull_request:
+    paths:
+      - .github/workflows/weekly.yaml
+  schedule:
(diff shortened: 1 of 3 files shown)


gaborbernat:
I'm always reluctant to add this so early because there's no easy way for the check to fail but not fail the entire CI 🤔 would want to have that before we commit to this.

neutrinoceros:
It could also run as a weekly cron job instead of regular CI. What do you think ?

gaborbernat:
That would work better 🤔

gaborbernat (approved):

Issue #3614, opened 2025-09-26
dash vs. comma format in bracketed section names in .ini is either broken or needs better docs


zzzeek:
## Issue

in the docs, we see that there's a syntax we can use like this:

```
"Both enumerations ({1,2,3}) and numerical ranges ({1-3}) are supported, and can be mixed together:"
[tox]
env_list = py3{8-10, 11, 13-14}
```

wow!  i never knew that.  the dash part.

But that's in the env_list.  Where is the specification for this format as used elsewhere?  we see an example like this:

```
[testenv]
deps =
    py311,py310: urllib3
    py{311,310}-sqlite: mock
```

Can we use the dash format there too?   Nope.   The bug is, that every `{}` syntax should work everywhere (preferred!) or the docs need to really talk about what's allowed where and be very specific what's going on.

Below is an example of these syntaxes.  It illustrates that while we can use `py{39,310,311,312,313}` just fine in a key, we cannot use `py{39-313}` - gets ignored silently.    As a bonus, someone today also tried to use the `{}` nesting syntax that works for `{env:}` with the environment name tags, that doesnt work either.

Ive been listing out `py{39,310,311}` type things for many years, and it's tedious.  It seems like tox by now should have gained some more flexibility with these things.     Is this....better in TOML format ?   is .ini the old way replaced by TOML ?

Example:

```
[tox]
envlist = py

[extras1]
deps=
    greenlet

[extras2]
deps=
    typing-extensions

[extras3]
deps=
    requests


[testenv]

deps=
     # a syntax that works.
    somespec-py{38,39,310,311,312,313,314}: {[extras1]deps}

     # can we use the dash syntax instead of listing out every version?
     # nope
    somespec-py{38-314}: {[extras2]deps}

     # can we nest inside these things?  nope, this syntax is not accepted
     # somespec-{py{312,313},otherspec}: {[extras3]deps}

allowlist_externals =
    sh

commands=
    sh -c echo  "hello world"
```

run this as:

```
$ tox -re somespec-py313 --notest
```

only greenlet gets installed.  typing-extensions does not:

```
[classic@framework tmp]$ tox -re somespec-py313 --notest
somespec-py313: remove tox env folder /home/classic/tmp/.tox/somespec-py313
somespec-py313: install_deps> python -I -m pip install greenlet
  somespec-py313: OK (0.78 seconds)
  congratulations :) (0.81 seconds)

```

gaborbernat:
[...]

Issue #3571, opened 2025-07-30
Braced range expansion in `tox.envlist` vs `testenv`


moreati:
## Issue

1. In `tox.envlist` braced lists and ranges are expanded the same. So `foo{3-5}` <=> `foo{3,4,5}` <=> `foo3 foo4 foo5`.
2. In `testenv` only braced lists are expanded. So `foo{3-5}` is *not* equivalent to `foo{3,4,5}` or `foo3,foo4,foo5`

I'm unsure if the different behaviour in `testenv`  is intended or a bug. The docs at https://tox.wiki/en/latest/config.html#generative-environment-list don't explicitly state either way.
 
## Environment

Provide at least:

- OS:

<details open>
<summary>Output of <code>pip list</code> of the host Python, where <code>tox</code> is installed</summary>

```console
➜  tox_range uv pip list -p ~/.local/share/uv/tools/tox
Package                   Version
------------------------- --------
ansible-compat            25.6.0
ansible-core              2.19.0
attrs                     25.3.0
cachetools                6.1.0
cffi                      1.17.1
chardet                   5.2.0
colorama                  0.4.6
cryptography              45.0.5
distlib                   0.4.0
execnet                   2.1.1
filelock                  3.18.0
iniconfig                 2.1.0
jinja2                    3.1.6
jsonschema                4.25.0
jsonschema-specifications 2025.4.1
markupsafe                3.0.2
packaging                 25.0
pip                       25.1.1
platformdirs              4.3.8
pluggy                    1.6.0
pycparser                 2.22
pygments                  2.19.2
pyproject-api             1.9.1
pytest                    8.4.1
pytest-ansible            25.6.3
pytest-plus               0.8.1
pytest-sugar              1.0.0
pytest-xdist              3.8.0
pyyaml                    6.0.2
referencing               0.36.2
resolvelib                1.2.0
rpds-py                   0.26.0
subprocess-tee            0.4.2
termcolor                 3.1.0
tox                       4.28.3
tox-ansible               25.5.5
tox-uv                    1.26.2
typing-extensions         4.14.1
uv                        0.8.3
virtualenv                20.32.0
```

</details>

## Output of running tox

<details open>
<summary>Output of <code>tox -rvv</code></summary>

```console
➜  tox_range tox -rvv
py313-foo1: 151 W remove tox env folder /Users/alex/tmp/tox_range/.tox/py313-foo1 [tox/tox_env/api.py:357]
[...]

PR #3631, opened 2025-10-20
Allow braced range syntax in internal sections of tox.ini file

marcosboger:
Closes https://github.com/tox-dev/tox/issues/3571
Partially fixes https://github.com/tox-dev/tox/issues/3614 -> This PR covers the complaint about using dash syntax in `deps`. It does not address the complaint about nested braces syntax.

This PR refactors how `expand_ranges` is used when loading .ini config files. Instead of calling `expand_ranges` directly when expanding envs or section names, we expand it once we start finding the factor groups, in `find_factor_groups()` function. 

This means that ranges in sections like `deps` and `setenv` will also be expanded. Before, only env_list and section names were expanded. Considered as bugfix due to the original issue classification, but could also arguably be considered a small feature.

The approach was done in this way so that we don't have to "remember" to call expand_ranges everywhere that we use the brace syntax {}. Instead, whenever we are expanding factors, transparently the ranges will be expanded.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1663,6 +1663,12 @@ Both enumerations (``{1,2,3}``) and numerical ranges (``{1-3}``) are supported, 
     [tox]
     env_list = py3{8-10, 11, 13-14}
 
+    [testenv]
+    deps =
+        py{310,311-314}: urllib3
+    setenv =
+        py{310,311-314}: FOO=bar
+
 will create the following envs:
 
 .. code-block:: shell


marcosboger:
Two doubts, open to input:

- Would adding examples to the docs already be enough? Or should we add a new section explaining specifically this?
- Should we also create a more complete parameterized test like `test_env_list_expands_ranges` for internal sections? Would be kinda redundant as they'll be leveraging same function for expansion, but open to hear opinions

gaborbernat (approved):

PR #3692, opened 2026-02-14
🚀 perf(tests): prevent stale fixture state and fix flaky timing assertions

gaborbernat:
Test suite had stale `.tox` directories persisting in test fixture folders between runs, causing potential hangs during environment validation. Additionally, a timing-sensitive assertion in the spinner test was flaky on slower systems. 🐌

The fixes add cleanup to the `demo_pkg_inline` session-scoped fixture to remove `.tox` directories before and after test runs, preventing state pollution. The spinner test assertion now checks for flexible timing ("0" to "0.01" seconds) rather than exact zero seconds. Test markers (`@pytest.mark.slow` and `@pytest.mark.integration`) categorize tests by execution time, enabling developers to run fast subsets via `tox -e fast`. CI gains speedup through wheel caching in `.wheel_cache/`, saving ~1.5s per job.

The main benefit is robustness: preventing stale fixture state from causing hangs and making timing assertions resilient to system variations. ✨ The fast environment runs 1938 non-slow, non-integration tests in ~49 seconds with parallel execution, providing quick feedback during development. Provision tests remain 20-30 seconds as they're integration tests that actually create virtual environments—correct behavior ensuring tox provisioning works reliably.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -42,6 +42,14 @@ jobs:
           enable-cache: true
           cache-dependency-glob: "pyproject.toml"
           github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Cache wheel downloads
+        uses: actions/cache@v4
+        with:
+          path: .wheel_cache
+          key: wheel-cache-${{ runner.os }}-${{ matrix.py }}-${{ hashFiles('pyproject.toml') }}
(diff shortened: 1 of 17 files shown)


gaborbernat:
This is now available in [version 4.36.0](https://github.com/tox-dev/tox/releases/tag/4.36.0) 🎉

PR #3706, opened 2026-02-15
🐛 fix(release): make release script more robust

gaborbernat:
The release script had two issues that caused the failed 4.36.0 release:

**Issue 1: Hardcoded remote name**
The cleanup step hardcoded `upstream/main` but GitHub Actions uses `origin` as the remote name. Fixed by using the dynamically resolved remote name from the `get_upstream()` function.

**Issue 2: Cleanup on failure**
The `finally` block always executed cleanup even when pushes failed, leaving the remote in a bad state while local state was cleaned up. This made debugging and recovery difficult.

Changed to only cleanup after successful pushes. If an error occurs before pushing, the remote release branch is cleaned up automatically.

**Testing**
Tested locally with both upstream and origin remote configurations.


gaborbernat:
Squashed into the main PR commit instead via force push

PR #3715, opened 2026-02-16
🐛 fix(execute): adopt CPython subprocess stream handling

gaborbernat:
Subprocess output reading could deadlock on Windows during parallel test execution or when handling large amounts of output. The root cause was a chicken-and-egg problem where reader threads waited for the subprocess to close its pipes, but the subprocess wouldn't be terminated until after the readers exited. 🔒 This manifested as timeouts in CI when tox tried to clean up long-running backend processes.

The fix adopts CPython's approach to subprocess stream handling. On Unix, we switched from `select.select()` to the `selectors` module with proper EOF detection and interrupt handling. On Windows, we use overlapped I/O with non-blocking polling, mirroring CPython's implementation. Most importantly, we reordered the shutdown sequence in `pep517_backend.py` to terminate subprocesses before stopping reader threads, ensuring pipes close and pending I/O operations complete naturally. ⚡

This eliminates arbitrary timeouts and race conditions in the process cleanup logic. The implementation now handles EINTR signals gracefully and reads larger chunks (32KB instead of 1KB) for better performance with high-volume output.

**References:**
- CPython's subprocess implementation: https://github.com/python/cpython/blob/main/Lib/subprocess.py
- Windows overlapped I/O: https://docs.microsoft.com/en-us/windows/win32/fileio/synchronous-and-asynchronous-i-o
- Python `selectors` module: https://docs.python.org/3/library/selectors.html
- `_overlapped` module: https://github.com/python/cpython/blob/main/Modules/overlapped.c

### DIFF
--- a/src/tox/execute/local_sub_process/read_via_thread_windows.py
+++ b/src/tox/execute/local_sub_process/read_via_thread_windows.py
@@ -29,18 +29,12 @@ class ReadViaThreadWindows(ReadViaThread):  # pragma: win32 cover
                 except OSError:
                     break
 
-                start_time = time.monotonic()
-                max_wait = 1.0
                 while True:
                     try:
                         data = ov.getresult(False)  # noqa: FBT003
                         break
                     except OSError as exception:
                         if getattr(exception, "winerror", None) != ERROR_IO_INCOMPLETE:
-                            return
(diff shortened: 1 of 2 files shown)


gaborbernat:
This is now available in [version 4.36.1](https://github.com/tox-dev/tox/releases/tag/4.36.1) 🎉

PR #3725, opened 2026-02-17
Fix factor selection via TOX_FACTORS environment variable

### LINKED ISSUE #3557: Factor parsing is broken when using env vars
## Issue

Dug a bit into #3555. The first issue was a bit of confusion over the env var naming scheme. The CLI displays a metavar of `factor`, but the actual argument dest is `factors`, so the env var lookups are `TOXFACTORS`/`TOX_FACTORS`, not `TOXFACTOR`/`TOX_FACTOR`. You can't really determine what the env vars should be without looking through the source code, and I don't know if there's an easy or good way to convey this in the CLI help text. Regardless, this was a bit confusing.

As to the actual bug, factor parsing is broken when provided by an env var. I've debugged this as far as I can for now, but essentially it seems that [`get_env_var`](https://github.com/tox-dev/tox/blob/4.27.0/src/tox/config/cli/env_var.py#L28-L30) converts the env value to a list, but [`EnvSelector._parse_factors`](https://github.com/tox-dev/tox/blob/4.27.0/src/tox/session/env_select.py#L372-L377) is expecting the `raw_factors` to be a list of lists. This can be demonstrated with: 

```console
 TOX_FACTORS="3.12 django" tox -f 3.13 django r 
```

Setting a breakpoint in `EnvSelector._mark_active`, the parsed `factors` are:

```console
({'o', '1', ' ', 'd', '2', 'g', '3', 'a', 'n', '.', 'j'}, {'3.13', 'django'})
```

## Minimal example

```console
$ tox -f 3.13 l
3.13 -> run the tests with pytest under 3.13
$ TOX_FACTORS=3.13 tox l
$
```

worksbyfriday:
## Summary
- Fix `TOX_FACTORS` env var producing wrong results (each character treated as a separate factor instead of each factor string)
- Root cause: `get_type()` inferred `list[str]` for `_AppendAction` with `nargs="+"`, but the actual runtime type is `list[list[str]]`
- The existing `Convert` infrastructure already handles nested generics, so fixing the type inference is sufficient

Closes #3557

## Test plan
- [x] Added parametrized test `test_factor_select_via_env_var` covering single factor, AND (comma-separated), and OR (semicolon-separated)
- [x] All 65 env_select tests pass
- [x] All 6355 config tests pass
- [x] All 6 CLI env var tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/tox/config/cli/parser.py
+++ b/src/tox/config/cli/parser.py
@@ -66,7 +66,7 @@ class ArgumentParserWithEnvAndConfig(ArgumentParser):
         of_type: type[Any] | None = getattr(action, "of_type", None)
         if of_type is None:
             if isinstance(action, argparse._AppendAction):  # noqa: SLF001
-                if action.nargs in ("+", "*") or (isinstance(action.nargs, int) and action.nargs > 1):
+                if action.nargs in {"+", "*"} or (isinstance(action.nargs, int) and action.nargs > 1):
                     of_type = list[list[action.type]]  # ty: ignore[invalid-type-form] # nargs produces list per invocation
                 else:
                     of_type = list[action.type]  # ty: ignore[invalid-type-form] # runtime generic from argparse action type
--- a/tests/session/test_env_select.py
+++ b/tests/session/test_env_select.py
@@ -157,8 +157,19 @@ def test_factor_select(
     [
         ("cov", ("py310-django20-cov", "py310-django21-cov", "py39-django20-cov", "py39-django21-cov")),
         ("py39,django20", ("py39-django20-cov", "py39-django20")),
-        ("py39;py310", ("py310-django20-cov", "py310-django20", "py310-django21-cov", "py310-django21",
-                        "py39-django20-cov", "py39-django20", "py39-django21-cov", "py39-django21")),
+        (
+            "py39;py310",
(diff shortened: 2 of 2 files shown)


worksbyfriday:
Rebased on main to pick up latest changes. Will mark ready for review once CI passes.

worksbyfriday:
Rebased on main, all tests pass. The only failures are `type`/`type-min` which also fail on main. Ready for review.

worksbyfriday:
Understood — converted back to draft. All test jobs pass (3.10-3.14 on Ubuntu/macOS/Windows), but the `type`/`type-min` jobs are failing. Those same jobs also fail on main — is there something I need to do to fix them, or are they a known issue on the main branch?

worksbyfriday:
Note: the type/type-min CI failures are caused by `argcomplete` (optional dependency) not being installed in the ty check environment. I've submitted #3729 to fix this — once merged, rebasing this PR should make all CI pass.

worksbyfriday:
Rebased on main (includes #3728). All CI checks passing — type/type-min now green. Marking ready for review.

gaborbernat (approved):

PR #3738, opened 2026-02-17
🐛 fix(legacy): --parallel-no-spinner no longer suppresses output

### LINKED ISSUE #3193: [4.12.0][regression] The presence of `TOX_PARALLEL_NO_SPINNER=1` in CI suppresses output of single env run tox commands
## Issue

I have a number for globally set environment variables, one of each is `TOX_PARALLEL_NO_SPINNER`: https://github.com/ansible/pylibssh/blob/ef77a85/.github/workflows/ci-cd.yml#L75.

In one of the steps, I run `tox -e a-single-env -qq | tee a-file.txt`. In tox 4.11.4, it produces output from the command specified in `tox.ini`. Starting with tox 4.12.0, the output is empty.

I'm pretty convinced this is a side effect of #3159.

Previously, setting this var allowed having it in one place, while allowing for parallel and non-parallel runs different job steps, not having to track whether to disable or enable the spinner in each. So I consider this to be a regression.

## Environment

- OS: ubuntu-latest @ GHA

`tox == 4.12.0`

## Output of running tox

No output.

## Minimal example

```console
$ TOX_PARALLEL_NO_SPINNER=1 tox -e a-single-env --skip-pkg-install -qq
```

gaborbernat:
Since tox 4.12.0, setting `TOX_PARALLEL_NO_SPINNER=1` or passing `--parallel-no-spinner` in the legacy command unconditionally forced parallel execution mode, even when no `-p` flag was given. 🔇 In parallel mode, output is buffered and only shown on failure, so CI environments that set this variable lost all command output from successful runs.

The fix removes `parallel_no_spinner` from the condition that routes between parallel and sequential execution in the legacy command. The flag now only controls spinner display when parallel mode is already active via `-p`. The `--parallel-no-spinner` help text is also corrected to reflect this intended behavior.

⚠️ This is a behavioral change for users who relied on `--parallel-no-spinner` implying `--parallel` in the legacy command. Those users should explicitly pass `-p` alongside `--parallel-no-spinner` to get parallel execution.

Fixes #3193

### DIFF
--- a/docs/changelog/3193.bugfix.rst
+++ b/docs/changelog/3193.bugfix.rst
@@ -1,0 +1,2 @@
+``TOX_PARALLEL_NO_SPINNER`` / ``--parallel-no-spinner`` no longer forces parallel mode in the legacy command, fixing
+output suppression for sequential runs in CI - by :user:`gaborbernat`.
--- a/src/tox/session/cmd/legacy.py
+++ b/src/tox/session/cmd/legacy.py
@@ -110,7 +110,7 @@ def legacy(state: State) -> int:
             option.env = CliEnv(["py"])
         option.devenv_path = Path(option.devenv_path)
         return devenv(state)
-    if option.parallel_no_spinner is True or option.parallel != 0:  # only 0 means sequential
+    if option.parallel != 0:  # only 0 means sequential
         return run_parallel(state)
     return run_sequential(state)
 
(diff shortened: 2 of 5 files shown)


sparkiegeek:
This is great but needs at least a mention on #3158 since it explicitly undoes that change and underlying request.

Issue #3276, opened 2024-05-02
Regression: Complex factor conditions (including negative factor conditions) completely omitted from recent (v4?) docs


posita:
## Issue

From https://tox.wiki/en/3.27.1/config.html#complex-factor-conditions :

    Sometimes you need to specify the same line for several factors or create a special case for a combination of factors. ... They have all been added only to help demonstrate the following:

    * how factor expressions get expanded the same way as in envlist
    * how to use negated factor conditions by prefixing negated factors with !
    * that the order in which factors are hyphenated together does not matter

An equivalent section seems to be completely absent from https://tox.wiki/en/latest/

gaborbernat:
PR welcome.

posita:
This likely goes beyond that single section. You might want to audit how much of your documentation was lost along the way. In my experience, the former docs were much more navigable/readable.

webknjaz:
Oh, I was just about to report the same, having been unable to find negative factors anymore.. I wonder if some LLM could be pointed to tox3 and tox4 docs for the purposes of finding what's missing.

webknjaz:
Out of curiosity, I ran Claude Code against those and here's what it generated: https://claude.ai/public/artifacts/f76ec3e8-3d8b-446a-af13-e15ef63b70c6. Some things are probably irrelevant, but there's a few ideas worth fixing.

gaborbernat:
This is now available in [version 4.36.0](https://github.com/tox-dev/tox/releases/tag/4.36.0) 🎉

PR #3753, opened 2026-02-18
🐛 fix(env): reject partial testenv section matches

### LINKED ISSUE #3219: Partial match on testenv results in use of base testenv

gaborbernat:
When a testenv section uses generative syntax like `[testenv:functional{-py310}]`, only `functional-py310` should be a valid environment. However, running `tox -e functional-py312` or `tox -e functional` silently passed validation and fell back to the base `[testenv]`, producing unexpected results. 🔍 This was a regression from tox 3, where these would correctly error as unknown environments.

The root cause was that `_ensure_envs_valid` decomposed *all* known env names — including those from section headers — into individual factors, then treated them as freely combinable. So `functional` (from `functional-py310`) became a valid factor that could be mixed with any Python version factor. The fix distinguishes between factors that are genuinely combinable (from `env_list` generative expressions and conditional markers) and complete env names from section headers. Section names are recognized as whole identifiers but their component parts can no longer be arbitrarily recombined via `-e`.

⚠️ This is a behavioral change: users who relied on partial section name matching to silently fall back to `[testenv]` will now get a clear error message pointing them to the correct environment name. Matrix-style `env_list` combinations like `py312-django50` continue to work as expected.

Fixes #3219

### DIFF
--- a/docs/changelog/3219.bugfix.rst
+++ b/docs/changelog/3219.bugfix.rst
@@ -1,0 +1,3 @@
+Environments like ``functional-py312`` no longer incorrectly match when only ``functional{-py310}`` is defined as a
+testenv section -- factors from section headers are no longer treated as freely combinable with CLI ``-e`` selections -
+by :user:`gaborbernat`.
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -211,11 +211,17 @@ class EnvSelector:
             yield label_envs.keys(), False
 
     def _ensure_envs_valid(self) -> None:
-        valid_factors = set(chain.from_iterable(env.split("-") for env in self._state.conf))
-        valid_factors.add(".pkg")  # packaging factor
+        known_envs = set(self._state.conf)
+        # factors that can be freely combined: from env_list entries and known env names themselves
+        combinable = set(chain.from_iterable(env.split("-") for env in self._state.conf.core["env_list"]))
+        combinable.update(known_envs)
+        combinable.add(".pkg")
+        # broader pool for suggestions includes factors from all known env names
+        all_factors = set(chain.from_iterable(env.split("-") for env in known_envs))
+        all_factors.update(combinable)
         invalid_envs: dict[str, str | None] = {}
         for env in self._cli_envs or []:
-            if env.startswith(".pkg_external"):  # external package
+            if env.startswith(".pkg_external") or env in known_envs:
                 continue
             factors: dict[str, str | None] = dict.fromkeys(env.split("-"))
             found_factors: set[str] = set()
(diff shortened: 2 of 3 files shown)


edabor:
Hello !

First, let me thank you for providing such powerful tool. It has greatly helped simplifying CI in [pyvista](https://github.com/pyvista/pyvista)

It seems that this PR now prevents using complex factor conditions as described in https://tox.wiki/en/3.27.1/config.html#complex-factor-conditions
I recognize it was documented in an old version of `tox` that disappeared in latest docs (https://github.com/tox-dev/tox/issues/3276) but it was working as I expected it before this PR.

Here is MWE with the behavior changes induced by this PR:

```ini
[tox]

env_list = py3.{10-14}

[testenv]

deps =
    matplotlib
    np: numpy
    np-cov: coverage
```

| command                | before this PR                          | This PR                                                              |
|------------------------|-----------------------------------------|----------------------------------------------------------------------|
| `tox r -e py3.10-np-cov`| installs coverage, matplotlib and numpy | HandledError \| provided environments not found in configuration file|

We managed to find a workaround in https://github.com/pyvista/pyvista/pull/8343 by explicitly declaring factors but I was curious if this behavior change introduced by this PR was intentional.

Thanks in advance.

PR #3797, opened 2026-02-20
✨ feat(toml): add generative env_list via product dict

gaborbernat:
TOML users have had to manually enumerate every environment in `env_list` when they need a test matrix, while INI users could write concise expressions like `py3{12,13}-django{42,50}`. This gap made TOML configuration verbose and error-prone for projects with large matrices, and was one of the most visible feature disparities between the two formats.

✨ This introduces a TOML-native `product` dict syntax that leverages TOML's type system instead of mimicking INI's string-based brace expansion. Factor groups are expressed as arrays of strings or range dicts, and tox computes the Cartesian product joining combinations with `-`. Range dicts (`{ prefix = "py3", start = 12, stop = 14 }`) generate sequential factors, with open-ended bounds matching INI's `py3{10-}` behavior. An `exclude` key allows skipping specific combinations — a capability not available in INI at all.

```toml
env_list = [
    "lint",
    { product = [
        { prefix = "py3", start = 12, stop = 14 },
        ["django42", "django50"],
    ] },
]
```

Documentation has been updated across all four Diataxis dimensions (tutorial, how-to, reference, explanation) with TOML shown first in all tab pairs. The "TOML feature gaps" section has been renamed to "Format comparison" since generative `env_list` was the last major functional gap — only generative section names remain INI-exclusive. The JSON schema has been extended to validate `product` dicts with their factor groups and optional `exclude` arrays.

### DIFF
--- a/docs/changelog/3797.feature.rst
+++ b/docs/changelog/3797.feature.rst
@@ -1,0 +1,2 @@
+Add TOML-native generative ``env_list`` via ``product`` dict syntax -- Cartesian product of factor groups with optional
+range dicts and exclusions - by :user:`gaborbernat`.
--- a/docs/explanation.rst
+++ b/docs/explanation.rst
@@ -199,9 +199,10 @@ detects changes and applies them automatically.
 Open-ended range bounds
 =======================
 
-INI generative environment lists support open-ended ranges like ``py3{10-}`` and ``py3{-13}``. Instead of probing the
-system for available interpreters (which would be slow and environment-dependent), tox tracks the `supported CPython
-versions <https://devguide.python.org/versions/>`_ via two constants:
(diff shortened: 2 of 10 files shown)


nedbat:
The tox.ini syntax was compact and understandable.  Why not allow that same syntax in .toml?  This dictionary seems cumbersome.

gaborbernat:
> The tox.ini syntax was compact and understandable.  Why not allow that same syntax in .toml?  This dictionary seems cumbersome.

I think this is one of those cases where it was understandable for people who knew. I was constantly reminded when talking to people unfamiliar with the project that they have no idea what's happening. Furthermore learning from the mistakes of various string markers being very difficult to compose I purposefully wanted to not use any string level information and have this schema be fully type safe via native toml. See how now we have a concrete IDE automatically validated schema generated and published.

PR #3847, opened 2026-03-03
fix: handle OSError in pos_args path existence check

yarikoptic:
When positional arguments exceed the filesystem's filename length limit (typically 255 characters), `Path.exists()` raises `OSError: [Errno 36] File name too long` instead of returning `False`.

This causes `tox` to crash with an internal error when **long** `-k` expressions are passed via `{posargs}`, e.g.:

    tox -e py3 -- -k "test_foo or test_bar or test_baz or ..."

The traceback is:

    File ".../tox/config/main.py", line 58, in pos_args
        if path_arg.exists() and not path_arg.is_absolute():
    OSError: [Errno 36] File name too long: 'test_foo or test_bar or ...'

This is because `pos_args()` checks each argument with `path_arg.exists()` to determine whether to rewrite relative paths for `changedir`. An argument that is too long for the filesystem clearly cannot be a path, so the fix catches `OSError` and treats such arguments as non-paths.

- [ ] ran the linter to address style issues (`tox -e fix`) [XXX]
- [x] added news fragment in `docs/changelog` folder -- 

<details><summary>[XXX]: tried but failed to install sphinx... might try again later</summary>

```
[INFO] Using pre-commit with uv 0.10.7 via pre-commit-uv 4.2.1
An unexpected error has occurred: CalledProcessError: command: ('/home/yoh/proj/misc/tox/.tox/fix/bin/uv', '--project', '/home/yoh/proj/misc/tox', 'pip', 'install', '.', 'sphinx>=9.1')
return code: 1
stdout: (none)
stderr:
    Using Python 3.11.12 environment at: py_env-python3
      × No solution found when resolving dependencies:
      ╰─▶ Because the current Python version (3.11.12) does not satisfy
          Python>=3.12 and sphinx==9.1.0 depends on Python>=3.12, we can conclude
          that sphinx==9.1.0 cannot be used.
          And because only sphinx<=9.1.0 is available and you require sphinx>=9.1,
          we can conclude that your requirements are unsatisfiable.
Check the log at /home/yoh/.cache/pre-commit/pre-commit.log
fix: exit 3 (24.65 seconds) /home/yoh/proj/misc/tox> pre-commit run --all-files --show-diff-on-failure pid=4127791
  fix: FAIL code 3 (25.28=setup[0.62]+cmd[24.65] seconds)
  evaluation failed :( (25.34 seconds)

```
</details>

### DIFF
--- a/docs/changelog/3847.bugfix.rst
+++ b/docs/changelog/3847.bugfix.rst
@@ -1,3 +1,3 @@
 Handle ``OSError`` when checking if positional arguments are existing paths in ``pos_args()`` — long arguments (e.g.,
-pytest ``-k`` expressions exceeding the 255-character filesystem name limit) no longer crash with
-``OSError: [Errno 36] File name too long`` - by :user:`yarikoptic`.
+pytest ``-k`` expressions exceeding the 255-character filesystem name limit) no longer crash with ``OSError: [Errno 36]
+File name too long`` - by :user:`yarikoptic`.


yarikoptic:
that was quick -- kudos!

PR #3875, opened 2026-03-12
🧪 test(docs): validate doc config examples to prevent regressions

gaborbernat:
Discussion #3874 showed that many TOML examples in the documentation were silently broken — `replace = "if"` conditionals inside `deps` and `commands` lists were missing `extend = true`, causing tox to fail with type errors when users copied them. Fixing these by hand is fragile; the examples will drift again with the next doc edit. 📝

This PR adds a parametrized test that dynamically discovers every `.. code-block:: toml` and `.. code-block:: ini` directive across all RST files under `docs/`, classifies each block by config format (`tox.toml`, `pyproject.toml`, `tox.ini`, `setup.cfg`), and runs `tox c` to verify the config parses cleanly. The test currently validates 138 code blocks across 7 RST files. Blocks requiring unavailable infrastructure (specific Python interpreters, auto-provisioning plugins, virtualenv version pinning) are detected at collection time and skipped with clear reasons.

The doc fixes themselves add the missing `extend = true` to ~35 conditional replacement examples and correct a broken `ref of` path in the raw reference documentation (`["env", "extras"]` → `["env", "src", "extras"]`). 🐛 Going forward, any new or edited config example that breaks will be caught by CI before it reaches users.

### DIFF
--- a/docs/changelog/3875.doc.rst
+++ b/docs/changelog/3875.doc.rst
@@ -1,0 +1,2 @@
+Fix broken TOML examples missing ``extend = true`` on conditional replacements inside lists, fix incorrect ``ref of``
+path in raw reference docs, and add a test that validates all doc config examples - by :user:`gaborbernat`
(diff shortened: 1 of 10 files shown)


luziferius:
Thanks for the fixes!

PR #3887, opened 2026-03-18
🐛 fix(config): apply testenv overrides to TOML env templates

### LINKED ISSUE #3860: TOX_OVERRIDE for pass_env shows in config but doesn't actually pass env vars
Setting `TOX_OVERRIDE="testenv.pass_env+=NIX_LD,NIX_LD_LIBRARY_PATH"` makes `tox c` show the variables in `pass_env`, but they are not actually passed through to the tox environment at runtime.

Setting it directly in `tox.ini` works fine. The override mechanism appears to update the config view but not the actual behavior.

Ref: https://github.com/tox-dev/tox/discussions/3225

gaborbernat:
Users setting `TOX_OVERRIDE` with the `testenv` prefix found their overrides ignored in TOML-based projects, while the same syntax worked perfectly in INI configs. 🔧 For example, `testenv.pass_env+=CUSTOM_VAR` would appear in the config output but the variable wouldn't actually pass through at runtime, breaking CI pipelines and NixOS setups that rely on environment overrides.

TOML uses different section naming conventions than INI. Where INI uses `testenv` as the section prefix, TOML uses `env_run_base` for base templates and `tool.tox.env` for individual environments. The loader only looked up overrides by exact section key match, so `testenv.*` overrides never matched TOML section names and were silently dropped.

The fix maps the familiar `testenv` namespace to TOML-style sections by checking the override map for `testenv` entries and applying them to TOML env templates and env-prefixed sections. ✨ This preserves the consistent `testenv.*` override syntax that users expect to work across both config formats, without requiring them to learn TOML-specific override paths.

Fixes #3860

### DIFF
--- a/src/tox/config/source/ini.py
+++ b/src/tox/config/source/ini.py
@@ -53,10 +53,13 @@ class IniSource(Source):
             # if no matching section/prefix is found, use the requested section key as-is (for custom prefixes)
             key = section.key
         if self._parser.has_section(key):
+            overrides = list(override_map.get(section.key, []))
+            if section.prefix is not None:
+                overrides.extend(override_map.get(section.prefix, []))
             return IniLoader(
                 section=section,
                 parser=self._parser,
-                overrides=override_map.get(section.key, []),
+                overrides=overrides,
                 core_section=self.CORE_SECTION,
                 section_key=key,
             )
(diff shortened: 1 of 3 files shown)


gaborbernat:
Closing this PR - mapping INI syntax to TOML is error prone and confusing. Users should use the proper syntax for their config format.

PR #3912, opened 2026-03-30
✨ feat(runner): add PEP 723 inline script metadata support

### LINKED ISSUE #3897: Pep 723 support

We should allow creating environments fewly from the pep723 header file. 

This would require a new environment type that pretty much replaces a lot of our current configuration, such as the Python version or dependency specification. 

## Alternative Solutions

N.A.

gaborbernat:
Closes #3897.

PEP 723 lets Python scripts declare their dependencies and required Python version inline via comment blocks. Currently tox users must duplicate this information in their tox config, which drifts out of sync and defeats the purpose of the standard. The new `virtualenv-pep-723` runner reads metadata directly from the script file, so a tox environment needs only a `runner` and `script` key.

```ini
[testenv:check]
runner = virtualenv-pep-723
script = tools/check.py
```

The shared PEP 723 logic lives in `Pep723Mixin` at `tox.tox_env.python.pep723`, independent of any venv backend. The concrete `Pep723Runner` composes the mixin with `VirtualEnv` and `RunToxEnv`, skipping `PythonRun` entirely so config keys like `deps`, `extras`, `dependency_groups`, and `pylock` are structurally absent rather than registered and rejected. Packaging is unconditionally disabled via `_register_package_conf` returning `False`. Setting `base_python` explicitly is rejected to prevent conflicts with the script's `requires-python` specifier.

The mixin design means tox-uv can add PEP 723 support by composing `Pep723Mixin` with `UvVenv` and `RunToxEnv` — zero code duplication.

### DIFF
--- a/docs/changelog/3897.feature.rst
+++ b/docs/changelog/3897.feature.rst
@@ -1,0 +1,2 @@
+Add ``virtualenv-pep-723`` runner that reads dependencies and Python version from :PEP:`723` inline script metadata — no
+need to duplicate them in tox config - by :user:`gaborbernat`.
--- a/docs/explanation.rst
+++ b/docs/explanation.rst
@@ -245,6 +245,46 @@ invocation. No configuration changes will be needed.
 **Change detection** works the same as for ``deps``: tox caches the resolved requirements. When packages are added, only
 the new ones are installed. When packages are removed, the environment is recreated.
 
+.. _pep723-explanation:
+
+Inline script metadata (PEP 723)
+================================
+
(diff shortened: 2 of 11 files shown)


rahuldevikar (approved):

henryiii:
> The mixin design means tox-uv can add PEP 723 support by composing Pep723Mixin with UvVenv and RunToxEnv — zero code duplication.

tox-uv will also need an update to use it there?

gaborbernat:
Yes.

PR #3951, opened 2026-06-02
🐛 fix(config): propagate overrides through config references

### LINKED ISSUE #3950: `TOX_OVERRIDE` does not propagate through config references when overriding a *base configuration*
## Issue

`TOX_OVERRIDE` does not propagate through config references when overriding a *base configuration* value that is later referenced via `{replace = "ref", ...}` (TOML) or `{[testenv]key}` (INI). The override is ignored and the original file value is used instead.

### Acknowledgment

I'm aware this is a sensitive area with several related fixes in the past (#3127, #3348, #3736, #3737). I've reviewed those issues and believe this is a distinct case that wasn't covered by the previous fixes.

### TOML format (pyproject.toml)

```toml
[tool.tox.env_run_base]
extras = ["RED"]

[tool.tox.env.test]
extras = [
  {replace = "ref", of = ["tool", "tox", "env_run_base", "extras"], extend = true},
  "GREEN"
]
```

```bash
# Override is ignored:
TOX_OVERRIDE="tool.tox.env_run_base.extras=BLUE" tox config -e test -k extras
# Output: extras = green, red  (expected: blue, green)

# Workaround - override the final environment directly:
TOX_OVERRIDE="tool.tox.env.test.extras=BLUE" tox config -e test -k extras
# Output: extras = blue  (this works, but defeats the purpose)
```

### INI format (tox.ini)

```ini
[testenv]
extras = RED

[testenv:test]
extras = {[testenv]extras} GREEN
```

```bash
# Override is ignored:
TOX_OVERRIDE="testenv.extras=BLUE" tox config -e test -k extras
# Output: extras = RED GREEN  (expected: BLUE GREEN)

# Workaround - override the final environment directly:
TOX_OVERRIDE="testenv:test.extras=BLUE" tox config -e test -k extras
# Output: extras = BLUE  (this works, but defeats the purpose)
```

### Expected Behavior

When using `TOX_OVERRIDE="tool.tox.env_run_base.extras=BLUE"` with the TOML config above, the `test` environment should resolve `extras` to `["BLUE", "GREEN"]` because the override should apply to `env_run_base.extras` before it's referenced.

### Observed Behavior

The override is ignored during reference resolution. The `test` environment resolves `extras` to `["RED", "GREEN"]`, using the original file value instead of the overridden value.

### Root Cause Analysis

I'm not familiar with the codebase and have relied on LLM code assistance for analysis, so I defer to maintainers whether this is a bug or a behavior that needs documentation. The issue appears to be that reference resolution bypasses the override system:

- **TOML:** `Unroll._replace_ref()` calls `load_raw_from_root()` which directly reads from the parsed TOML dictionary without checking for overrides
- **INI:** `ReplaceReferenceIni` similarly loads raw configuration values without going through the override machinery

The override system is designed to work at the `Loader.load()` level, but reference resolution happens at a lower level that accesses raw config data directly.

### Use Case

This pattern is useful for shared configuration where different environments need to override base values via `TOX_OVERRIDE` (e.g., in CI pipelines or development scripts) while still benefiting from the DRY principle of config references.

I'd be happy to work on a fix (probably relying on LLM code assistants) if the maintainers can provide guidance on the preferred approach. Thank you for maintaining tox!

## Environment

- OS: Linux (Fedora 44, kernel 7.0.10)
- Python: 3.14.5

<details open>
<summary>Output of <code>pip list</code> of the host Python, where <code>tox</code> is installed</summary>

```console
$ pip show tox
Name: tox
Version: 4.55.0
Summary: tox is a generic virtualenv management and test command line tool
Home-page: http://tox.readthedocs.org
Author: 
Author-email: Bernát Gábor <gaborjbernat@gmail.com>
License-Expression: MIT
Location: /home/tdaapare/.local/lib/python3.14/site-packages
Requires: cachetools, colorama, filelock, packaging, platformdirs, pluggy, pyproject-api, python-discovery, tomli-w, virtualenv
Required-by:
```

</details>

## Output of running tox

<details open>
<summary>Output of <code>TOX_OVERRIDE="tool.tox.env_run_base.extras=BLUE" tox config -e test -k extras</code></summary>

```console
$ TOX_OVERRIDE="tool.tox.env_run_base.extras=BLUE" tox config -e test -k extras
[testenv:test]
extras =
  green
  red
```

**Expected output:**
```console
[testenv:test]
extras =
  blue
  green
```

</details>

tales-aparecida:
`TOX_OVERRIDE` (and the `-x`/`--override` flag) silently lost effect whenever it targeted a base value that another environment pulled in by reference. Overriding `env_run_base.extras` and then reading it from an environment that referenced that base returned the original file value, which defeated the common pattern of overriding shared configuration once and having every dependent environment inherit it — handy in CI pipelines and local scripts. 🔧

Overrides are normally applied inside `Loader.load`, but reference resolution reads values straight from the parsed config and never reaches that path, so the override was dropped. The fix folds any matching override back onto the raw value at the two places references resolve — the `of` form of a TOML `{replace = "ref", ...}` and the `{[section]key}` form in ini — through a shared helper that reads from a new `Config.overrides` view and infers the shape (list, dict, or scalar) from the value being overridden.

Both replacement (`=`) and append (`+=`) propagate now. The `{replace = "ref", env = ..., key = ...}` TOML form already worked because it resolves through a full environment load, so only the raw-read paths needed this. ⚠️ ini references are textual, so appending to a base value that sits inline with other tokens splices the appended item into that line — expected given how ini substitution works, and noted in the docs.

Fixes #3950.

### DIFF
--- a/docs/changelog/3950.bugfix.rst
+++ b/docs/changelog/3950.bugfix.rst
@@ -1,0 +1,4 @@
+``TOX_OVERRIDE`` (and ``-x``/``--override``) now propagates through configuration references. Previously, overriding a
+base value that was later referenced via ``{[section]key}`` (ini) or ``{replace = "ref", of = [...]}`` (toml) was
+ignored because reference resolution read the raw file value, bypassing the override system - by
+:user:`tales-aparecida`. (:issue:`3950`)
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -2234,6 +2234,50 @@ Or reset override and append to that (note the first override is ``=`` and not `
 
        tox -x testenv.deps=pytest-xdist -x testenv.deps+=pytest-cov
 
+Overrides propagate through references
+======================================
+
+.. versionadded:: 4.56
+
+An override targeting a base value is applied before that value is referenced elsewhere, so overriding shared
+configuration once updates every environment that references it. Given a base ``extras`` that the ``test`` environment
+extends:
+
+.. tab:: TOML
+
+    .. code-block:: toml
+
+       [tool.tox.env_run_base]
+       extras = [ "red" ]
+
+       [tool.tox.env.test]
+       extras = [
+         { replace = "ref", of = [ "tool", "tox", "env_run_base", "extras" ], extend = true },
+         "green",
+       ]
+
+    Overriding the base propagates into ``test`` (``test`` resolves to ``blue, green``):
+
+    .. code-block:: bash
+
+       tox c -e test -k extras -x tool.tox.env_run_base.extras=blue
+
+.. tab:: INI
+
+    .. code-block:: ini
+
+       [testenv]
+       extras = red
+
+       [testenv:test]
+       extras = {[testenv]extras} green
+
+    Overriding the base propagates into ``test`` (``test`` resolves to ``blue green``):
+
+    .. code-block:: bash
+
+       tox c -e test -k extras -x testenv.extras=blue
+
 Escaping dots in override keys
 ==============================
 
--- a/src/tox/config/loader/api.py
+++ b/src/tox/config/loader/api.py
@@ -220,6 +220,41 @@ class Loader(Convert[T]):
         return self.to(raw, of_type, factory)
 
 
+def apply_overrides_to_raw(overrides: Iterable[Override], key: str, value: T) -> T:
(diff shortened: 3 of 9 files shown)


tales-aparecida:
This is mostly a reproducer for https://github.com/tox-dev/tox/issues/3950

tales-aparecida:
Thanks for working on this!

PR #3966, opened 2026-06-23
🐛 fix(config): restore skip_missing_interpreters default to False

gaborbernat:
The tox 4 rewrite unintentionally changed the default of `skip_missing_interpreters` from `False` to `True`. This means tox reports success even when configured environments can't find their Python interpreter — a typo in `env_list` or a genuinely missing interpreter goes unnoticed.

This restores the tox 3 behavior: missing interpreters fail the run by default. Users who rely on the current behavior can set `skip_missing_interpreters = true` in their tox config or pass `--skip-missing-interpreters` on the command line.


fghaas:
I quite liked the changed default. :) [xkcd1172](https://xkcd.com/1172/)

PR #3989, opened 2026-07-19
Run sequentially for run-parallel -p 0 instead of at CPU count

chuenchen309:
### Problem

`tox run-parallel -p 0` runs every environment at once (CPU count), even though the `-p` help says **"zero is turn off"**:

```
-p VAL, --parallel VAL   run tox environments in parallel, the argument controls
                         limit: all, auto - cpu count, some positive number,
                         zero is turn off
```

So `-p 0` behaves identically to `-p auto`, not as "off".

### Cause

`run_parallel` computed:

```python
max_workers=auto_detect_cpus() if option.parallel == 0 else option.parallel
```

`OFF_VALUE` is `0` and the legacy command already treats `parallel == 0` as sequential (`legacy.py`: `if option.parallel != 0:  # only 0 means sequential`), but `run_parallel` mapped `0` to the CPU count. This was a leftover from when the argument default *was* the CPU count; since #3109 made the default `"auto"`, `0` now only arrives from an explicit `-p 0`.

### Fix

Map `-p 0` to a single worker (sequential), matching the help text, `OFF_VALUE`, and the legacy path:

```python
max_workers=1 if option.parallel == OFF_VALUE else option.parallel
```

`-p all`/`-p auto`/`-p N` are unchanged.

### Testing

Added `test_parallel_zero_turns_off` (mocks `execute`, asserts `max_workers == 1` for `-p 0`); it fails on `main` (`max_workers` is the CPU count) and passes here. `tests/session/cmd/test_parallel.py` + legacy tests pass; `ruff` clean.

---

*Disclosure: authored by an AI coding agent (Claude Code) running on this account — it found the bug, wrote the repro and test, and this description. I review every change and am accountable for it; the verification is real and re-runnable from the diff.*

### DIFF
--- a/docs/changelog/3989.bugfix.rst
+++ b/docs/changelog/3989.bugfix.rst
@@ -1,2 +1,2 @@
-Make ``tox run-parallel -p 0`` turn parallelism off and run sequentially, as the ``-p`` help ("zero is turn off")
-and the legacy command already promise, instead of running at the CPU count - by :user:`chuenchen309`.
+Make ``tox run-parallel -p 0`` turn parallelism off and run sequentially, as the ``-p`` help ("zero is turn off") and
+the legacy command already promise, instead of running at the CPU count - by :user:`chuenchen309`.
--- a/docs/changelog/3989.bugfix.rst
+++ b/docs/changelog/3989.bugfix.rst
@@ -1,0 +1,2 @@
+Make ``tox run-parallel -p 0`` turn parallelism off and run sequentially, as the ``-p`` help ("zero is turn off")
+and the legacy command already promise, instead of running at the CPU count - by :user:`chuenchen309`.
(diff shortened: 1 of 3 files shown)


gaborbernat (changes requested):
So what's the way to run it with cpu count now?

chuenchen309:
`-p auto` — that's the documented "auto - cpu count" mode, and since #3109 it's also the default (a bare `-p` / `tox p` already runs at CPU count). This PR doesn't touch that path; it only stops `-p 0` from being a second, undocumented alias for `-p auto`, so that `0` matches what the `-p` help already promises ("zero is turn off") and what the legacy command already does. So after this change: `-p auto` (or the default) = CPU count, `-p N` = N workers, `-p 0` = sequential.

gaborbernat (approved):

PR #4012, opened 2026-08-03
🔥 chore: remove the misc changelog category

gaborbernat:
A changelog entry filed under `misc` tells the reader nothing. Anything real belongs under one of the seven remaining categories, so `misc` is gone from towncrier and from the contributing docs.

The filename pre-commit hook had drifted from the towncrier config as well: it allowed `removal`, which is not a category, and turned away `breaking`, `deprecation`, `packaging` and `contrib`. Both lists now match.

### DIFF
--- a/docs/changelog/4012.contrib.rst
+++ b/docs/changelog/4012.contrib.rst
@@ -1,0 +1,2 @@
+Remove the ``misc`` changelog category, and align the changelog filename check with the categories towncrier accepts -
+by :user:`gaborbernat`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -282,5 +282,4 @@ type = [
   { directory = "doc", name = "Improved documentation", showcontent = true },
   { directory = "packaging", name = "Packaging updates and notes for downstreams", showcontent = true },
   { directory = "contrib", name = "Contributor-facing changes", showcontent = true },
-  { directory = "misc", name = "Miscellaneous internal changes", showcontent = true },
 ]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,8 +62,8 @@ repos:
       - id: changelogs-rst
         name: changelog filenames
         language: fail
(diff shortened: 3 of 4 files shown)


docs/changelog/4012.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Remove the ``misc`` changelog category, and align the changelog filename check with the categories towncrier accepts -
```
webknjaz, line 1: Honestly, the problem, in my opinion, isn't that the category is catch-all (well, it isn't great), but the cognitive effect of how it leads people to compose prose for the release notes in general. Humans are rather bad at writing coherent chunks of text following certain style unless you police them heavily. This doesn't just go for misc/other/etc. All the categories are affected. It's probably okay to keep it as long as you stick https://vale.sh and/or an LLM that would help the contributor iterate over the thing, extract the useful bits out of their mind and then turn them into adequate human-consumable fragments that fit well w/ the rest of the change log following common guidelines...

pyproject.toml
webknjaz, line 1: Side note: I've been considering having like highlights/prelude/foreword/banner and afterword in my various projects. Inspired by reno among other things. Ended up added some of these (highlights+afterword) in pip-tools.

Here's something to feed into your agent if you're ever interested in the related discussions/debates: https://github.com/jazzband/pip-tools/pull/2279#discussion_r2558075686 + https://github.com/jazzband/pip-tools/issues/2287.

Plus the implementation: https://github.com/jazzband/pip-tools/pull/2322.

Note that we've migrated our Towncrier changelog to Markdown (MyST-Parser on the Sphinx level) some time earlier, it plays better with more ecosystems.

docs/changelog/4012.contrib.rst
gaborbernat: I disagree, if something is not a bug/feature/docs/removal doesn't need a changelog entry.

pyproject.toml
gaborbernat: I'm fine with we have for now.

PR #4020, opened 2026-08-11
Add {home} and {tox_root_name} substitutions

WhyNotHugo:
Basically, I want to configure:

```dosini
[tox]
work_dir = ~/.local/state/tox/{tox_root_name}
```

These two patches enable this to work.

`docs/changelog` looks empty; there's no fragments there (and nothing to use as reference).

### DIFF
--- a/src/tox/config/sets.py
+++ b/src/tox/config/sets.py
@@ -289,7 +289,10 @@ class CoreConfigSet(ConfigSet):
         return self.get("work_dir", Path) / ".tmp"
 
     def _work_dir_post_process(self, folder: Path) -> Path:
-        return self._conf.work_dir if self._conf.options.work_dir else folder
+        if self._conf.options.work_dir:
+            return self._conf.work_dir
+        raw = str(folder).replace("{tox_root_name}", self._root.name).replace("${home}", str(Path.home()))
+        return Path(raw)
 
     def register_config(self) -> None:
         self.add_constant(keys=["config_file_path"], desc="path to the configuration file", value=self._src_path)
--- a/src/tox/config/sets.py
+++ b/src/tox/config/sets.py
@@ -289,10 +289,7 @@ class CoreConfigSet(ConfigSet):
         return self.get("work_dir", Path) / ".tmp"
 
     def _work_dir_post_process(self, folder: Path) -> Path:
-        if self._conf.options.work_dir:
-            return self._conf.work_dir
-        raw = str(folder).replace("{tox_root_name}", self._root.name).replace("${home}", str(Path.home()))
-        return Path(raw)
+        return self._conf.work_dir if self._conf.options.work_dir else folder
 
     def register_config(self) -> None:
         self.add_constant(keys=["config_file_path"], desc="path to the configuration file", value=self._src_path)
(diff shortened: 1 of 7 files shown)


WhyNotHugo:
The ["Timeline protection"](https://github.com/tox-dev/tox/pull/4020/checks?check_run_id=93820001358) output leads to "Error Fetching Resource", not sure what's up there.

gaborbernat (changes requested):
Generally better to start with an issue and not a PR. That being said I'm strong -1 on ~ support,  instead add a `{home}` constant users can refer to. Also note https://github.com/tox-dev/tox/pull/4017, we no longer accept INI format improvements.

CAOShurong:
I tested `ce2b0da2926f135ab10ba8ac3691766bcbcaada7` on native Windows 11 with CPython 3.13.1.

- `python -m pytest -q tests/config/test_sets.py` → `1 failed, 22 passed`; the failing test is `test_config_work_dir_expands_user`.
- I checked the six GitHub Windows jobs (Python 3.10–3.15); all six failure logs identify that test, while the Ubuntu and macOS jobs pass.
- A direct `pathlib` matrix shows the cause: changing only `HOME` leaves `Path('~').expanduser()` at the Windows profile; changing `USERPROFILE` changes the result. Removing `USERPROFILE` still falls back to the Windows profile rather than `HOME`.
- I also exercised the requested user-config workflow on this commit. With `TOX_USER_CONFIG_FILE` pointing to a config containing `work_dir = ~\state\tox\{tox_root_name}` and `USERPROFILE` set to an E: test profile, `tox c --core -k work_dir --format json` returned the expected `E:\...\state\tox\tox-pr-4020-native-windows` path.

So the current Windows failure comes from the test's POSIX-only `HOME` assumption; runtime `~` expansion follows native Windows `pathlib` semantics. This is test evidence only: it does **not** address the requested design change from `~` to a `{home}` constant, and I have not pushed changes to the contributor branch.

gaborbernat:
@CAOShurong stop your agent posting comments here. We are fine with using an agent to do the work, but agents transient outputs should not be posted here.

WhyNotHugo:
> -1 on ~ support, instead add a {home} constant users can refer to.

Switched to `{home}`.

> we no longer accept INI format improvements.

I've adapted this to work on `tox.toml` instead, although it also works for the ini too, but mostly as a side effect.

src/tox/config/sets.py
```diff
@@ -289,7 +289,10 @@ def _default_temp_dir(self, conf: Config, env_name: str | None) -> Path:  # ruff
         return self.get("work_dir", Path) / ".tmp"
 
     def _work_dir_post_process(self, folder: Path) -> Path:
-        return self._conf.work_dir if self._conf.options.work_dir else folder
+        if self._conf.options.work_dir:
+            return self._conf.work_dir
+        raw = str(folder).replace("{tox_root_name}", self._root.name).replace("${home}", str(Path.home()))
```
gaborbernat: Definitely not replaces like this, tox already has builtin support for substitution and we should use that.

gaborbernat (changes requested):

gaborbernat (approved):

PR #4022, opened 2026-08-12
Provision before loading env_list

### LINKED ISSUE #4021: Provisioning fails when config has backwards incompatible options
Hi. Real scenario with older tox:

 1. Install tox < 4.43
 2. Create tox config that requires tox >= 4.43 and uses `product` in `env_list` (feature added in tox 4.43)
 3. Run tox

Expected: tox 4.43 or newer is provisioned, `env_list` is read with new tox and it supports prodcut.

Actual: Tox correctly determines that provisioning is needed, but fails to provision because `env_list` is not valid.

```toml
# tox.toml
requires = ["tox>=4.43"]
env_list = [
    { product = [{ prefix = "3.", start = 10, stop = 15 }]},
]
```

```bash
(venv)$ uv pip install 'tox<4.43'
...
(venv)$ tox -l
ROOT: will run in automatically provisioned tox, host .../venv/bin/python3 is missing [requires (has)]: tox>=4.43 (4.42.0)
  File ".../venv/bin/tox", line 10, in <module>
    sys.exit(run())
             ~~~^^
  File ".../venv/lib64/python3.14/site-packages/tox/run.py", line 23, in run
  File ".../venv/lib64/python3.14/site-packages/tox/run.py", line 45, in main
    result = provision(state)
  File ".../venv/lib64/python3.14/site-packages/tox/provision.py", line 133, in provision
  File ".../venv/lib64/python3.14/site-packages/tox/provision.py", line 153, in run_provision
    tox_env: PythonRun = cast("PythonRun", state.envs[name])
                                           ~~~~~~~~~~^^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/session/env_select.py", line 520, in __getitem__
    env = self._defined_envs[item].env
          ^^^^^^^^^^^^^^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/session/env_select.py", line 338, in _defined_envs
    env_name_to_active = self._env_name_to_active()
  File ".../venv/lib64/python3.14/site-packages/tox/session/env_select.py", line 309, in _env_name_to_active
    for a_collection, is_active in self._collect_names():
                                   ~~~~~~~~~~~~~~~~~~~^^
  File ".../venv/lib64/python3.14/site-packages/tox/session/env_select.py", line 237, in _collect_names
    env_list, everything_active = self._state.conf.core["env_list"], False
                                  ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/config/sets.py", line 150, in __getitem__
    return self.load(item)
           ~~~~~~~~~^^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/config/sets.py", line 162, in load
    return config_definition.__call__(self._conf, self.loaders, ConfigLoadArgs(chain, self.name, self.env_name))  # noqa: PLC2801
           ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/config/of_type.py", line 113, in __call__
    value = loader.load(primary_key, self.of_type, self.factory, conf, args, all_keys=alias_keys)
  File ".../venv/lib64/python3.14/site-packages/tox/config/loader/api.py", line 153, in load
    converted = self.build(alias, of_type, factory, conf, raw, args)
  File ".../venv/lib64/python3.14/site-packages/tox/config/loader/toml/__init__.py", line 75, in build
    result = self.to(exploded, of_type, factory)
  File ".../venv/lib64/python3.14/site-packages/tox/config/loader/convert.py", line 49, in to
    return cast("V", self.to_env_list(raw))
                     ~~~~~~~~~~~~~~~~^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/config/loader/toml/__init__.py", line 125, in to_env_list
    return EnvList(envs=list(TomlLoader.to_list(value, str)))
                             ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/config/loader/toml/__init__.py", line 100, in to_list
    result = validate(value, cast("type[list[Any]]", list[of_type]))  # ty: ignore[invalid-type-form] # runtime generic from type variable
  File ".../venv/lib64/python3.14/site-packages/tox/config/loader/toml/_validate.py", line 33, in validate
    validate(va, entry_type)
    ~~~~~~~~^^^^^^^^^^^^^^^^
  File ".../venv/lib64/python3.14/site-packages/tox/config/loader/toml/_validate.py", line 74, in validate
    raise TypeError(msg)
TypeError: {'product': [{'prefix': '3.', 'start': 10, 'stop': 15}]} is not of type 'str'
```

----

Reproducer with current tox 4.59.0 and a hypothetical future feature:

```toml
# tox.toml
requires = ["tox>=6"]
env_list = [1, 2, 3]  # hypothetical future syntax
```

```bash
(venv)$ tox -l
ROOT: will run in automatically provisioned tox, host .../venv/bin/python3 is missing [requires (has)]: tox>=6 (4.59.0)
ROOT: HandledError| failed to load core.env_list: env_list items must be strings, product dicts, range dicts, or labeled dicts, got int
```

I would expect this to fail because tox>=6 is not a thing yet. But it won't get to that.

CAOShurong:
Fixes #4021.

When tox decides it must provision a newer version, `EnvSelector` still reads the current version's `env_list` while constructing the provision environment. If the requested tox introduced new `env_list` syntax, that read fails before the newer tox can be installed.

This change stops environment-name collection after selecting the provision environment when provisioning is active. The provisioned tox then parses the rest of the configuration itself. Normal runs and the no-provision path keep their existing behavior.

The regression test uses an intentionally unsupported `env_list` value with `tox>=999` and intercepts execution. Before the fix, the current tox raises while parsing `env_list`; after the fix, it reaches the provision execution path.

Validation:

- `python -m pytest tests/test_provision.py --run-integration -q` — 28 passed
- `python -m pytest -m "not integration and not slow" -n auto --no-cov -q` — 7973 passed, 26 skipped, 2 unrelated failures; both existing Windows tests hard-code the English WinError 2 text, while this Traditional Chinese Windows installation returns the localized equivalent
- pre-commit on all changed files — passed
- `ty check --python-platform all --output-format concise --error-on-warning .` — passed
- `mypy` — passed
- `pyrefly check` — passed with 0 errors
- wheel and sdist build, `twine check`, `check-wheel-contents`, and `towncrier build --draft` — passed
- real Windows workflow: patched tox 4.42 provisioned tox 4.59, listed product-generated environments 3.10 through 3.15, and ran the selected 3.13 environment successfully
- real failure path: an unavailable `tox>=999` requirement reached dependency installation and exited nonzero instead of failing on the old `env_list` parser

Checklist:

- [x] ran the linter to address style issues
- [x] ensured there are tests validating the fix
- [x] added a news fragment in `docs/changelog`
- [x] documentation update is not needed; this restores the documented provisioning boundary without changing user-facing configuration

### DIFF
--- a/tests/test_provision.py
+++ b/tests/test_provision.py
@@ -178,6 +178,21 @@ def test_provision_requires_checks_true_markers(tox_project: ToxProjectCreator) 
     assert "pkg-does-not-exist" in outcome.out
 
 
+def test_provision_before_loading_env_list(tox_project: ToxProjectCreator) -> None:
+    project = tox_project({
+        "tox.toml": """
+            requires = ["tox>=999"]
+            env_list = [1]
+        """,
+    })
+    project.patch_execute(lambda _: 0)
+
+    outcome = project.run("l")
+
+    outcome.assert_success()
+    assert "will run in automatically provisioned tox" in outcome.out
+
+
 @pytest.mark.integration
 @pytest.mark.usefixtures("_pypi_index_self")
 @pytest.mark.timeout(120)
--- a/tests/test_provision.py
+++ b/tests/test_provision.py
@@ -178,19 +178,53 @@ def test_provision_requires_checks_true_markers(tox_project: ToxProjectCreator) 
     assert "pkg-does-not-exist" in outcome.out
 
 
-def test_provision_before_loading_env_list(tox_project: ToxProjectCreator) -> None:
+@pytest.mark.integration
+@pytest.mark.timeout(120)
+@pytest.mark.parametrize("subcommand", ["l", "c"])
+def test_provision_before_loading_env_list(
+    tox_project: ToxProjectCreator,
+    tmp_path: Path,
+    monkeypatch: MonkeyPatch,
+    pkg_builder: Callable[[Path, Path, Sequence[DistributionType], bool], Path],
+    subcommand: str,
+) -> None:
+    future_tox = tmp_path / "future-tox"
+    (future_tox / "tox").mkdir(parents=True)
+    (future_tox / "pyproject.toml").write_text(
+        """
+        [build-system]
+        requires = ["hatchling"]
+        build-backend = "hatchling.build"
+
+        [project]
+        name = "tox"
+        version = "999"
+        """,
+        encoding="utf-8",
+    )
+    (future_tox / "tox" / "__init__.py").write_text("", encoding="utf-8")
+    (future_tox / "tox" / "__main__.py").write_text(
+        "import sys\n\n"
+        'command = next(arg for arg in sys.argv[1:] if arg in {"l", "c"})\n'
+        'print(f"future tox ran {command}")\n',
+        encoding="utf-8",
+    )
+    wheel = pkg_builder(tmp_path / "dist", future_tox, ["wheel"], False)
+    monkeypatch.setenv("PIP_FIND_LINKS", str(wheel.parent))
+    monkeypatch.setenv("PIP_NO_INDEX", "1")
+
     project = tox_project({
         "tox.toml": """
             requires = ["tox>=999"]
             env_list = [1]
         """,
     })
-    project.patch_execute(lambda _: 0)
 
-    outcome = project.run("l")
+    outcome = project.run(subcommand)
 
     outcome.assert_success()
     assert "will run in automatically provisioned tox" in outcome.out
+    assert f"future tox ran {subcommand}" in outcome.out
 
 
 @pytest.mark.integration
--- a/docs/changelog/4021.bugfix.rst
+++ b/docs/changelog/4021.bugfix.rst
@@ -1,0 +1,2 @@
+Provision the requested tox version before reading ``env_list``, allowing configuration syntax introduced by that
+version - by :user:`CAOShurong`
--- a/src/tox/session/env_select.py
+++ b/src/tox/session/env_select.py
@@ -257,6 +257,9 @@ class EnvSelector:
         """:returns: sources of tox environments defined with name and if is marked as target to run"""
         if self._provision is not None:  # pragma: no branch
             yield (self._provision[1],), False
+            if self._provision[0]:
+                # The provisioned tox may support configuration that this version does not.
+                return
         env_list, everything_active = self._state.conf.core["env_list"], False
         if self._cli_envs is None or self._cli_envs.is_default_list:
             yield env_list, True


tests/test_provision.py
```diff
@@ -178,6 +178,21 @@ def test_provision_requires_checks_true_markers(tox_project: ToxProjectCreator)
     assert "pkg-does-not-exist" in outcome.out
 
 
+def test_provision_before_loading_env_list(tox_project: ToxProjectCreator) -> None:
```
gaborbernat: Can we add a test  here that confirm that both `tox l` and `tox c` operate as expected; e.g. we need to actually spot check the output of `l` and `c` and not just assert that `"will run in automatically provisioned tox" in outcome.out` is present.

gaborbernat (changes requested):

tests/test_provision.py
hroncok: That's hard to simulate, as it would likely require a fake installable future tox version, wouldn't it?

gaborbernat: Could do the other way, install an old version that would fail, no?

hroncok: Install an older version and assert that the failure is originating from that one? Sounds possible as long as the old version supports the Python version that runs the tests.

CAOShurong: Added in 21de6da9. I used a local installable `tox==999` wheel with `PIP_NO_INDEX` instead of a published old tox: installing an old release would exercise provisioning, but would not distinguish this ordering regression because the current host can already parse old-version configuration. The parametrized test now runs both `tox l` and `tox c` and asserts the provisioned process prints `future tox ran l` / `future tox ran c`.

Spot-checks: the same test on base cb7eeca7 fails both cases before provisioning with `failed to load core.env_list ... got int`; on head 21de6da9 both pass. `tests/test_provision.py --run-integration` (29 passed), `tests/session/test_env_select.py` (92 passed), target-file pre-commit, and `tox -e type` all pass.

gaborbernat (approved):

PR #4036, opened 2026-08-21
docs: drop the claim that tox -e py skips tests with missing dependencies

MohammedAlkindi:
`docs/development.rst` said `tox -e py` "will skip tests for which you are missing dependencies", which is not what happens. `tests/config/cli/test_argcomplete.py` imports the optional `argcomplete` extra at module level, so without it collection aborts and no tests run at all:

```
ImportError while importing test module 'tests/config/cli/test_argcomplete.py'
E   ModuleNotFoundError: No module named 'argcomplete'
!!!!! Interrupted: 1 error during collection !!!!!
1 skipped, 19 deselected, 1 error in 2.56s
```

This drops that clause, per review feedback that guarding the import risks the coverage requirements and the documentation is the thing to correct.

Anyone running `tox -e py` is unaffected either way: `tox.toml` sets `extras = ["completion"]` on `env_run_base`, so argcomplete is always installed there.

Drafted with AI assistance; I ran the collection failure above myself on Windows 11 / Python 3.14.7.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -31,8 +31,7 @@ section. This section contains only highlights; it's not a substitute for readin
   testing you'll need to install your own "top-level" tox (using pipx_ or similar is fine) and use the following targets
   (tox environments):
 
-  - :samp:`tox -e py [-- {pytest-arg ...}]` to `test code changes <#running-tests>`_. This will skip tests for which you
-    are missing dependencies, but those tests will still be run by GitHub Actions.
+  - :samp:`tox -e py [-- {pytest-arg ...}]` to `test code changes <#running-tests>`_.
   - ``tox -e type`` to typecheck changes. (All new code should have complete type annotations.)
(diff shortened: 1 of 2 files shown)


gaborbernat:
Without this path, there is a good chance that the coverage requirements will fail. I think in this case we should actually fix the documentation to not say this, rather than make this actually optional.

MohammedAlkindi:
Fair point on coverage, that is not something I could see from outside. Happy to switch this to a docs change instead. Should the sentence in `development.rst` just lose the "will skip tests for which you are missing dependencies" clause, or say that the `completion` extra is needed for a full local run?

gaborbernat:
> Should the sentence in `development.rst` just lose the "will skip tests for which you are missing dependencies" clause

Yes!

MohammedAlkindi:
Done, that clause is gone and the test guard is reverted, so this is docs-only now.

PR #4053, opened 2026-08-31
🐛 fix(config): stop splitting factor names that look like ranges

MohammedAlkindi:
tox pulled an environment name ending in a digit, a hyphen and another digit apart, as though the tail were a generative range. `py313-django4-2` never got created: tox made `py313-django4`, `3` and `2` instead, and said nothing. `py310-1,py310-2` turned into more than three hundred environments, counting down from 310.

A range now has to open a factor, so digits carrying on a name stay part of it. The documented range forms keep working, including one written after a comma and a space, and a version-shaped name such as `3.10-2` survives too.

The same expansion sits behind `-e` on the command line, behind generative section names and behind factor conditions, so those pick up the fix as well.

The left-open form `{-13}` keeps its narrower rule. Widening it to match would swallow the trailing `-2` in `py3{10-11}-2,x`.

### DIFF
--- a/docs/changelog/4053.bugfix.rst
+++ b/docs/changelog/4053.bugfix.rst
@@ -1,0 +1,1 @@
+Only expand a generative digit range at a factor boundary, so a name like ``py313-django4-2`` is no longer split into separate environments.
--- a/docs/changelog/4053.bugfix.rst
+++ b/docs/changelog/4053.bugfix.rst
@@ -1,2 +1,2 @@
-Only expand a generative digit range at a factor boundary, so a name like ``py313-django4-2`` is no longer split into
-separate environments.
+A generative digit range now expands only where a factor starts, so an environment such as ``py313-django4-2`` keeps its
+name instead of splitting into ``py313-django4``, ``3`` and ``2`` - by :user:`MohammedAlkindi`.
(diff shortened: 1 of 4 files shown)


gaborbernat:
I reviewed this by diffing behaviour against the parent revision over 77 inputs covering the documented range forms, ranges after every kind of preceding character, negation, globs and whitespace. Nine inputs change and all nine are corrections. Nothing that expands today stops expanding. Worth recording that the same bug reaches `-e` on the command line and factor conditions, not just `env_list`, because `find_factor_groups` runs the same expansion.

I pushed two things to the branch.

The guard appeared on both range alternatives with no note of what it is for. It now sits once, in a non-capturing group over the two, which keeps the group numbering `_expand` unpacks, and it carries the reason.

The dot in that guard had no test behind it. Removing it left all 64 tests green while `3.10-2` went back to expanding into a countdown, which matters because `3.NN` is how tox names Python environments now. Two cases pin it, and I checked they fail for the right reason. Dropping the dot fails those two alone; removing the guard fails those two plus the two you added.

The changelog fragment lacked the `- by :user:` attribution every other fragment carries. I added yours and named what the environment splits into.

One boundary needs a decision rather than a change. A range-looking run after a factor separator still expands, so `mypy-1-2,other` gives `mypy-1,2,other`, and `{mypy-1-2}` gives `{mypy-1,2}`. That follows the rule this PR states, since `1-2` there does open a factor, so I left it. It only matters if bare ranges should expand inside braces alone, which is a wider question than this fix.

PR #4065, opened 2026-09-13
fix(config): accept empty set_env lists

yuefdev:
In TOML, `set_env = []` failed to load with `reduce() of empty iterable with no initial value`, but `set_env = {}` worked. An empty list now works the same way as an empty table and clears the inherited value.

AI assistance was used to investigate, implement, and test this change.


yuefdev:
The type errors also occur on main: https://github.com/tox-dev/tox/actions/runs/34758554641. The separate docs failure is a docutils AttributeError; I have not established its cause.

gaborbernat (approved):

PR #4066, opened 2026-09-13
fix(config): reject boolean TOML range bounds

yuefdev:
TOML factor ranges accepted `true` and `false` as `start` or `stop` because `bool` is a subclass of `int`. So `start = true` started the range at 1, and `stop = false` produced no environments. Booleans now raise the same error as any other non-integer bound.


yuefdev:
The remaining type errors also occur on main: https://github.com/tox-dev/tox/actions/runs/34758554641. They concern optional cwd values in tox_env/api.py.

gaborbernat (approved):

PR #4067, opened 2026-09-13
fix: correct coverage startup environment variable

yuefdev:
The move to `tox.toml` merged two keys into `COVERAGE_FILECOVERAGE_PROCESS_START`, so the test environments never set `COVERAGE_PROCESS_START`. With the name fixed, coverage's startup hook can measure Python subprocesses again.


yuefdev:
The two type errors also occur on main: https://github.com/tox-dev/tox/actions/runs/34758554641. Both are in tox_env/api.py, outside this change.

gaborbernat (approved):

gaborbernat (approved):

